PR #20, opened 2020-07-24
Fix build with nightly compiler

lperlaki:
depends on: [avr-device/27](https://github.com/Rahix/avr-device/pull/27)

### DIFF
--- a/avr-hal-generic/src/lib.rs
+++ b/avr-hal-generic/src/lib.rs
@@ -1,5 +1,5 @@
 #![no_std]
-#![feature(asm)]
+#![feature(asm, llvm_asm)]
 
 pub extern crate embedded_hal as hal;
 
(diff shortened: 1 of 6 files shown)


lperlaki:
This patch enables this libary to be build with `cargo build -Z build-std=core --target <avr-target>` on nightly-2020-07-24

avr-hal-generic/src/lib.rs
```diff
@@ -1,5 +1,5 @@
 #![no_std]
-#![feature(asm)]
+#![feature(asm, llvm_asm)]
```
Rahix, line 2: We can drop the `asm` feature for now as everything is using the old
syntax at this time.

Rahix:
Merged, thanks!

PR #22, opened 2020-07-24
Allow splitting Serial into Reader and Writer

lights0123:
I'm doing some fun stuff with `async`/`await` and came across the need to have one task reading separately from writing without blocking the other.

### DIFF
--- a/avr-hal-generic/src/serial.rs
+++ b/avr-hal-generic/src/serial.rs
@@ -89,6 +89,32 @@ macro_rules! impl_usart {
                     _clock: ::core::marker::PhantomData,
                 }
             }
+		}
+
+		$crate::paste::item! {
+	        impl<CLOCK, RX_MODE> $Usart<CLOCK, RX_MODE>
+	        where
+	            CLOCK: $crate::clock::Clock,
+	            RX_MODE: $crate::port::mode::InputMode,
+	        {
+	            /// Helper method for splitting this read/write object into two halves.
+	            ///
+	            /// The two halves returned implement the `Read` and `Write` traits, respectively.
+	            pub fn split(self) -> ([<Read $Usart>]<CLOCK, RX_MODE>, [<Write $Usart>]<CLOCK>) {
+	                (
+	                    [<Read $Usart>] {
+	                        p: unsafe { ::core::ptr::read(&self.p) },
+	                        rx: self.rx,
+	                        _clock: self._clock,
+	                    },
+	                    [<Write $Usart>] {
+	                        p: self.p,
+	                        tx: self.tx,
+	                        _clock: self._clock,
+	                    }
+	                )
+	            }
+	        }
         }
 
         impl<CLOCK, RX_MODE> $crate::hal::serial::Write<u8> for $Usart<CLOCK, RX_MODE>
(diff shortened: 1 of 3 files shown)


avr-hal-generic/src/serial.rs
```diff
@@ -89,6 +89,32 @@ macro_rules! impl_usart {
...
+	            CLOCK: $crate::clock::Clock,
+	            RX_MODE: $crate::port::mode::InputMode,
+	        {
+	            /// Helper method for splitting this read/write object into two halves.
+	            ///
+	            /// The two halves returned implement the `Read` and `Write` traits, respectively.
+	            pub fn split(self) -> ([<Read $Usart>]<CLOCK, RX_MODE>, [<Write $Usart>]<CLOCK>) {
+	                (
+	                    [<Read $Usart>] {
+	                        p: unsafe { ::core::ptr::read(&self.p) },
```
lights0123, line 106: I don't really like this, but I'm not sure there's a better way.

Rahix:
> I'm doing some fun stuff with async/await and came across the need to have one task reading separately from writing without blocking the other.

Just out of curiosity: I was also playing around with that but I'm running into tons of miscompilations (at least as far as I can tell ...).  Do you mind sharing you approach if you have something working?

avr-hal-generic/src/serial.rs
Rahix: If I'm not mistaken, this breaks safety.  The reader and write halves are now completely separate so I could move one of the two e.g. into an interrupt context and there would be no synchronization to ensure the two don't interfere.

I see a few ways to fix this:

- Use an async mutex (similar to [this one](https://github.com/ferrous-systems/async-on-embedded/blob/master/async-cortex-m/src/unsync/mutex.rs)) to wrap the peripheral and grant access to multiple tasks.  This would mean, no splitting is needed.
- Use a mutex of sorts internally to synchronize access between reader and writer.  E.g. a plain RefCell which would ensure the halves have to remain in the same hardware thread.
- Synchronize the halves by turning off interrupts in the read and write methods.  I feel like this is not a nice approach though, because it could potentially lock up the system for long times.

lights0123:
@Rahix I'm working on a blog post about getting this to work, but [here's what I have so far](https://github.com/lights0123/async-avr). I didn't run into a single compilation problem, and have async UART and SPI working just fine. My next step is to get interrupts working so it can go to sleep when not running a task, but I'll need to work on getting timers working before then.

avr-hal-generic/src/serial.rs
lights0123: Although I could mark it as `!Send` to prevent you from sending it across an interrupt bound (we should have that invariant, as otherwise you could do some bad sync stuff), I don't think that's a problem in the first place—reading and writing is done in the same register, but I don't think there's anything that can go wrong. Reading is done in two (main) instructions:

1. Read control register A
2. Read data register

Writing is done in two (main) instructions as well:

1. Read control register A
2. Write data register

But I don't see how placing an interrupt in between either of those could cause problems, as it's not like writing can empty the read buffer or reading can add to the send buffer.

Rahix:
I see, you've chosen a different design from what I tried, which does not need trait objects.  Because that's probably what bit me: https://github.com/rust-lang/rust/issues/74743

I have to say, this looks really cool!

avr-hal-generic/src/serial.rs
Rahix: Right, I've re-read the datasheet and it looks like you are correct!  In that case, I think your code is okay.

lights0123:
Fixed—IntelliJ decided to ignore the existing indentation style.

lights0123:
@Rahix here's my blog post in case you're interested: http://localhost:8082/blog/2020/07/25/async-await-for-avr-with-rust/

Rahix:
Really cool!  I hope to iron out the remaining uglyness of getting `avr-device` to work and finally get it published to crates.io, soon.  And then, `avr-hal` as well, of course.

Rahix:
For waking up your tasks, maybe the [`leonardo-interrupt`](https://github.com/Rahix/avr-hal/blob/master/boards/arduino-leonardo/examples/leonardo-interrupt.rs) example I recently added helps?

lights0123:
Yeah, I know *how* to do it, I just need to do it. I think I'm going to send a PR here to expose an API in the Usart* structs to enable interrupts (because you overwrite previously set control registers), and then have a global function in my crate that you call for each interrupt type (i.e. "I got a byte from a serial port!"). I just need to figure out if I want to have a global e.g. 32-byte buffer in case the task isn't resumed in time (a la Arduino) or just wake up the task.

I would prefer, though, if your interrupt macros took an interrupt name and then figured out how to map that to the chip (I think you have everything you need to do that, because you have feature flags), i.e. [this](https://github.com/nh2/quadcopter-simulation/blob/8a3652b8a704877156e91c0691ce8ce37588eb53/arduino-accel-rust/src/main.rs#L300-L327). Also, is there any need for the interrupt handlers to always be `unsafe`?

Also, would it be possible to use form, svd2rust, and atdf2svd as build deps, and figure out a way to get around the pyyaml requirement? That would be *amazing*.

Rahix:
> I think I'm going to send a PR here to expose an API in the Usart* structs to enable interrupts (because you overwrite previously set control registers)

Ouch, true ...  Yeah, please do!

>  would prefer, though, if your interrupt macros took an interrupt name and then figured out how to map that to the chip (I think you have everything you need to do that, because you have feature flags)

Hm, not sure if that is possible.  The interrupt names collide between different chips and because of how cargo's features work, this can be tricky.  Originally I only allowed a single feature to be enabled, but that would also lead to its own issues ... If you have a solution, feel free to send a PR, though!

> Also, is there any need for the interrupt handlers to always be unsafe?

Huh, that should not be necessary?  If it is, that's a bug in the macro.

> Also, would it be possible to use form, svd2rust, and atdf2svd as build deps, and figure out a way to get around the pyyaml requirement? That would be amazing.

That's what I'm working on right now; once ready for crates.io, none of those dependencies will be needed anymore :)

BTW, don't forget to submit your blog post for TWiR!

lights0123:
Oh yeah, I misread this line and thought that it said that `unsafe` was required instead of what it actually is, optional.

https://github.com/Rahix/avr-device/blob/c314463726a2867b038b243f818b1f999c8ca22a/macros/src/lib.rs#L63

PR #23, opened 2020-07-25
Adc implementation

### LINKED ISSUE #11: ADC Implementation
Implement the traits from [`embedded_hal::adc`](https://docs.rs/embedded-hal/0.2.3/embedded_hal/adc/index.html) as applicable.  Request from avr-rust/ruduino#13.

lperlaki:
My attempt of implementing ADC currently only for the Arduino Uno

Needs [avr-device/29](https://github.com/Rahix/avr-device/pull/29)

Fixes #11.

### DIFF
--- a/avr-hal-generic/src/adc.rs
+++ b/avr-hal-generic/src/adc.rs
@@ -23,7 +23,7 @@ impl Default for ClockRateDivision {
 
 /// Select the voltage reference for the ADC peripheral
 ///
-/// The inernal voltage reference options may not be used if an external reference voltage isbeing applied to the AREF pin.
+/// The internal voltage reference options may not be used if an external reference voltage is being applied to the AREF pin.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum ReferenceVoltage {
     /// Voltage applied to AREF pin.
--- a/avr-hal-generic/src/adc.rs
+++ b/avr-hal-generic/src/adc.rs
@@ -1,5 +1,8 @@
 
-
+/// The division factor between the system clock frequency and the input clock to the AD converter.
+///
+/// To get 10bit precision, clock from 50kHz to 200kHz must be supplied. If you need less precision, you can supply higher clock.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum ClockRateDivision {
     Factor2,
     Factor4,
@@ -10,23 +13,39 @@ pub enum ClockRateDivision {
     Factor128,
 }
 
+
+impl Default for ClockRateDivision {
+    fn default() -> Self {
+       Self::Factor128
+    }
+}
+
+
+/// Select the voltage reference for the ADC peripheral
+///
+/// The inernal voltage reference options may not be used if an external reference voltage isbeing applied to the AREF pin.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum ReferenceVoltage {
+    /// Voltage applied to AREF pin.
     Aref,
-    Vcc,
+    /// Default referece voltage.
+    AVcc,
+    /// Internal reference voltage
     Internal,
 }
 
-pub struct AdcSettings{
-    pub adps: ClockRateDivision,
-    pub aref: ReferenceVoltage
-}
-
-
-impl Default for AdcSettings {
+impl Default for ReferenceVoltage {
     fn default() -> Self {
-        Self { adps: ClockRateDivision::Factor128, aref: ReferenceVoltage::Vcc}
+       Self::AVcc
     }
 }
+
+#[derive(Default, Debug, Clone, Copy, PartialEq, Eq)]
+pub struct AdcSettings{
+    pub clock_divider: ClockRateDivision,
+    pub ref_voltage: ReferenceVoltage
+}
+
 
 #[macro_export]
 macro_rules! impl_adc {
@@ -34,7 +53,7 @@ macro_rules! impl_adc {
         pub struct $Adc:ident {
             type ChannelID = $ID:ty;
             peripheral: $ADC:ty,
-            pins: {$($pxi:ident: ($PXi:ident, $ChannelID:expr, $name:ident),)+}
+            pins: {$($pxi:ident: ($PXi:ident, $ChannelIDExpr:expr, $ChannelIDPat:pat, $name:ident),)+}
         }
     ) => {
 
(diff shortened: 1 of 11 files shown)


avr-hal-generic/src/adc.rs
```diff
@@ -0,0 +1,129 @@
...
+                        Ok(self.peripheral.adc.read().bits().into())
+                    },
+                    (false, _) => {
+                        self.is_reading = true; self.peripheral.admux.modify(|_, w| w.adlar().clear_bit()
+                                                                                     .mux().variant(PIN::channel()));
+                        self.peripheral.adcsra.modify(|_, w| w.adsc().set_bit());
+                        Err(nb::Error::WouldBlock)
+                    }
+                }
+            }
```
Rahix, line 109: I think this method has a problem:

If I call it once with a pin 1, it will return `WouldBlock`.  If I then poll with a different pin 2, until I receive a result, I will get a result for pin 1.  Not sure what upstream `embedded-hal` thinks of this; the docs don't mention this case ... I would suggest to restart the measurement and start sampling pin 2 instead. Or take a look what other implementations are doing here ...

lperlaki: Changed the implementation so every pin has its own `is_reading`flag, I think that should work

avr-hal-generic/src/adc.rs
```diff
@@ -0,0 +1,129 @@
...
+
+            fn read(&mut self, _pin: &mut PIN) -> nb::Result<WORD, Self::Error> {
+                match (self.is_reading, self.peripheral.adcsra.read().adsc().bit_is_set()) {
+                    (true, true) =>  Err(nb::Error::WouldBlock),
+                    (true, false) => {
+                        self.is_reading = false;
+                        Ok(self.peripheral.adc.read().bits().into())
+                    },
+                    (false, _) => {
+                        self.is_reading = true; self.peripheral.admux.modify(|_, w| w.adlar().clear_bit()
```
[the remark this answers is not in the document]
lperlaki: Forgot to remove that

lperlaki:
@Rahix I am not to sure about my into_analog Pin implementation if you have some input on that

boards/arduino-uno/examples/uno-adc.rs
```diff
@@ -0,0 +1,46 @@
...
+
+
+    ufmt::uwriteln!(&mut serial, "Reading Analog Input on PORT a0\r").unwrap();
+
+    loop {
+        // Read the Analog value
+        let aread: u16 = nb::block!{adc.read(&mut a0)}.unwrap();
+
+        // Write it to Serial
+        ufmt::uwriteln!(&mut serial, "read: {:?}", aread);
```
Rahix, line 44: `{}` should do the trick as well, no need for `{:?}`.  Also, I'd add a `\r` at the end for nicer output.

```suggestion
        ufmt::uwriteln!(&mut serial, "read: {}\r", aread);
```

avr-hal-generic/src/adc.rs
```diff
@@ -0,0 +1,165 @@
...
+impl Default for ClockRateDivision {
+    fn default() -> Self {
+       Self::Factor128
+    }
+}
+
+
+/// Select the voltage reference for the ADC peripheral
+///
+/// The inernal voltage reference options may not be used if an external reference voltage isbeing applied to the AREF pin.
```
Rahix, line 26:
```suggestion
/// The internal voltage reference options may not be used if an external reference voltage is being applied to the AREF pin.
```

avr-hal-generic/src/adc.rs
```diff
@@ -0,0 +1,165 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ReferenceVoltage {
+    /// Voltage applied to AREF pin.
+    Aref,
+    /// Default referece voltage.
```
Rahix, line 31:
```suggestion
    /// Default reference voltage.
```

avr-hal-generic/src/adc.rs
```diff
@@ -0,0 +1,165 @@
...
+                    $ChannelIDExpr
+                }
+            }
+
+            impl<MODE> $PXi<MODE> {
+                    /// Make this pin a analog input and enable the internal pull-up
+                    pub fn into_analog_input(self, adc: &mut $Adc) -> $PXi<Analog> {
+                        adc.peripheral.didr0.modify(|_, w| w.$name().set_bit());
+                        $PXi { _mode: core::marker::PhantomData }
+                    }
```
Rahix, line 161: This is what I'd do as well, so :+1:.  Though, looking into the future, ATmega32U4 has 2 _Digital Input Disable Registers_ so if you want, I would suggest making the register name a macro parameter as well.

avr-hal-generic/src/adc.rs
Rahix: Yes, this would work but I feel like 8 `bool`s just to work around a corner case is a bit much to pay.  Can't we reuse the value of `ADMUX[MUX]` here?  I think it should be possible to compare it against the enum variant for the pin (`PIN::channel()`) and base the decision off of that.

avr-hal-generic/src/adc.rs
lperlaki: Ok changed that.

avr-hal-generic/src/adc.rs
lperlaki: Right a single field with an enum would be better.
I would like to use ADMUX[MUX] but it doesn't have a state for no Channel selected which is my initial state of the Adc,

Rahix: Fair point, the new implementation looks good!

avr-hal-generic/src/adc.rs
```diff
@@ -0,0 +1,154 @@
...
+                        Ok(self.peripheral.adc.read().bits().into())
+                    },
+                    (Some(_), _) => Err(nb::Error::Other(AdcError::AlreadyReaddingOtherChannel)),
+                    (None, _) => {
+                        self.reading_channel = Some(PIN::channel());
+                        self.peripheral.admux.modify(|_, w| w.mux().variant(PIN::channel()));
+                        self.peripheral.adcsra.modify(|_, w| w.adsc().set_bit());
+                        Err(nb::Error::WouldBlock)
+                    }
+                }
```
Rahix, line 133: I think the error when re-reading with a different pin is not the right choice.  If you don't want to support this case, I would `panic!()` as this would clearly be a programmer error and not an unforseeable hardware error.  But I would instead suggest to restart the measurement on the other pin when this happens.  Rough outline (untested of course):

```rust
match (self.reading_channel, self.peripheral.adcsra.read().adsc().bit_is_set()) {
    // Measurement on current pin is ongoing
    (Some(channel), true) if channel == PIN::channel() =>  Err(nb::Error::WouldBlock),
    // Measurement on current pin completed
    (Some(channel), false) if channel == PIN::channel() => {
        self.reading_channel = None;
        Ok(self.peripheral.adc.read().bits().into())
    },
    // Measurement on other pin is ongoing
    (_, true) => {
        // Abort the current measurement when requesting a different pin.
        self.reading_channel = None;
        Err(nb::Error::WouldBlock)
    },
    // Start measurement
    (None, false) => {
        self.reading_channel = Some(PIN::channel());
        self.peripheral.admux.modify(|_, w| w.mux().variant(PIN::channel()));
        self.peripheral.adcsra.modify(|_, w| w.adsc().set_bit());
        Err(nb::Error::WouldBlock)
    }
}
```

What do you think?

---

Looking at other HALs, I see that `read()` is implemented synchroneously in a lot of cases which means they don't have to deal with this at all.  An asynchroneous version is of course a lot nicer.

lperlaki: Sound good to me

lperlaki:
I did some tests similar to the `uno-adc` example on an Arduino Uno board
Would also like to implement adc for the `atmetga32u4` but I don't currently have a board with that chip.

Rahix:
>  Would also like to implement adc for the atmetga32u4 but I don't currently have a board with that chip.

I can take care of that, should not be difficult ontop of this PR :)

PR #26, opened 2020-07-26
Add ADC support for ATmega32U4

Rahix:
Add support for the ADC peripheral for `ATmega32U4` (i.e. Arduino Leonardo).  This needs a slight change in the `impl_adc!{}` macro, because `ATmega32U4` has slightly different register layouts (the MUX field is split across 2 registers).

### DIFF
--- a/chips/atmega328p-hal/src/adc.rs
+++ b/chips/atmega328p-hal/src/adc.rs
@@ -8,6 +8,9 @@ avr_hal::impl_adc! {
     pub struct Adc {
         type ChannelID = MUX_A;
         peripheral: crate::atmega328p::ADC,
+        set_mux: |peripheral, id| {
+            peripheral.admux.modify(|_, w| w.mux().variant(id));
+        },
         pins: {
             pc0: (PC0, MUX_A::ADC0, didr0::adc0d),
             pc1: (PC1, MUX_A::ADC1, didr0::adc1d),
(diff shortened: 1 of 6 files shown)


Rahix:
@lperlaki: Could you take a quick look if this is alright?

Rahix:
I've added types for the **V<sub>bandgap</sub>**, **GND**, and **Temperature Sensor** channels, also for `ATmega328P`.

chips/atmega328p-hal/src/adc.rs
```diff
@@ -18,3 +21,25 @@ avr_hal::impl_adc! {
...
+        fn channel() -> Self::ID { MUX_A::ADC_VBG }
+    }
+
+    /// Channel for _GND_
+    pub struct Gnd;
+    impl Channel<super::Adc> for Gnd {
+        type ID = MUX_A;
+        fn channel() -> Self::ID { MUX_A::ADC_GND }
+    }
+}
```
lperlaki, line 45: I looks like you forgot to add the Temperature Channel for the atmega328p

Rahix:
> Wouldn't it be better to create a Device patch for the mux register?

I tried writing the patch but because each value can have two different meanins (depending in `ADCSRB[MUX5]`) it's hard to create descriptive names for the enumerated values.  So I opted to use the raw values directly.  Though maybe I could create and enum here in `atmega32u4-hal` which maps descriptive names to the entire 6-bit mux values.

chips/atmega328p-hal/src/adc.rs
Rahix: Ack, I wasn't entirely sure how the temperature sensor works on ATmega328P.  I'll have to take another look at the datasheet.

PR #30, opened 2020-07-31
Add Arduino Mega 2560, ATmega2560

couchand:
More or less just copied from the ATmega1280 and BigAVR6 definitions, with the minimal changes needed to get the examples running.

Note that the `avr-device` dep is currently specified using a relative path, this presumably should be updated to use the new version once https://github.com/Rahix/avr-device/pull/32 lands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,7 @@ The chip-HAL crates currently support the following peripherals:
   - [x] `PORTA`, `PORTB`, `PORTC`, `PORTD`, `PORTE`, `PORTF`, `PORTG`, `PORTH`, `PORTJ`, `PORTK`, `PORTL` as digital IO
   - [x] `USART0`, `USART1`, `USART2`, `USART3` for serial communication
   - [x] I2C using `TWI`
+  - [x] SPI
   - [x] ADC (no differential channels yet)
 * [`atmega328p-hal`](./chips/atmega328p-hal)
   - [x] Spinning Delay
--- a/README.md
+++ b/README.md
@@ -99,6 +99,7 @@ The chip-HAL crates currently support the following peripherals:
   - [x] `PORTA`, `PORTB`, `PORTC`, `PORTD`, `PORTE`, `PORTF`, `PORTG`, `PORTH`, `PORTJ`, `PORTK`, `PORTL` as digital IO
   - [x] `USART0`, `USART1`, `USART2`, `USART3` for serial communication
   - [x] I2C using `TWI`
+  - [x] ADC (no differential channels yet)
 * [`atmega328p-hal`](./chips/atmega328p-hal)
   - [x] Spinning Delay
   - [x] `PORTB`, `PORTC`, `PORTD` as digital IO (v2)
--- a/README.md
+++ b/README.md
@@ -119,6 +119,8 @@ In `boards/` there are crates for the following hardware.  Please note that this
   - [Website](https://www.arduino.cc/en/Main/Arduino_BoardLeonardo)
 * [Arduino Uno](./boards/arduino-uno)
   - [Website](https://store.arduino.cc/usa/arduino-uno-rev3)
+* [Arduino Mega 2560](./boards/arduino-mega2560)
+  - [Website](http://arduino.cc/en/Main/ArduinoBoardMega2560)
 * [Adafruit Trinket (3V3 or 5V)](./boards/trinket) (**not** PRO!)
   - [Website](https://learn.adafruit.com/introducing-trinket)
 * [BigAVR 6](./boards/bigavr6)
--- a/README.md
+++ b/README.md
@@ -94,6 +94,11 @@ The following peripherals are supported in `avr-hal-generic`:
 
 ### HAL Status
 The chip-HAL crates currently support the following peripherals:
+* [`atmega2560-hal`](./chips/atmega2560-hal)
+  - [x] Spinning Delay
+  - [x] `PORTA`, `PORTB`, `PORTC`, `PORTD`, `PORTE`, `PORTF`, `PORTG`, `PORTH`, `PORTJ`, `PORTK`, `PORTL` as digital IO
+  - [x] `USART0`, `USART1`, `USART2`, `USART3` for serial communication
+  - [x] I2C using `TWI`
 * [`atmega328p-hal`](./chips/atmega328p-hal)
   - [x] Spinning Delay
   - [x] `PORTB`, `PORTC`, `PORTD` as digital IO (v2)
--- a/boards/arduino-mega2560/examples/mega2560-serial.rs
+++ b/boards/arduino-mega2560/examples/mega2560-serial.rs
@@ -1,0 +1,32 @@
+#![no_std]
+#![no_main]
+
+extern crate panic_halt;
+use arduino_mega2560::prelude::*;
+
+#[arduino_mega2560::entry]
+fn main() -> ! {
+    let dp = arduino_mega2560::Peripherals::take().unwrap();
+
+    let mut porte = dp.PORTE.split();
+    let mut serial = arduino_mega2560::Serial::new(
+        dp.USART0,
+        porte.pe0,
+        porte.pe1.into_output(&mut porte.ddr),
+        57600,
+    );
+
+    // The following would also work, but needs +600% more bytes
+    // writeln!(serial, "Hello from Arduino!\r").unwrap();
+    serial.write_str("Hello from Arduino!\r\n").void_unwrap();
+
+    loop {
+        // Read a byte from the serial connection
+        let b = nb::block!(serial.read()).void_unwrap();
+
+        // Answer
+        serial.write_str("You pressed ").void_unwrap();
+        nb::block!(serial.write(b)).void_unwrap();
+        serial.write_str("!\r\n").void_unwrap();
+    }
+}
--- a/boards/arduino-mega2560/examples/mega2560-serial.rs
+++ b/boards/arduino-mega2560/examples/mega2560-serial.rs
@@ -16,9 +16,7 @@ fn main() -> ! {
         57600,
     );
 
-    // The following would also work, but needs +600% more bytes
-    // writeln!(serial, "Hello from Arduino!\r").unwrap();
-    serial.write_str("Hello from Arduino!\r\n").void_unwrap();
+    ufmt::uwriteln!(&mut serial, "Hello from Arduino!\r").void_unwrap();
 
     loop {
         // Read a byte from the serial connection
--- a/boards/arduino-mega2560/src/lib.rs
+++ b/boards/arduino-mega2560/src/lib.rs
@@ -10,6 +10,7 @@ mod pins;
 pub use atmega2560_hal::atmega2560;
 pub use crate::atmega2560::Peripherals;
 pub use atmega2560_hal::prelude;
+pub use atmega2560_hal::adc;
 pub use crate::pins::*;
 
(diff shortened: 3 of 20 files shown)


chips/atmega2560-hal/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+[features]
+rt = ["avr-device/rt"]
+
+[dependencies]
+avr-hal-generic = { path = "../../avr-hal-generic/" }
+
+[dependencies.avr-device]
+version = "0.1.0"
+features = ["atmega2560"]
+path = "../../../avr-device"
```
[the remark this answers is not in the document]
couchand: I haven't seen that option, but it will be very useful here and elsewhere, thanks!

README.md
```diff
@@ -119,6 +119,8 @@ In `boards/` there are crates for the following hardware.  Please note that this
   - [Website](https://www.arduino.cc/en/Main/Arduino_BoardLeonardo)
 * [Arduino Uno](./boards/arduino-uno)
   - [Website](https://store.arduino.cc/usa/arduino-uno-rev3)
+* [Arduino Mega 2560](./boards/arduino-mega2560)
+  - [Website](http://arduino.cc/en/Main/ArduinoBoardMega2560)
```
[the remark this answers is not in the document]
couchand: Can do.

boards/arduino-mega2560/src/lib.rs
```diff
@@ -0,0 +1,14 @@
...
+#[cfg(feature = "rt")]
+pub use hal::entry;
+
+pub use atmega2560_hal::atmega2560;
+pub use crate::atmega2560::Peripherals;
+pub use atmega2560_hal::prelude;
+
+pub type Delay = hal::delay::Delay<hal::clock::MHz16>;
+pub type Serial<IMODE> = atmega2560_hal::usart::Usart0<hal::clock::MHz16, IMODE>;
+pub type I2c<M> = hal::i2c::I2c<hal::clock::MHz16, M>;
```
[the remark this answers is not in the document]
couchand: Of course.  It looks like multiplexed pins just have one alias with doc comments specifying other usages, and as far as I can tell every relevant pin on the Mega 2560 has a plain `dn`/`an` identifier, so there won't be dedicated `miso`, `mosi`, `sck`, `led_tx` and `led_rx` entries.  Does that sound right?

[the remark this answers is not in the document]
Rahix: For a start, just add the labelled pins.

If you want to go further you'll have to take a look at the schematic whether any pins are connected elsewhere.  E.g. the `ICSP` connector, on-board peripherals, ... Though if pins have multiple connections, always prefer the label name, of course.

couchand:
@Rahix I've made the various changes.  The pins file was a beast, this thing has an awful lot of pins!  Would certainly appreciate another pair of eyes on it to make sure everything is in order.  Thanks!

couchand:
Oh, I just noticed that I would need to actually do some implementation of features if I want to claim support for them in the README...  I can take a look at the analogous Leonardo implementations to try to figure those out.

If you have specific pointers I'd certainly appreciate it, and in the meantime perhaps those boxes should not be checked.

Rahix:
> @Rahix I've made the various changes. The pins file was a beast, this thing has an awful lot of pins! Would certainly appreciate another pair of eyes on it to make sure everything is in order. Thanks!

I'll take a look!

> Oh, I just noticed that I would need to actually do some implementation of features if I want to claim support for them in the README... I can take a look at the analogous Leonardo implementations to try to figure those out.
> 
> If you have specific pointers I'd certainly appreciate it, and in the meantime perhaps those boxes should not be checked.

I suggest for the initial version we'll leave those out and you can then send a follow-up MR with the implementations :) You're best off looking at the leonardo code as a basis (`atmega32u4-hal`) and if you have any questions, just ask!

README.md
```diff
@@ -94,6 +94,13 @@ The following peripherals are supported in `avr-hal-generic`:
 
 ### HAL Status
 The chip-HAL crates currently support the following peripherals:
+* [`atmega2560-hal`](./chips/atmega2560-hal)
+  - [x] Spinning Delay
+  - [x] `PORTA`, `PORTB`, `PORTC`, `PORTD`, `PORTE`, `PORTF`, `PORTG`, `PORTH`, `PORTJ`, `PORTK`, `PORTL` as digital IO
+  - [x] `USART0`, `USART1`, `USART2`, `USART3` for serial communication
+  - [x] I2C using `TWI`
+  - [x] SPI
+  - [x] ADC
```
Rahix, line 103: As discussed, please remove the ones which aren't yet implemented.

chips/atmega2560-hal/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+edition = "2018"
+
+[features]
+rt = ["avr-device/rt"]
+
+[dependencies]
+avr-hal-generic = { path = "../../avr-hal-generic/" }
+
+[dependencies.avr-device]
+version = "0.1.0"
```
Rahix, line 14: Update was released!

```suggestion
version = "0.1.1"
```

boards/arduino-mega2560/src/pins.rs
```diff
@@ -0,0 +1,346 @@
...
+        /// * `PCINT2`: External Interrupt (Pin Change)
+        pub d51: portb::pb2::PB2,
+        /// `D52`
+        ///
+        /// * `SCK`: SPI bus Serial Clock
+        /// * `PCINT1`: External Interrupt (Pin Change)
+        pub d52: portb::pb1::PB1,
+        /// `D53`
+        ///
+        /// * `SS`: SPI bus Chip Select
```
Rahix, line 266:
```suggestion
        /// * `CS`: SPI bus Chip Select
```

boards/arduino-mega2560/src/pins.rs
```diff
@@ -0,0 +1,346 @@
...
+        ///
+        /// * `ADC14`: A/D converter input 14
+        /// * `PCINT22`: External Interrupt (Pin Change)
+        pub a14: portk::pk6::PK6,
+        /// `A15`
+        ///
+        /// * `ADC15`: A/D converter input 15
+        /// * `PCINT23`: External Interrupt (Pin Change)
+        pub a15: portk::pk7::PK7,
+    }
```
Rahix, line 345: I'd additionally add `led_tx` (= `PD5`) and `led_rx` (= `PD4`) as well.

README.md
couchand: Done.

chips/atmega2560-hal/Cargo.toml
couchand: Done.

boards/arduino-mega2560/src/pins.rs
couchand: Well I tend to prefer the terminology "Chip Select" when possible, but since the datasheet uses "Slave Select" I've switched it to be consistent with that.

boards/arduino-mega2560/src/pins.rs
couchand: Those LEDs are connected to port D on the ATmega32U4 that manages the USB interface, not port D of the ATmega2560.

couchand:
Comments addressed, please take a look, thanks!

couchand:
Well it really wasn't that much more effort to get the ADC and SPI support in.  Thanks for establishing the patterns that made this pretty easy to do!

boards/arduino-mega2560/src/pins.rs
Rahix: Ah, I see!

chips/atmega2560-hal/src/adc.rs
```diff
@@ -0,0 +1,56 @@
...
+use crate::port::portk::{PK0, PK1, PK2, PK3, PK4, PK5, PK6, PK7};
+
+use crate::atmega2560::adc::admux::MUX_A;
+
+avr_hal::impl_adc! {
+    pub struct Adc {
+        type ChannelID = MUX_A;
+        peripheral: crate::atmega2560::ADC,
+        set_mux: |peripheral, id| {
+            peripheral.admux.modify(|_, w| w.mux().variant(id));
```
Rahix, line 13: There's a hidden issue here ... the `ADMUX[MUX]` field is only 5 bits wide; the 6th bit is `ADCSRB[MUX5]`.  However, the enumerated values for for `ADMUX[MUX]` contain _all_ 6bit values valid for the combination of the two.  The problem is than when writing here, the uppermost bit will get cut off and ignored.  Leading to the wrong channels being read for MUX values >31.  E.g. `ADC8` will actually read `ADC0`, `ADC9` will read `ADC10`, etc.

This is obviously a problem and the most correct way to deal with this would be to remove all enumerated values >31 with an SVD patch so you can't accidentally do this in the first place.  This would then mean you'll have to approach this similar to what I did in [`atmega32u4_hal::adc`](https://github.com/Rahix/avr-hal/blob/e3a39bf14f349cc13b916ac6910330cdc38e268a/chips/atmega32u4-hal/src/adc.rs#L7-L29) where I redefined the enumerated values on my own.

However, this is a lot of unnecessary work IMO when we can do a quicker fix here by just extracting the uppermost bit and setting that separately:

```suggestion
            peripheral.admux.modify(|_, w| w.mux().variant(id));
            // Maybe add a comment why this is necessary ...
            peripheral.adcsrb.modify(|_, w| w.mux5().bit((id as u8) & 0b100000 != 0));
```

boards/arduino-mega2560/examples/mega2560-serial.rs
```diff
@@ -0,0 +1,32 @@
...
+    let mut serial = arduino_mega2560::Serial::new(
+        dp.USART0,
+        porte.pe0,
+        porte.pe1.into_output(&mut porte.ddr),
+        57600,
+    );
+
+    // The following would also work, but needs +600% more bytes
+    // writeln!(serial, "Hello from Arduino!\r").unwrap();
+    serial.write_str("Hello from Arduino!\r\n").void_unwrap();
```
Rahix, line 21: The idiomatic way for writing to serial is to use `ufmt::uwriteln!()`.  I suspect you got this from another place in the code-base which wasn't updated yet?

```suggestion
    ufmt::uwriteln!(&mut serial, "Hello from Arduino!\r").void_unwrap();
```

chips/atmega2560-hal/src/adc.rs
couchand: Indeed!  That's a pretty clever way to address the issue.

I confirmed that without this change, attempts to read single-ended ADC8-15 results in reads of ADC0-7, and that with the fix in place the reads are correct.  I made no attempts to test differential reads.

boards/arduino-mega2560/examples/mega2560-serial.rs
couchand: Done.  I believe the code was copied from the corresponding BigAVR6 example.

couchand:
Fixes applied, please take another look, thanks.

boards/arduino-mega2560/examples/mega2560-serial.rs
Rahix: Thanks! I'll try to remember I have to update that one as well.

chips/atmega2560-hal/src/adc.rs
Rahix: Sounds good!

PR #34, opened 2020-08-04
Add support for ATtiny88

couchand:
This much of the tiny88 was not too hard once I got the `avr-device` patches in place.

Looks like none of the other `chips` have example files, but I'd like to add some here, any objection?

I'm a bit hung up on the ADC support, feel free to take a look at https://github.com/couchand/avr-hal/tree/202008/attiny88 (relevant commit: https://github.com/couchand/avr-hal/commit/8bc9f59bc97c585c9b94dad12c622b7771ad13e6) to see my current attempt.  One thing I really don't like over there is the `ReferenceVoltage` enum having an extra variant that's not relevant to this model.  Maybe I should open a separate PR for that discussion?

Of course this will need an updated version of `avr-device` before it will build.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ members = [
     "chips/atmega328p-hal",
     "chips/atmega32u4-hal",
     "chips/attiny85-hal",
+    "chips/attiny88-hal",
 
     # The board crates
     "boards/arduino-leonardo",
--- a/README.md
+++ b/README.md
@@ -111,6 +111,9 @@ The chip-HAL crates currently support the following peripherals:
 * [`attiny85-hal`](./chips/attiny85-hal)
   - [x] Spinning Delay
   - [x] `PORTB` as digital IO (v2)
+* [`attiny88-hal`](./chips/attiny88-hal)
+  - [x] Spinning Delay
+  - [x] `PORTA`, `PORTB`, `PORTC`, `PORTD` as digital IO
 
 ### Supported Hardware
 In `boards/` there are crates for the following hardware.  Please note that this project is in no way affiliated with any of the vendors.
--- a/README.md
+++ b/README.md
@@ -114,6 +114,8 @@ The chip-HAL crates currently support the following peripherals:
 * [`attiny88-hal`](./chips/attiny88-hal)
   - [x] Spinning Delay
   - [x] `PORTA`, `PORTB`, `PORTC`, `PORTD` as digital IO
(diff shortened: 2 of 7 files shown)


Rahix:
> Looks like none of the other chips have example files, but I'd like to add some here, any objection?

I'd prefer examples to always be related to a concrete hardware where they can run directly without any further hardware or software modification.  For HAL crates, I'd move "examples" into their documentation because that is where someone looking how to use a HAL crate would ideally look first.  If there's some very chip-specific thing that makes sense to demonstrate in the HAL crate, that would totally fit as an example though.  I just wouldn't add the common "blink, serial, adc, i2c, spi" examples, because they just add more maintenance burden to keep them updated without a real board to test them against.

There's also the issue that your examples will need the `rt` feature which isn't enabled by default for HAL crates.  So for building them one would need to keep in mind to compile with `cargo build --examples --features rt` ...

> I'm a bit hung up on the ADC support, feel free to take a look at https://github.com/couchand/avr-hal/tree/202008/attiny88 (relevant commit: couchand@8bc9f59) to see my current attempt. One thing I really don't like over there is the ReferenceVoltage enum having an extra variant that's not relevant to this model. Maybe I should open a separate PR for that discussion?

Yeah, I would suggest keeping it out of this PR for now and opening a new one.  I see two options for the implementation:

1. Either do it like you did, panic on unsupported values.  Of course it isn't too nice to provide an option that doesn't actually exist ...
2. Instead, we could push the settings struct & enum definitions into the macro and define possible values as part of the macro invocation.

Rahix:
FYI: I've updated `avr-device` to `0.2.0` so ATtiny88 support is now available.

couchand:
Great!  I finally sorted out that I *don't* want to use the `-D` flag when programming and my little demo code is flashing some LEDs, so this looks ready to go.  I'll open a followup PR with ADC support.

Thanks!

Rahix:
Just one thing to keep in mind: CI currently does not build this crate because there isn't a board for it.  I'll look into adding jobs for no-board HALs as well.

couchand:
Noted, thanks.  It might not exactly be a priority at the moment (since the only way to use this code is to checkout the repo :smile:), but worth setting up before publishing the crates.

Rahix:
Well, it does matter for me, because it gives me a quick way to see whether a PR breaks any of the crates.  Especially changes in the generic part could silently break a HAL crate the author didn't test-build ;)

PR #36, opened 2020-08-04
Added u8 ms delay

Marcuss2:
I needed to implement it to make hd44780-driver work, as it needs u8 delay. The author is already making changes to u16, but it could be needed for other libraries.


Marcuss2:
I encountered some problems with invalid instructions while testing the changed examples, I hope it shows up well on the CI.

Marcuss2:
Changes made, the CI now passes.

PR #46, opened 2020-08-15
Update path of json for ATmega32U4 on readme

asukiaaa:
Thank you for sharing interesting project.
Path of `avr-atmega32u4.json` on readme seems old so I updated it.


Rahix:
Thanks!

PR #48, opened 2020-08-16
spi: Fix F_osc/128 configuration

Rahix:
The configuration code mistakenly set up the peripheral for F_osc/64 instead of the selected 128 clock divider.  Fix this.


Rahix:
cc @jonahbron

PR #49, opened 2020-08-17
Support floating MISO in SPI

### LINKED ISSUE #45: Support externally-pulled SPI
The current implementation demands that `SCLK` and `MOSI` be in `mode::Output` and `MISO` be in `mode::Input<mode::PullUp>` before creating the `Spi`.  But the hardware would support more general bounds.

The input `MISO` could be floating, to allow for an external pull-up, and it seems this change can be made without any other code changes.

And while normal SPI operation is push-pull, it is occasionally configured as open-drain, which also seems to be supported by the hardware.  The datasheets clearly state the output pins' pin directions are user-defined.  Since I haven't found authoritative documentation on the matter and haven't set up a test bench, I'm reading between the lines a bit, but I interpret this to mean the SPI peripheral will output the signals by toggling the DDR rather than the pin.  Perhaps real testing is needed before making code changes here.

couchand:
Closes #45.

### DIFF
--- a/avr-hal-generic/src/spi.rs
+++ b/avr-hal-generic/src/spi.rs
@@ -86,7 +86,7 @@ macro_rules! impl_spi {
 
         type SCLK = $sclkmod::$SCLK<$crate::port::mode::Output>;
         type MOSI = $mosimod::$MOSI<$crate::port::mode::Output>;
-        type MISO = $misomod::$MISO<$crate::port::mode::Input<$crate::port::mode::PullUp>>;
+        type MISO<InputMode> = $misomod::$MISO<$crate::port::mode::Input<InputMode>>;
 
         /// Behavior for a SPI interface.
         ///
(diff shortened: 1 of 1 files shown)


Rahix:
Looks good!  One thing I'd do differently:  I would create two different constructors to make it more obvious that a pull-up is needed here.  E.g.

```rust
fn new(... MISO<PullUp> ...) {}
fn with_external_pullup(... MISO<Floating> ...) {}
```

This way, when using `::new()` everything is configured correctly from the get-go.  And for the non-standard case you have to be explicit by using another method, which hopefully makes a programmer think twice about what they're doing.  With just a single constructor, this is not as obvious in my opinion.

couchand:
>  I would create two different constructors to make it more obvious that a pull-up is needed here.

That makes a lot of sense.  Will update.

couchand:
Updated, please take a look.

Rahix (approved):
Exactly what I was looking for, thanks!

Issue #16, opened 2019-10-24
Open drain behaviour simulation


rumatoest:
Hello.

As I can see it is not possible to have Pin that implements InputPin + OutputPin trait at the same time.
This happens because pin it consumes itself while mode change. Thus implement open/drain behavior is extremely hard in 3rd party code.

But the thing is that Arduino have modules that  communicates via single wire with pull up resistor in both direction (send/receive data). From Arduino libraries you can see that they try to emulate open drain behavior by switching pin to input mode when it is needed to set pin high or read state, and only to set pin low it is switched into output mode.

It would be really nice to have such behavior here with  InputPin + OutputPin traits implementation at the same time.

Rahix:
I think it should be possible to add an additional [mode](https://github.com/Rahix/avr-hal/blob/master/avr-hal-generic/src/port.rs#L4) where the pin type does exactly what you describe.  Though I'm not sure what the semantics should be when e.g. setting the pin low and _then_ reading.  This would 'toggle' the state implicitly and this could be confusing ...

jonahbron:
What about another pin struct that implements both traits, that can internally switch itself?  Something roughly like this (likely full of mistakes)
```
enum InputOrOutputPin {
    Output(OutputPin),
    Input(InputPin),
}
struct InputOutputPin {
    pin: InputOrOutputPin
}

impl InputOutputPin {
    def get_output(&mut self) -> OutputPin {
        match self.pin {
            Output(output_pin) => {
                output_pin
            },
            Input(input_pin) => {
                self.pin = Output(input_pin.release().into_output())
            },
        }
    }
    def get_input(&mut self) -> InputPin {
        match self.pin {
            Input(input_pin) => {
                input_pin
            },
            Output(output_pin) => {
                self.pin = Input(output.release().into_input())
            },
        }
    }
}

impl OutputPin for InputOutputPin {
    def set_high(&mut self) -> () {
        self.get_output().set_high()
    }
    def set_low(&mut self) -> () {
        self.get_output().set_low()
    }
}
impl InputPin for InputOutputPin {
    def is_high(&mut self) -> () {
        self.get_input().is_high()
    }
    def is_low(&mut self) -> () {
        self.get_input().is_low()
    }
}
```

kallemooo:
The proposed InputOutputPin will not work with the one-wire-bus library as it expects the set_high() to release the bus.
The InputOutputPin approach will force the the pin to high state (PushPull).
[...]

PR #51, opened 2020-08-17
generic: port: Support for open-drain mode

kallemooo:
Solution for issue #16.

A new port mode (OpenDrain) that Suports open-drain mode with external pull-up.
The solution uses the DDRxn to switch between a high state with the pin
input floating mode where the external pull-up is setting the pin to high state,
and low state by setting the pin in output low to pull the pin low.

Input traits is reading the external electrical state independent of if
the pin is set high or low.

Signed-off-by: Karl Thorén <karl.h.thoren@gmail.com>

### DIFF
--- a/avr-hal-generic/src/port.rs
+++ b/avr-hal-generic/src/port.rs
@@ -48,6 +48,9 @@
 //!
 //! // Convert into output
 //! let pd2: PD2<mode::Output> = pd2.into_output(&mut portd.ddr);
+//!
+//! // Convert into open drain input and output.
+//! let pd2: PD2<mode::OpenDrain> = pd2.into_open_drain(&mut portd.ddr);
 //! ```
 //!
 //! ### Digital Input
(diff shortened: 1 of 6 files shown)


avr-hal-generic/src/port.rs
```diff
@@ -257,6 +281,66 @@ macro_rules! impl_generic_pin {
...
+
+                fn set_low(&mut self) -> Result<(), Self::Error> {
+                    match self {
+                        $(
+                            $GenericPin::$PortEnum(i, _) => unsafe {
+                                (*<$PORTX>::ptr())
+                                    .$reg_ddr
+                                    .modify(|r, w| {
+                                        w.bits(r.bits() | (1 << *i))
+                                    })
```
Rahix, line 311: Minor nit: Can you match the style here with what you did further down?  I'd think it is a bit more readable as it is more compact:

```rust
(*<$PORTX>::ptr()).$reg_port.modify(|r, w| {
    w.bits(r.bits() & !(1 << $i))
});
```

kallemooo: I will fix that.

Rahix (approved):
Looks good!

PR #55, opened 2020-08-18
Re-export interrupt for convenience

couchand:
Note that this re-exports both the attribute macro (what I was looking for) and the module (probably okay, but not what I had intended).

If you think this is too niche to re-export, feel free to close.


Rahix:
Hmm, I intentionally left out this re-export as I think the interrupt API design is not yet mature enough to be a part of the HAL.  There are a few issues with it as it is currently, for example, it depends on an unstable ABI and the macro needs the MCU name to resolve vectors ... I have ideas for solving this at some point, but we're not yet.

So for the time being, it is available via `avr-device`, but not 'advertised' in the HAL.  I.e. people can use it, but there aren't any safe and ergonomic HAL abstractions to make it easy.

couchand:
Fair enough!

PR #57, opened 2020-08-19
Rename OpenDrain to TriState and update comments

couchand:
Commit e00abb6c introduced a Mode that uses DDRx to toggle the pin
state between driving low and floating high (presumably using an
external pull-up).  This commit renames that mode to TriState.

Calling this mode OpenDrain suggests that it would be reasonable for
the external pull-up to connect to any rail above GND, but this is not
generally valid for the AVR I/O cell.  This output can be safely
pulled-up only as far as VCC + 0.5 before the protection diode goes
into conducting mode.

### DIFF
--- a/avr-hal-generic/src/port.rs
+++ b/avr-hal-generic/src/port.rs
@@ -49,8 +49,8 @@
 //! // Convert into output
 //! let pd2: PD2<mode::Output> = pd2.into_output(&mut portd.ddr);
 //!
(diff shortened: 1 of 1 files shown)


Rahix:
cc @kallemooo, @rumatoest

kallemooo:
The description is clearer now what it do, and the limitations.
I think this change is ok.

PR #58, opened 2020-08-19
Add pwm support for mega2560

sepotvin:
Add support for the pwm (6 timers) of atmega2560 chips and mega2560 board. It also fixes a typo in the atmega328p which was disconnecting the wrong port.

### DIFF
--- a/boards/arduino-mega2560/src/lib.rs
+++ b/boards/arduino-mega2560/src/lib.rs
@@ -32,12 +32,12 @@ pub type I2c<M> = hal::i2c::I2c<hal::clock::MHz16, M>;
 ///     dp.PORTJ, dp.PORTK, dp.PORTL,
 /// );
 ///
-/// let mut timer1 = arduino_mega2560::pwm::Timer1Pwm::new(
-///     dp.TC1,
+/// let mut timer0 = arduino_mega2560::pwm::Timer0Pwm::new(
+///     dp.TC0,
 ///     arduino_mega2560::pwm::Prescaler::Prescale64,
 /// );
 ///
-/// let mut pin = pins.b7.into_output(&mut pins.ddr).into_pwm(&mut timer1);
+/// let mut pin = pins.d13.into_output(&mut pins.ddr).into_pwm(&mut timer0);
 ///
 /// pin.set_duty(128);
 /// pin.enable();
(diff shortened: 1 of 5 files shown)


chips/atmega2560-hal/src/pwm.rs
```diff
@@ -0,0 +1,363 @@
...
+                },
+            },
+        },
+    }
+}
+
+avr_hal::impl_pwm! {
+    /// Use `TC1` for PWM (pins `PB5`, `PB6`)
+    ///
+    /// Note: `PB7` is not currently supported (enter in conflict with `PB7` from timer0).
```
Rahix, line 91: You can support one pin from multiple timers; take a look at `atmega32u4-hal` where this is done, for example.  You'll have one method `.into_pwm()` for the default timer that you think most people would want to use and a secondary `.into_pwm1()` for the alternative (or `.into_pwm(whatever timer number)()`).

Please also mention this in the overview table as is done in `atmega32u4-hal`.

Rahix (commented):
Hey, thanks a lot for working on this!  I have one minor thing I could see in the diff, I left a comment.  I don't have this board on hand to test whether everything works, so I'll have to ask: Did you give all pins a quick try whether the waveform looks okay?

chips/atmega2560-hal/src/pwm.rs
sepotvin: Thanks for the pointer, I'll have a look and submit an update

boards/arduino-mega2560/src/lib.rs
```diff
@@ -17,3 +17,53 @@ pub use crate::pins::*;
...
+/// | `PE4` | `.into_pwm(&mut timer3)` |
+/// | `PE5` | `.into_pwm(&mut timer3)` |
+/// | `PG5` | `.into_pwm(&mut timer0)` |
+/// | `PH3` | `.into_pwm(&mut timer4)` |
+/// | `PH4` | `.into_pwm(&mut timer4)` |
+/// | `PH5` | `.into_pwm(&mut timer4)` |
+/// | `PH6` | `.into_pwm(&mut timer2)` |
+/// | `PL3` | `.into_pwm(&mut timer5)` |
+/// | `PL4` | `.into_pwm(&mut timer5)` |
+/// | `PL5` | `.into_pwm(&mut timer5)` |
```
couchand, line 64: Perhaps here the board pin numbers should be added or used instead?

Rahix: For Arduino Uno & Leonardo, I have created two versions of this table:  In the HAL crate (`atmega328p-hal`/`atmega32u4-hal`) the table uses MCU pin names and in the BSP crate (`arduino-uno`/`arduino-leonardo`) is uses board pin names.  I would suggest doing the same here.

Rahix (approved):
Looks good! I've taken the `ATmega328P` patch and applied that separately already.

PR #61, opened 2020-08-20
generic: delay: Account for the modulo 4096 part

sepotvin:
This fixes small delays (less than 4096us) and all delays not an exact multiple of 4096 us.


Rahix (approved):
Makes sense, thanks!  Out of curiosity, did you do measurements?  Do you know how big the overhead caused by the calculations and looping is?

sepotvin:
I didn't do any precise measurements (office is closed due to the pandemic and I don't have a scope at home). I noticed the problem because I had a loop generating a 500Hz square wave to a buzzer that was running so fast that no sound was generated.

Rahix:
Alright, good catch anyway!

PR #65, opened 2020-08-23
Implemented Watchdog(En/Dis) for ATmega328P

jkeli:
Hello, this is an implementation of the watchdog timer peripheral for the atmega328p. I didn't see another watchdog implementation to base this one on, so I hope I'm not too far off the mark. Thanks!

### DIFF
--- a/avr-hal-generic/src/wdt.rs
+++ b/avr-hal-generic/src/wdt.rs
@@ -1,9 +1,5 @@
 //! WDT Implementation
 
-
-pub use embedded_hal::watchdog;
-
-
 /// Implement traits for the watchdog interface
 #[macro_export]
 macro_rules! impl_wdt {
@@ -29,6 +25,8 @@ macro_rules! impl_wdt {
             peripheral: $WDT:ty,
         }
     ) => {
+        use $crate::hal::watchdog::*;
+
         /// Approximate length of the time-out period before the watchdog provides a system reset.
         ///
         /// After enabling the watchdog timer, call [`Watchdog::feed`] before the period ends to prevent a
--- a/avr-hal-generic/src/wdt.rs
+++ b/avr-hal-generic/src/wdt.rs
@@ -4,10 +4,12 @@ pub use embedded_hal::watchdog;
 
 /// Approximate length of the time-out period before the watchdog provides a system reset.
 ///
-/// After enabling the watchdog timer, call [`feed`](trait.Watchdog.html#tymethod.feed) before the
-/// period ends to prevent a reset.
+/// After enabling the watchdog timer, call [`Watchdog::feed`] before the period ends to prevent a
+/// reset.
 ///
 /// Some AVR models do not support all period lengths.
+///
+/// [`Watchdog::feed`]: watchdog/trait.Watchdog.html#tymethod.feed
 #[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
 pub enum WatchdogTimeOutPeriod {
     /// 16 milliseconds
--- a/avr-hal-generic/src/wdt.rs
+++ b/avr-hal-generic/src/wdt.rs
@@ -29,8 +29,8 @@ macro_rules! impl_wdt {
 
         /// Approximate length of the time-out period before the watchdog provides a system reset.
         ///
-        /// After enabling the watchdog timer, call [`Watchdog::feed`] before the period ends to prevent a
-        /// reset.
+        /// After enabling the watchdog timer, call [`Watchdog::feed`] before the period ends to
+        /// prevent a reset.
         ///
         /// [`Watchdog::feed`]: watchdog/trait.Watchdog.html#tymethod.feed
         #[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
@@ -41,10 +41,10 @@ macro_rules! impl_wdt {
         /// Provides a system reset when a counter reaches a given time-out value.
         ///
         /// # Note
-        /// Changing the watchdog configuration requires two separate writes to WDTCSR where the second
-        /// write must occur within 4 cycles of the first or the configuration will not change. You may need
-        /// to adjust optimization settings to prevent other operations from being emitted between these two
-        /// writes.
+        /// Changing the watchdog configuration requires two separate writes to WDTCSR where the
+        /// second write must occur within 4 cycles of the first or the configuration will not
+        /// change. You may need to adjust optimization settings to prevent other operations from
+        /// being emitted between these two writes.
         ///
         /// # Example
         /// ```
--- a/avr-hal-generic/src/wdt.rs
+++ b/avr-hal-generic/src/wdt.rs
@@ -1,0 +1,141 @@
+//! WDT Implementation
+
+pub use embedded_hal::watchdog;
+
+/// Approximate length of the time-out period before the watchdog provides a system reset.
+///
+/// After enabling the watchdog timer, call [`feed`](trait.Watchdog.html#tymethod.feed) before the
+/// period ends to prevent a reset.
+///
+/// Some AVR models do not support all period lengths.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub enum WatchdogTimeOutPeriod {
+    /// 16 milliseconds
+    Ms16,
+    /// 32 milliseconds
+    Ms32,
+    /// 64 milliseconds
+    Ms64,
+    /// 125 milliseconds
+    Ms125,
+    /// 250 milliseconds
+    Ms250,
+    /// 500 milliseconds
+    Ms500,
+    /// 1 second
+    Ms1000,
+    /// 2 seconds
+    Ms2000,
+    /// 4 seconds
+    Ms4000,
+    /// 8 seconds
+    Ms8000,
+}
+
+/// Implement traits for the watchdog interface
+#[macro_export]
+macro_rules! impl_wdt {
+    (
+        $(#[$wdt_attr:meta])*
+        pub struct $Wdt:ident {
+            mcu_status_register: $MCUSR:ty,
+            peripheral: $WDT:ty,
+        }
+    ) => {
+        use avr_device::generic::Reg;
+        pub use avr_hal::hal::watchdog::*;
+
+        pub struct $Wdt {
+            peripheral: $WDT,
+        }
+
+        impl $Wdt {
+            /// Initializes a Wdt.
+            ///
+            /// If a prior reset was provided by the watchdog, the WDRF in MCUSR would be set, so
+            /// WDRF is also cleared to allow for re-enabling the watchdog.
+            pub fn new(mcu_status_register: &Reg<$MCUSR>, peripheral: $WDT) -> Self {
+                mcu_status_register.modify(|_, w| w.wdrf().clear_bit());
+                Wdt { peripheral }
+            }
+        }
+
+        impl WatchdogEnable for $Wdt {
+            type Time = WatchdogTimeOutPeriod;
+
+            fn start<T>(&mut self, period: T)
+            where
+                T: Into<Self::Time>,
+            {
+                // The sequence for changing time-out configuration is as follows:
+                //
+                //     1. In the same operation, write a logic one to the Watchdog change enable bit
+                //        (WDCE) and WDE. A logic one must be written to WDE regardless of the
+                //        previous value of the WDE bit.
+                //     2. Within the next four clock cycles, write the WDE and Watchdog prescaler
+                //        bits (WDP) as desired, but with the WDCE bit cleared. This must be done in
+                //        one operation.
+                avr_hal::avr_device::interrupt::free(|_| {
+                    // Reset the watchdog timer
+                    self.feed();
+                    // Enable watchdog configuration mode
+                    self.peripheral
+                        .wdtcsr
+                        .modify(|_, w| w.wdce().set_bit().wde().set_bit());
+                    // Enable watchdog and set interval
+                    self.peripheral.wdtcsr.write(|w| {
+                        (match period.into() {
+                            WatchdogTimeOutPeriod::Ms16 => w.wdpl().cycles_2k_512k(),
+                            WatchdogTimeOutPeriod::Ms32 => w.wdpl().cycles_4k_1024k(),
+                            WatchdogTimeOutPeriod::Ms64 => w.wdpl().cycles_8k(),
+                            WatchdogTimeOutPeriod::Ms125 => w.wdpl().cycles_16k(),
+                            WatchdogTimeOutPeriod::Ms250 => w.wdpl().cycles_32k(),
+                            WatchdogTimeOutPeriod::Ms500 => w.wdpl().cycles_64k(),
+                            WatchdogTimeOutPeriod::Ms1000 => w.wdpl().cycles_128k(),
+                            WatchdogTimeOutPeriod::Ms2000 => w.wdpl().cycles_256k(),
+                            WatchdogTimeOutPeriod::Ms4000 => {
+                                w.wdpl().cycles_2k_512k().wdph().set_bit()
+                            }
+                            WatchdogTimeOutPeriod::Ms8000 => {
+                                w.wdpl().cycles_4k_1024k().wdph().set_bit()
+                            }
+                        })
+                        .wde()
+                        .set_bit()
+                        .wdce()
+                        .clear_bit()
+                    });
+                });
+            }
+        }
+
+        impl Watchdog for $Wdt {
+            #[inline]
+            fn feed(&mut self) {
+                avr_device::asm::wdr();
+            }
+        }
+
+        impl WatchdogDisable for $Wdt {
+            fn disable(&mut self) {
+                // The sequence for clearing WDE is as follows:
+                //
+                //     1. In the same operation, write a logic one to the Watchdog change enable bit
+                //        (WDCE) and WDE. A logic one must be written to WDE regardless of the
+                //        previous value of the WDE bit.
+                //     2. Within the next four clock cycles, clear the WDE and WDCE bits.
+                //        This must be done in one operation.
+                avr_hal::avr_device::interrupt::free(|_| {
+                    // Reset the watchdog timer
+                    self.feed();
+                    // Enable watchdog configuration mode
+                    self.peripheral
+                        .wdtcsr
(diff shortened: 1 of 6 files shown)


lights0123:
Although I wish there was a way to only pass ownership of one bit, can you take [`Reg<atmega328p::cpu::mcusr::MCUSR_SPEC>`](https://rahix.github.io/avr-hal/atmega328p_hal/atmega328p/cpu/struct.RegisterBlock.html#structfield.mcusr) instead? That way, all the other registers in the CPU block are still available for other purposes.

chips/atmega328p-hal/src/wdt.rs
```diff
@@ -0,0 +1,108 @@
...
+                .set_bit()
+            });
+        });
+    }
+}
+
+impl<'wdt> Watchdog for Wdt<'wdt> {
+    #[inline]
+    fn feed(&mut self) {
+        unsafe { llvm_asm!("WDR") }
```
Rahix, line 88: Please use `avr_device::asm::wdr()` instead.

chips/atmega328p-hal/src/wdt.rs
```diff
@@ -0,0 +1,108 @@
...
+
+impl<'wdt> WatchdogEnable for Wdt<'wdt> {
+    type Time = WatchdogTimeOutPeriod;
+
+    fn start<T>(&mut self, period: T)
+    where
+        T: Into<Self::Time>,
+    {
+        // Disable interrupts while starting the watchdog timer
+        avr_hal::avr_device::interrupt::free(|_| {
```
Rahix, line 55: Can you elaborate why this is necessary?

chips/atmega328p-hal/src/wdt.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+pub struct Wdt<'wdt> {
+    mcu_status_register: &'wdt Reg<crate::atmega328p::cpu::mcusr::MCUSR_SPEC>,
+    peripheral: crate::atmega328p::WDT,
+}
+
+impl<'wdt> Wdt<'wdt> {
+    pub fn new(
+        mcu_status_register: &'wdt Reg<crate::atmega328p::cpu::mcusr::MCUSR_SPEC>,
```
Rahix, line 37: This will prevent moving `Peripherals.CPU` anyway so I'd just take a reference to the entire peripheral here (`&'wdt crate::atmega328p::CPU`).

chips/atmega328p-hal/src/wdt.rs
```diff
@@ -0,0 +1,108 @@
...
+//!
+//! loop {
+//!     watchdog.feed();
+//! }
+//! ```
+
+use avr_device::generic::Reg;
+use avr_hal::hal::watchdog::*;
+
+pub enum WatchdogTimeOutPeriod {
```
Rahix, line 17: `struct`s and `enum`s should eagerly derive common traits ([C-COMMON-TRAITS](https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits)).

Rahix (commented):
Hey, thanks a lot for this PR!

Usually, to keep code-duplication manageable, we create a macro in `avr-hal-generic` containing the actual implementation with just the chip-specific parts being passed in as macro inputs.  I think the same would make sense here, maybe you can take a look at some of the other peripherals to see how they do things?

chips/atmega328p-hal/src/wdt.rs
jkeli: Interrupts were disabled in the code samples in the data sheet. My understanding is that it is because the watchdog configuration must be changed within 4 cycles of enabling changes (writing 1 to WDE and WDCE).

Rahix: Ok, makes sense!  Please document this a bit more in a comment, so any future reader knows immediately what is going on.  Maybe this text from the data-sheet is enough explanation?

> The sequence for clearing WDE and changing time-out configuration is as follows:
> 1. In the same operation, write a logic one to the Watchdog change enable bit (WDCE) and WDE. A logic one must be written to WDE regardless of the previous value of the WDE bit.
> 2. Within the next four clock cycles, write the WDE and Watchdog prescaler bits (WDP) as desired, but with the WDCE bit cleared. This must be done in one operation.

chips/atmega328p-hal/src/wdt.rs
```diff
@@ -0,0 +1,108 @@
...
+                    WatchdogTimeOutPeriod::Ms125 => w.wdpl().cycles_16k(),
+                    WatchdogTimeOutPeriod::Ms250 => w.wdpl().cycles_32k(),
+                    WatchdogTimeOutPeriod::Ms500 => w.wdpl().cycles_64k(),
+                    WatchdogTimeOutPeriod::Ms1000 => w.wdpl().cycles_128k(),
+                    WatchdogTimeOutPeriod::Ms2000 => w.wdpl().cycles_256k(),
+                    WatchdogTimeOutPeriod::Ms4000 => w.wdpl().cycles_2k_512k().wdph().set_bit(),
+                    WatchdogTimeOutPeriod::Ms8000 => w.wdpl().cycles_4k_1024k().wdph().set_bit(),
+                })
+                .wde()
+                .set_bit()
```
Rahix, line 79: It's not strictly necessary but for clarity, I'd explicitly set `WDCE` to zero here as well.  I.e. add

```rust
.wdce().clear_bit()
```

chips/atmega328p-hal/src/wdt.rs
jkeli: Do you think it'd be reasonable to clear the WDRF in the MCUSR in `new` instead of clearing it in enable/disable (and only holding onto MCUSR until `new` returns)? It does seem like moving `Peripherals.CPU` into `Wdt` only to reset the flag shouldn't be necessary. Destroying the `Wdt` instance after enabling the watchdog doesn't seem possible since `feed` must be called within each time-out period.

Rahix: > Do you think it'd be reasonable to clear the WDRF in the MCUSR in new instead of clearing it in enable/disable (and only holding onto MCUSR until new returns)?

I've read through the datasheet a bit more and I think this should be ok.  The interactions aren't entirely obvious but unless I'm mistaken, it works like this:

The `WDRF` has two purposes:

1. It signals a starting system that reset was caused by the watchdog.  This is in line with all the other bit-fields in `MCUSR`.
2. It acts as a safeguard to clearing `WDE`.  The idea seems to be that a system which was reset by the watchdog has to 'acknowledge' the watchdog-reset (by clearing `WDRF`) before it can turn off the watchdog.

With both, `WDRF` only becomes relevant in a failure condition and the following restart.  That means, clearing `WDRF` once at startup should be enough as it will only become set when a reset occurs.

This would get rid of the borrowed/owned `CPU` peripheral so it seems to me that this would be the favorable aproach.

jkeli: Thank you, much appreciated.

jkeli:
Will convert the current implementation into a macro.

avr-hal-generic/src/wdt.rs
```diff
@@ -0,0 +1,141 @@
...
+        $(#[$wdt_attr:meta])*
+        pub struct $Wdt:ident {
+            mcu_status_register: $MCUSR:ty,
+            peripheral: $WDT:ty,
+        }
+    ) => {
+        use avr_device::generic::Reg;
+        pub use avr_hal::hal::watchdog::*;
+
+        pub struct $Wdt {
```
Rahix, line 48: As far as I can see, the `$wdt_attr` is unused right now.  Add it to the struct like this:

```suggestion
        $(#[$wdt_attr])*
        pub struct $Wdt {
```

_but_, on the other hand, I'm thinking that it might be more sensible to directly document the `$Wdt` struct here because documentation should be the same for all watchdogs.  Or do you see an issue with that?

avr-hal-generic/src/wdt.rs
```diff
@@ -0,0 +1,141 @@
...
+            peripheral: $WDT,
+        }
+
+        impl $Wdt {
+            /// Initializes a Wdt.
+            ///
+            /// If a prior reset was provided by the watchdog, the WDRF in MCUSR would be set, so
+            /// WDRF is also cleared to allow for re-enabling the watchdog.
+            pub fn new(mcu_status_register: &Reg<$MCUSR>, peripheral: $WDT) -> Self {
```
Rahix, line 57: I would suggest using the type alias from `avr-device` instead of naming `Reg<>` here.  That means, I'd refer to [`avr_device::atmega328p::cpu::MCUSR`](https://docs.rs/avr-device/0.2.2/avr_device/atmega328p/cpu/type.MCUSR.html).

```suggestion
            pub fn new(mcu_status_register: &$MCUSR, peripheral: $WDT) -> Self {
```

boards/arduino-uno/examples/uno-watchdog.rs
```diff
@@ -0,0 +1,17 @@
...
+#![no_main]
+
+use arduino_uno::prelude::*;
+use panic_halt as _;
+
+#[arduino_uno::entry]
+fn main() -> ! {
+    let dp = arduino_uno::Peripherals::take().unwrap();
+    let mut watchdog = arduino_uno::wdt::Wdt::new(&dp.CPU.mcusr, dp.WDT);
+    watchdog.disable();
```
Rahix, line 11: Is `disable()` before `start()` necessary or is this just for demonstration?

chips/atmega328p-hal/src/wdt.rs
```diff
@@ -0,0 +1,29 @@
...
+//! let mut watchdog = board::wdt::Wdt::new(&dp.CPU.mcusr, dp.WDT);
+//! watchdog.disable();
+//! watchdog.start(board::wdt::WatchdogTimeOutPeriod::Ms8000);
+//!
+//! loop {
+//!     watchdog.feed();
+//! }
+//! ```
+
+extern crate avr_hal_generic as avr_hal;
```
Rahix, line 20: This line should not be necessary (`extern crate` is not needed in 2018 rust)

```suggestion
```

chips/atmega328p-hal/src/wdt.rs
```diff
@@ -0,0 +1,29 @@
...
+
+pub use avr_hal::wdt::*;
+
+avr_hal::impl_wdt! {
+    pub struct Wdt {
+        mcu_status_register: crate::atmega328p::cpu::mcusr::MCUSR_SPEC,
```
Rahix, line 26: As mentioned above, I would refer to the type alias for `MCUSR`:

```suggestion
        mcu_status_register: crate::atmega328p::cpu::MCUSR,
```

avr-hal-generic/src/wdt.rs
```diff
@@ -0,0 +1,141 @@
...
+macro_rules! impl_wdt {
+    (
+        $(#[$wdt_attr:meta])*
+        pub struct $Wdt:ident {
+            mcu_status_register: $MCUSR:ty,
+            peripheral: $WDT:ty,
+        }
+    ) => {
+        use avr_device::generic::Reg;
+        pub use avr_hal::hal::watchdog::*;
```
Rahix, line 46: This depends on the crate being imported as `avr_hal`.  In macros, one should usually refer to the own crate using `$crate` as that is guaranteed to always work.

That said, I think the imports heres should be removed; they would make it impossible to use this macro in a module that for some reason has these items already imported.  Instead, inside the macro, I suggest referring to everything by its full path.  So further down, e.g. refer to `WatchdogEnable` like this:

```rust
impl $crate::hal::watchdog::WatchdogEnable for $Wdt {
```

Rahix: (And `Reg` should no longer be necessary with the other suggested changes.)

avr-hal-generic/src/wdt.rs
```diff
@@ -0,0 +1,141 @@
+//! WDT Implementation
+
+pub use embedded_hal::watchdog;
```
Rahix, line 3: I think this import is unused?

Rahix (changes requested):
Thanks for putting in the work!  This will hopefully allow us to easily add watchdog support for all the other HALs and boards as well.  I went through your changes once more: overall it looks great; I just found a few more details that we should take a closer look at.  See my comments below.

avr-hal-generic/src/wdt.rs
```diff
@@ -0,0 +1,141 @@
...
+
+/// Approximate length of the time-out period before the watchdog provides a system reset.
+///
+/// After enabling the watchdog timer, call [`feed`](trait.Watchdog.html#tymethod.feed) before the
+/// period ends to prevent a reset.
+///
+/// Some AVR models do not support all period lengths.
```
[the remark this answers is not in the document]
jkeli: I have moved the timeout periods into the chip. I was hoping to use a simpler format for defining the available variants and their mappings to pre-scaler configuration bits, but from reviewing a couple of examples, it seems to require overly complicated macro rules to generate the enum. Instead, I split apart the Enum from the mapping, which are easy to parse separately.

avr-hal-generic/src/wdt.rs
jkeli: I chose to move the documentation into the macro.

boards/arduino-uno/examples/uno-watchdog.rs
jkeli: It sounds like it's to catch a (rare?) case where the watchdog is accidentally enabled. They suggest always disabling the watchdog on init, even if the watchdog isn't being used. From the data sheet:

> Note: If the watchdog is accidentally enabled, for example by a runaway pointer or brown-out condition, the
> device will be reset and the watchdog timer will stay enabled. If the code is not set up to handle the watchdog,
> this might lead to an eternal loop of time-out resets. To avoid this situation, the application software should
> always clear the watchdog system reset flag (WDRF) and the WDE control bit in the initialization routine, even
> if the watchdog is not in use.

avr-hal-generic/src/wdt.rs
jkeli: I think this is what allows the wdt module in the chip to bring the watchdog traits into scope.

```
//! WatchDog Timer Implementation

pub use avr_hal_generic::wdt::watchdog::*;
pub use avr_hal_generic::wdt::*;
```

avr-hal-generic/src/wdt.rs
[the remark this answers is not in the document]
Rahix: Okay, this makes sense.  Can we change it so you only have to spell out the variants once, though?  E.g. something similar to this:

```rust
avr_hal::impl_wdt! {
    pub enum WatchdogTimeOutPeriod {
        /// 16 milliseconds
        Ms16 = wdpl().cycles_2k_512k(),
        /// 32 milliseconds
        Ms32 = wdpl().cycles_8k(),
        /// 64 milliseconds
        Ms64 = wdpl().cycles_16k(),
        // ...
    }

    pub struct Wdt {
        mcu_status_register: crate::atmega328p::cpu::MCUSR,
        peripheral: crate::atmega328p::WDT,
    }
}
```

boards/arduino-uno/examples/uno-watchdog.rs
Rahix: Hm, I'm reading this a bit differently:  When you're not using the watchdog, you're supposed to disable it in your init code anyway, just in case it got accidentally enabled.  Otherwise it will stay enabled and cause an eternal loop of resets.  When you're actively using the watchdog (as you do in the example), it's not needed because you're configuring and feeding it properly.

So to me this says e.g. a bootloader or runtime should clear WDRF and WDE on startup.  But to code using the watchdog as intended, it's not relevant.

avr-hal-generic/src/wdt.rs
Rahix: Do we still need them?  The macro is explicitly referring to the full path so I'd say the import is unnecessary.  If downstream code wants to name them, it should be preferred to import `embedded-hal` directly.

avr-hal-generic/src/wdt.rs
```diff
@@ -0,0 +1,126 @@
...
+        }
+    ) => {
+        /// Approximate length of the time-out period before the watchdog provides a system reset.
+        ///
+        /// After enabling the watchdog timer, call [`Watchdog::feed`] before the period ends to prevent a
+        /// reset.
+        ///
+        /// [`Watchdog::feed`]: watchdog/trait.Watchdog.html#tymethod.feed
+        #[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
+        pub enum WatchdogTimeOutPeriod $supported_periods
```
Rahix, line 26: One minor nit:  I suggest naming the enum just `Timeout` to keep it short.  Otherwise code referring to this type will be streched across the screen quite a bit for no real benefit.

avr-hal-generic/src/wdt.rs
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        impl $Wdt {
+            /// Initializes a Wdt.
+            ///
+            /// If a prior reset was provided by the watchdog, the WDRF in MCUSR would be set, so
+            /// WDRF is also cleared to allow for re-enabling the watchdog.
+            pub fn new(mcu_status_register: &$MCUSR, peripheral: $WDT) -> Self {
+                mcu_status_register.modify(|_, w| w.wdrf().clear_bit());
+                Wdt { peripheral }
```
Rahix, line 57: This needs to be `Self` or `$Wdt` to work as expected.

avr-hal-generic/src/wdt.rs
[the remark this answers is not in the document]
jkeli: Sure, I cleaned up a bit and it now looks like:

```rust
avr_hal::impl_wdt! {

    pub enum Timeout {
        /// 16 milliseconds
        Ms16 { wdpl().cycles_2k_512k() },
        /// 32 milliseconds
        Ms32 { wdpl().cycles_4k_1024k() },
        /// 64 milliseconds
        Ms64 { wdpl().cycles_8k() },
        /// ...
    }

    pub struct Wdt {
        mcu_status_register: crate::atmega328p::cpu::MCUSR,
        peripheral: crate::atmega328p::WDT,
    }
}
```

boards/arduino-uno/examples/uno-watchdog.rs
jkeli: I removed the disable() call from the examples.

Rahix (approved):
Looks great, thanks a lot!

jkeli:
Thanks for all the help!

Issue #63, opened 2020-08-23
trait embedded_hal::blocking::spi::write::Default<u8> is not implemented for arduino_uno


Gregoor:
Heya, me again!

I'm trying to glue some led strip driver code (apa102) to the arduino board code, which yielded the error:
> trait `embedded_hal::blocking::spi::write::Default<u8>` is not implemented for `arduino_uno::spi::Spi<arduino_uno::atmega328p_hal::port::
mode::PullUp>

When I asked in Gitter it was suggested that this is a temporary hole, I could try to plug it. I assume it'd be a matter of pulling this:
https://github.com/rust-embedded/embedded-hal/blob/d81cf7cdb36d2214136891c7a6e0d45f9fe04723/src/blocking/spi.rs#L59
into avr-hal-generic::spi?

Rahix:
Implemented in commit cd5dd15c45ec297315f0d62d56e3fa044a24e8df.

Gregoor:
Looks like I was too slow! But at least the second impl is identical to what I had. Though I didn't grok what MisoInputMode was yet. Thanks for the fix!

PR #66, opened 2020-08-29
Add default implementation of trasfer and write for spi

no111u3:
Follow by #63.


Rahix (approved):
Thanks for taking care of this!

Fixes #66.

PR #68, opened 2020-09-05
generic: delay: Reduce overhead of u32 delay_us

amikhalev:
The range iterator was compiling to pretty ugly code with stack spilling,
so simplify by using a C-style for loop.

### DIFF
--- a/avr-hal-generic/src/delay.rs
+++ b/avr-hal-generic/src/delay.rs
@@ -214,9 +214,11 @@ where
     Delay<SPEED>: delay::DelayUs<u16>,
 {
     fn delay_us(&mut self, us: u32) {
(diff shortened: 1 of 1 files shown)


Rahix (commented):
Nice work, thanks a lot!  Though please leave the comment in for now, I'd still like to revisit this at some point and do precise measurements of how much overhead will remain and see how we could mitigate that for (hopefully) cycle accurate delays.

amikhalev:
I added the comment back in. For what it's worth, I measured the output of `uno-blink` on a 50MHz scope and the period looks pretty close to 200ms (I measured 199.8ms). Previous to the change the period measures 201.0ms.

Rahix (approved):
That's already much better than I would have expected!  Thanks a lot!

PR #72, opened 2020-09-18
Fix max-atomic-width target setting

Rahix:
Nightly rustc throws compiler errors since upstream commit [3be40b22c834 ("Fix AtomicPtr::from_mut align check: Avoid generic arg in const expr.")][1].

They seem to be related to some discrepancies in the alignment requirements for pointer types on AVR.  While this should probably be investigated at some point, for now let's set `max-atomic-width` to 8 to make everything build again but still retain at least some atomic types (those aren't actual atomics but intrinsics that are lowered to critical sections for atomic access).

[1]: https://github.com/rust-lang/rust/commit/3be40b22c83491e4a4d34e2ad06aa81b804b9fad


abusch:
HI, FYI I opened this bug recently about this problem: https://github.com/rust-lang/rust/issues/76833

PR #62, opened 2020-08-22
SparkFun Qwiic Pro Micro USB-C Atmega32u4

cecton:
Hello @Rahix 

I just started making the sparkfun-pro-micro crate by copying and renaming everything from the arduino-leonardo one (26698a0a191).

> @cecton, for the pins, just reference the [schematic for your board](https://cdn.sparkfun.com/assets/4/4/f/2/a/Qwiic_Pro_Micro_V2_0_USB_C_Schematic.pdf). It shows quite well which MCU pin is named what on the PCB.

It honestly took me some time to see it on the schematics :sweat_smile: I used the right most diagram.

It doesn't really show D11 to D13 but I supposed this is assumed because of the order? So I did the same with A4-A5 (Please double check that b66921e87)

Closes #52

PR #73, opened 2020-09-19
Add port F for Arduino Leonardo

cecton:
Related to https://github.com/Rahix/avr-hal/pull/62#issuecomment-689419628

### DIFF
--- a/boards/arduino-leonardo/src/pins.rs
+++ b/boards/arduino-leonardo/src/pins.rs
@@ -16,16 +16,28 @@ avr_hal_generic::impl_board_pins! {
     /// Reexport of the Leonardo's pins, with the names they have on the board
     pub struct Pins {
         /// `A0`
+        ///
+        /// * `ADC7` channel
         pub a0: portf::pf7::PF7,
         /// `A1`
+        ///
+        /// * `ADC6` channel
         pub a1: portf::pf6::PF6,
         /// `A2`
+        ///
+        /// * `ADC5` channel
         pub a2: portf::pf5::PF5,
         /// `A3`
+        ///
+        /// * `ADC4` channel
         pub a3: portf::pf4::PF4,
         /// `A4`
+        ///
+        /// * `ADC1` channel
         pub a4: portf::pf1::PF1,
         /// `A5`
+        ///
+        /// * `ADC0` channel
         pub a5: portf::pf0::PF0,
         /// `D0` / `RX`
         ///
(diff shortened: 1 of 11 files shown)


boards/arduino-leonardo/src/pins.rs
```diff
@@ -10,12 +10,23 @@ avr_hal_generic::impl_board_pins! {
...
+        /// `A1`
+        pub a1: portf::pf6::PF6,
+        /// `A2`
+        pub a2: portf::pf5::PF5,
+        /// `A3`
+        pub a3: portf::pf4::PF4,
+        /// `A4`
+        pub a4: portf::pf1::PF1,
+        /// `A5`
+        pub a5: portf::pf0::PF0,
```
Rahix, line 29: Please also document the alternate functions for these pins.  I'd also move this block to the end, so the digital pins are first in the documentation.

```suggestion
        /// `A0`
        ///
        /// * `ADC7` channel
        pub a0: portf::pf7::PF7,
        /// `A1`
        ///
        /// * `ADC6` channel
        pub a1: portf::pf6::PF6,
        /// `A2`
        ///
        /// * `ADC5` channel
        pub a2: portf::pf5::PF5,
        /// `A3`
        ///
        /// * `ADC4` channel
        pub a3: portf::pf4::PF4,
        /// `A4`
        ///
        /// * `ADC1` channel
        pub a4: portf::pf1::PF1,
        /// `A5`
        ///
        /// * `ADC0` channel
        pub a5: portf::pf0::PF0,
```

boards/sparkfun-pro-micro/examples/pro-micro-adc.rs
```diff
@@ -19,7 +20,7 @@ fn main() -> ! {
 
     ufmt::uwriteln!(&mut serial, "Reading analog inputs ...\r").void_unwrap();
 
-    let mut adc = sparkfun_pro_micro::adc::Adc::new(dp.ADC, Default::default());
+    let mut adc = adc::Adc::new(dp.ADC, Default::default());
```
Rahix, line 23: Let's keep those changes out of this MR and do them later (if at all) once we have decided what to do with the crate.

Rahix (changes requested):
Thanks a lot!  Minor nitpick in the comment below

boards/sparkfun-pro-micro/examples/pro-micro-adc.rs
cecton: done

boards/arduino-leonardo/src/pins.rs
cecton: done & done (ADCx are the alternate functions?)

Rahix: Correct, by default the pins are Digital I/O, and they can be reconfigured as ADC channels.

Rahix (approved):

PR #74, opened 2020-09-19
Fixes for pro micro

cecton:
> I also realized that I might have added too many pins. i.e. there are no pin A4 and A5 that I can see on the board! I probably made a mistake.

If there is no pin but there is still ACD1 and ACD0 on the schematics. Does that mean it still have a use? (For the alternate function --~~I'm not sure what's the alternate function~~ ok I think I know--) Then, should we keep them anyway?


cecton:
(I'm not sure if the failing example is valid on the pro micro. It seems it lacks the necessary pins)

Rahix:
> (I'm not sure if the failing example is valid on the pro micro. It seems it lacks the necessary pins)

If the pro micro has an onboard LED you should use that pin instead.  If not, any pin you like will do ...

Rahix:
Can you do this change so I can merge the PR? :)

cecton:
@Rahix can you check and let me know if the pin I chose make sense?

Rahix (approved):
Looks good, thank you very much!

Issue #32, opened 2020-08-03
Cleanup target files


Rahix:
Notes to myself about what needs to happen:

- [x] Move all target files into a single top-level directory.  Update all `.cargo/config.toml`s to point to the files in this place instead.  ->  This gets rid of the symlinks thus fixing #29.
- [x] Do [this cleanup](https://github.com/avr-rust/blink/pull/20) to all our target files as well.
- [ ] Check remaining settings against current upstream defaults (after https://github.com/rust-lang/rust/pull/74941)
- [x] Think about maybe adding infrastructure for automatically deriving the target files from upstream defaults to reduce maintenance burden.

explicite:
@Rahix  can you please mark second checkbox?

explicite:
Looks line in third one we have some additional setting which I cannot add find in our targets. 
https://github.com/rust-lang/rust/pull/74941/files#diff-e26bdcc4844624b6b557a83e64e20c0a
I cannot see  e.g `eh_frame_header` or `max-atomic-width`. Should we support all old and new options or remove legacy from our targets? Please find bellow all merged.
```
{
  "arch": "avr",
  "data-layout": "e-P1-p:16:8-i8:8-i16:8-i32:8-i64:8-f32:8-f64:8-n8-a:8",
  "llvm-target": "avr-unknown-unknown",
  "target-endian": "little",
  "target-pointer-width": "16",
  "linker-flavor": "gcc",
  "os": "unknown",
  "target-env": "",
  "target-vendor": "unknown",
  "target-c-int-width": "16",
  "cpu": "atmega2560",
  "exe-suffix": ".elf",
  "linker": "avr-gcc",
  "dynamic_linking": false,
  "executables": true,
  "linker_is_gnu": true,
  "has_rpath": false,
  "position_independent_executables": false,
  "eh_frame_header": false,
  "max-atomic-width": 8,
  "pre-link-args": {
    "gcc": [
      "-mmcu=atmega2560"
    ]
  },
  
  "post-link-args": {
    "gcc": [
      "-Wl,--gc-sections"
    ]
  },
  "singlethread": false,
  "no-builtins": false,
  "no-default-libraries": false,
  "eh-frame-header": false
}
```

Rahix:
This command shows the current upstream default target:

```bash
rustc --print target-spec-json -Z unstable-options --target avr-unknown-gnu-atmega328
```

Rahix:
[...]

PR #78, opened 2020-10-02
Target files cleanup

explicite:
Second point from #32
- rm `-Os` from  pre link args
- json fmt


Rahix (changes requested):
Oh, nice!  Didn't think anyone would want to take care of that ^^

One minor nit: Please add back the newline at end of file because that's what most editors will do (-> the next edit would produce an unnecessary diff change).

Rahix (approved):

Rahix:
Thanks!

PR #81, opened 2020-10-06
Add atmega328pb

jkristell:
WIP hal for atmega328pb

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ members = [
     "chips/atmega168-hal",
     "chips/atmega2560-hal",
     "chips/atmega328p-hal",
+    "chips/atmega328pb-hal",
     "chips/atmega32u4-hal",
     "chips/atmega48p-hal",
     "chips/attiny85-hal",
--- a/avr-hal-generic/Cargo.toml
+++ b/avr-hal-generic/Cargo.toml
@@ -9,7 +9,9 @@ cfg-if = "0.1.7"
 nb = "0.1.2"
 ufmt = "0.1.0"
 paste = "1.0.0"
-avr-device = "0.2.2"
+
+[dependencies.avr-device]
+path = "../../avr-device"
 
 [dependencies.embedded-hal]
 version = "0.2.3"
--- a/chips/atmega328pb-hal/Cargo.toml
+++ b/chips/atmega328pb-hal/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "atmega328pb-hal"
+version = "0.1.0"
+authors = ["Johan Kristell <johan@jott.se>"]
+edition = "2018"
+
+[features]
+rt = ["avr-device/rt"]
+# Package exposes the ADC6 and ADC7 pins (only 32TQFP, 32MLF, 32UFBGA)
+adc-pins = []
+
+[dependencies]
+avr-hal-generic = { path = "../../avr-hal-generic/" }
+
+[dependencies.avr-device]
+path = "../../../avr-device"
+#version = "0.2.2"
+features = ["atmega328pb"]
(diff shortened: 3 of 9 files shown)


jkristell:
Issues:

- [ ] spi0 and spi1 are put in modules because of the way the spi macro use `type`
- [ ] Pwm for TC4 can't be constructed because pin pd2 is used by both timer 3 and 4, so we get overlapping into_pwm implementations.

Rahix:
>  spi0 and spi1 are put in modules because of the way the spi macro use type

That should definitely be fixed in the macro!

>  Pwm for TC4 can't be constructed because pin pd2 is used by both timer 3 and 4, so we get overlapping into_pwm implementations.

This is actually not a rare occurence.  For this, one of the two impls should name the method `.into_pwm1()` (replacing 1 with the timer number).  There should then be a table in the docs with all the details: E.g. https://rahix.github.io/avr-hal/atmega32u4_hal/pwm/index.html

>  Decide if this hal should be a extension to the atmega328p hal instead. Once the above issues are taken care of, we can decide if that would be a good idea.

I know too little about the 328PB so maybe you can tell a bit more: How do the two compare?  I'd be inclined to merge them if possible, as that reduces maintenance burden :)

jkristell:
> > spi0 and spi1 are put in modules because of the way the spi macro use type
> 
> That should definitely be fixed in the macro!

I could try to do that if you want.
 
> > Pwm for TC4 can't be constructed because pin pd2 is used by both timer 3 and 4, so we get overlapping into_pwm implementations.
> 
> This is actually not a rare occurence. For this, one of the two impls should name the method `.into_pwm1()` (replacing 1 with the timer number). There should then be a table in the docs with all the details: E.g. https://rahix.github.io/avr-hal/atmega32u4_hal/pwm/index.html

That was an easy fix, thanks for the hint.

> > Decide if this hal should be a extension to the atmega328p hal instead. Once the above issues are taken care of, we can decide if that would be a good idea.
> 
> I know too little about the 328PB so maybe you can tell a bit more: How do the two compare? I'd be inclined to merge them if possible, as that reduces maintenance burden :)

From a software point of view the atmega328pb is a superset of the 328/p. Additional twi, spi, usart and timers. And 
also 4 additional gpios on the new PORTE,. Electronically they differ as the 328pb use two pins that are for power on the 328/p as io-pins.

atmega328pb


AT15007: Differences between ATmega328/P and ATmega328PB
https://www.microchip.com/wwwAppNotes/AppNotes.aspx?appnote=en591418

Rahix:
> I could try to do that if you want.

That would be great!

> From a software point of view the atmega328pb is a superset of the 328/p. Additional twi, spi, usart and timers. And also 4 additional gpios on the new PORTE,. Electronically they differ as the 328pb use two pins that are for power on the 328/p as io-pins.

Okay, then I would suggest doing it in one crate, similar to the way the stm32-rs HALs manage it:  Mutually-exclusive features for the different chips with `atmega328p` being the default.

jkristell:
> > I could try to do that if you want.
> 
> That would be great!

The spi macro needs to updated anyway to be able to handle the different register and field names for the secondary spi peripheral of the atmega328pb. Would it be a good idea to base the macro of the i2c macro?

Rahix:
> Would it be a good idea to base the macro of the i2c macro?

If we stick with the current register naming scheme, I'd take usart (`avr-hal-generic/src/serial.rs`) as the reference.  I'm using the `paste!()` macro there to create the register and field identifiers.

That said, I'd hold off from working on that until we decide on the patching strategy.  I didn't have much time in the last days to look deeper into it unfortunately ... Though the code in the stm32-rs crates does look a lot nicer, so if "full patching" turns out to be feasible, that is definitely the way I would prefer :)

Rahix:
Closing this in favor of #96.

PR #82, opened 2020-10-06
#32 avr-spec

explicite:
What was done:

- rm symlink
- avr-specs creation
- .cargo build targets update
- .DS_store cleanup


Rahix (approved):
Looks good to me!

PR #84, opened 2020-10-07
avr-specs/sync-from-upstream.py implementation

explicite:
Last two points from #32 
- `avr-specs/sync-from-upstream.py` implementation
- update specs with latest upstream spec


explicite:
@Rahix how should I fix that?
1. Review target and update only args which allows to build project
2. Review other project parts and keep upstream target

Rahix:
I would design this a bit differently.  I'd load the upstream spec, then iterate over a list of supported MCUs and write out modified versions of the upstream spec.  In the python script, at the top I'd define the list of supported chips:

```python
SPECS = {
    "atmega32u4": {
        "cpu": "atmega32u4",
    },
    "atmega328p": {
        "cpu": "atmega328p",
    },
}
```

avr-specs/sync-from-upstream.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/python3
```
Rahix, line 1: The recommended shebang is `/usr/bin/env python3` because that composes better with virtualenv and the likes:

```suggestion
#!/usr/bin/env python3
```

avr-specs/sync-from-upstream.py
```diff
@@ -0,0 +1,64 @@
...
+import json
+import os
+
+ENCODING = 'utf-8'
+
+
+def main():
+    process = subprocess.Popen(['rustc', '--print', 'target-spec-json', '-Z', 'unstable-options', '--target', 'avr-unknown-gnu-atmega328'],
+                               stdout=subprocess.PIPE,
+                               stderr=subprocess.PIPE)
```
Rahix, line 13: This only works when `rustc` defaults to nightly.  I'd include a check first, to try whether `rustc` refers to a nightly version.

avr-specs/sync-from-upstream.py
```diff
@@ -0,0 +1,64 @@
...
+
+    ref_spec_string, stderr = process.communicate()
```
Rahix, line 15: Instead of using `Popen`, you should probably use [`subprocess.run()`](https://docs.python.org/3/library/subprocess.html#subprocess.run).  Also, leave stderr as it is so errors from the compiler are shown on the console.  And lastly, using `check=True` here makes sure that errors are caught and not ignored.

```suggestion
    ref_spec_string = subprocess.run(
        ['rustc', '--print', 'target-spec-json', '-Z', 'unstable-options', '--target', 'avr-unknown-gnu-atmega328'],
        check=True,
        stdout=subprocess.PIPE,
    ).stdout
```

avr-specs/sync-from-upstream.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/python3
+
+import subprocess
+import json
+import os
+
+ENCODING = 'utf-8'
```
Rahix, line 7: I don't think explicitly defining the encoding should be necessary?  AFAIK Python already chooses a sane default.

avr-specs/sync-from-upstream.py
```diff
@@ -0,0 +1,66 @@
...
+
+import subprocess
+import json
+import copy
+import os
+
+SPECS = {
+    "atmega32u4": {
+        "cpu": "atmega32u4",
+        "no-default-libraries": False,
```
explicite, line 11: looks like is required to pass

avr-specs/sync-from-upstream.py
explicite: removed

avr-specs/sync-from-upstream.py
explicite: done

avr-specs/sync-from-upstream.py
Rahix: Suggestion: Add a second global for the common changes:

```python
COMMON = {
    "no-default-libraries": False,
}

# ...

spec.update(COMMON)
spec.update(settings)
```

avr-specs/sync-from-upstream.py
```diff
@@ -0,0 +1,75 @@
...
+
+def main():
+    rustc_version = subprocess.run(
+        ['rustc', '--version'],
+        check=True,
+        stdout=subprocess.PIPE,
+    ).stdout.decode()
+
+    if 'nightly' not in rustc_version:
+        raise RuntimeError('You need nightly rustc!')
```
Rahix, line 52: I think `RuntimeError` is meant more for Python internal exceptions.  I'd just use `Exception` instead:

```suggestion
        raise Exception('You need nightly rustc!')
```

avr-specs/sync-from-upstream.py
```diff
@@ -0,0 +1,75 @@
...
+        stdout=subprocess.PIPE,
+    ).stdout
+
+    if upstream_spec_string is not None:
+        upstream_spec = json.loads(upstream_spec_string)
+
+    for mcu, settings in SPECS.items():
+        spec = copy.deepcopy(upstream_spec)
+        spec.update(settings)
+        spec['pre-link-args']['gcc'][0] = '-mmcu=%s' % mcu
```
Rahix, line 67: Please use f-String formatting instead of the legacy %-formatting:

```suggestion
        spec['pre-link-args']['gcc'][0] = f"-mmcu={mcu}"
```

avr-specs/sync-from-upstream.py
Rahix: BTW, the reason we need `"no-default-libraries": False,` is that we're currently relying on the avr-gcc "runtime" instead of doing it all in rust.  The long-term goal is of course to move away from that by providing our own custom runtime.

avr-specs/sync-from-upstream.py
```diff
@@ -0,0 +1,72 @@
...
+
+    upstream_spec_string = subprocess.run(
+        ['rustc', '--print', 'target-spec-json', '-Z',
+            'unstable-options', '--target', 'avr-unknown-gnu-atmega328'],
+        check=True,
+        stdout=subprocess.PIPE,
+    ).stdout
+
+    if upstream_spec_string is not None:
+        upstream_spec = json.loads(upstream_spec_string)
```
Rahix, line 58: Starting to look really nice, thanks for working on this!

One more note here:  What if `upstream_spec_string` is `None`?  Then `upstream_spec` won't be defined and code further down will raise exceptions.  Under what conditions would `upstream_spec_string` be `None` anyway?  The only thing I can think of would be a command failure but that's already guarded by `check=True`.  So I'd just remove the if here:

```suggestion
    upstream_spec = json.loads(upstream_spec_string)
```

Rahix (commented):
One more comment below and a general note: Please don't touch the avr-spec.json files yet, I'll take care of updating them after this PR (with just the script) is merged.

Rahix (approved):
> Please don't touch the avr-spec.json files yet, I'll take care of updating them after this PR (with just the script) is merged.

Please revert the changes to all these files (`avr-specs/*.json`), then I can merge :) I will take care of updating the specs this evening hopefully ...

explicite:
> > Please don't touch the avr-spec.json files yet, I'll take care of updating them after this PR (with just the script) is merged.
> 
> Please revert the changes to all these files (`avr-specs/*.json`), then I can merge :) I will take care of updating the specs this evening hopefully ...

Rahix:
Thank you!

PR #85, opened 2020-10-09
Bring target specs in line with upstream

### LINKED ISSUE #32: Cleanup target files

Rahix:
@explicite, I've reviewed all the changes which upstream did since I initially created these target files.  I have carefully done each change as a separate commit so it is possible to retrace what happened.  As far as I can tell, this does not produce any regressions in avr-hal, but let's see what CI has to say about this.

Finally fixes #32.


Rahix:
CI's happy :)

PR #87, opened 2020-10-10
leonardo runner - support for macos

explicite:
Basic checks and OS base port.

### DIFF
--- a/boards/arduino-leonardo/leonardo-runner.sh
+++ b/boards/arduino-leonardo/leonardo-runner.sh
@@ -9,9 +9,9 @@ esac
 
 if ! command -v numfmt &> /dev/null
 then
-    echo "required numfmt could not be found!"
-    echo "please install https://command-not-found.com/numfmt"
-    exit 1
+    echo "numfmt is needed for human-readable sizes." >&2
+    echo "please install https://command-not-found.com/numfmt" >&2
+    alias numfmt=true
 fi
 
 if ! command -v avrdude &> /dev/null
--- a/boards/arduino-leonardo/leonardo-runner.sh
+++ b/boards/arduino-leonardo/leonardo-runner.sh
@@ -1,14 +1,10 @@
 #!/usr/bin/env sh
 set -e
 
-LINUX_OS="LINUX"
-MAC_OS="MAC"
-UNKNOWN_OS="UNKNOWN"
-
 case "$(uname -s)" in
-    Linux*)     OS=$LINUX_OS;;
-    Darwin*)    OS=$MAC_OS;;
-    *)          OS=$UNKNOWN_OS
+    Linux*)     OS="Linux";;
+    Darwin*)    OS="Mac";;
+    *)          OS="Unknown";;
 esac
 
 if ! command -v numfmt &> /dev/null
--- a/boards/arduino-leonardo/leonardo-runner.sh
+++ b/boards/arduino-leonardo/leonardo-runner.sh
@@ -1,6 +1,39 @@
 #!/usr/bin/env sh
 set -e
 
+LINUX_OS="LINUX"
+MAC_OS="MAC"
+UNKNOWN_OS="UNKNOWN"
+
+case "$(uname -s)" in
+    Linux*)     OS=$LINUX_OS;;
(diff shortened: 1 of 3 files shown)


boards/arduino-leonardo/leonardo-runner.sh
```diff
@@ -1,6 +1,39 @@
...
 
+LINUX_OS="LINUX"
+MAC_OS="MAC"
+UNKNOWN_OS="UNKNOWN"
+
+case "$(uname -s)" in
+    Linux*)     OS=$LINUX_OS;;
+    Darwin*)    OS=$MAC_OS;;
+    *)          OS=$UNKNOWN_OS
+esac
```
Rahix, line 12: I'd remove the OS variables and just use strings directly:

```suggestion
case "$(uname -s)" in
    Linux*)     OS="Linux";;
    Darwin*)    OS="Mac";;
    *)          OS="Unknown";;
esac
```

boards/arduino-leonardo/leonardo-runner.sh
```diff
@@ -1,6 +1,39 @@
...
+
+if ! command -v numfmt &> /dev/null
+then
+    echo "required numfmt could not be found!"
+    echo "please install https://command-not-found.com/numfmt"
+    exit 1
```
Rahix, line 18: IMO `numfmt` missing shouldn't be a hard error.  Let's just add a warning and replace numfmt with a dummy function:

```suggestion
    echo "numfmt is needed for human-readable sizes." >&2
    echo "please install https://command-not-found.com/numfmt" >&2
    alias numfmt=true
```

boards/arduino-leonardo/leonardo-runner.sh
```diff
@@ -1,6 +1,39 @@
...
+fi
+
+if ! command -v avrdude &> /dev/null
+then
+    echo "required avrdude could not be found!"
+    echo "please install https://command-not-found.com/avrdude"
```
Rahix, line 24: Error messages should go to `stderr`:

```suggestion
    echo "required avrdude could not be found!" >&2
    echo "please install https://command-not-found.com/avrdude" >&2
```

boards/arduino-leonardo/leonardo-runner.sh
```diff
@@ -1,6 +1,39 @@
...
+fi
+
+if [ $OS = $LINUX_OS ]; then
+    ARDUINO_UPLOAD_PORT="/dev/ttyACM0"
+elif [ $OS = $MAC_OS ]; then
+    ARDUINO_UPLOAD_PORT="/dev/cu.usbmodem146201"
+else
+    echo "Not supported OS!"
+    exit 1    
+fi
```
Rahix, line 35: Let's call the variable `SERIAL_PORT`.  Also, should we really not allow "unsupported" OSes at all?  I'd rather try _something_ and if it fails, the user has to fix it.  e.g.

```sh
echo "unsupported OS, things might not work" >&2
SERIAL_PORT="/dev/ttyACM0"
```

boards/arduino-leonardo/leonardo-runner.sh
```diff
@@ -28,4 +57,4 @@ read -r
 printf "Attempting to flash ...\n"
 printf "\n"
 
-avrdude -qq -C/etc/avrdude.conf -patmega32u4 -cavr109 -P/dev/ttyACM0 -b57600 -D "-Uflash:w:$1:e"
```
Rahix, line 31: I'll have to try whether removing `-C/etc/avrdude.conf` works for me as well (I expect it will but better to double check ...).  Apart from that, this looks good now!

One more thing: Can you please do the same changes for the runner scripts for the other supported boards as well?  I know, this is a bit ugly to have code duplicated all over the place, but as I already mentioned the current runner scripts aren't here to stay anyway ...

explicite: Done 👍

Rahix (approved):
LGTM!  I am going to give this a test today in the evening, then I will merge :)

Rahix:
Still working fine for me as well.  Thanks!

Issue #80, opened 2020-10-04
More fine grained control of serial baudrate


Rahix:
Currently, the code only allows specifying a baudrate which the HAL-driver will try to match as closely as possible.  The "generic" algorithm used for this will not always produce the most optimal settings, however.  This means, for example, that 115200 baud on Arduino Uno is completely broken.

It would be nice to also allow setting exact baudrate parameters or/and providing alternate algorithms.  There are quite a few ways to implement this, here are a few ideas:

1. A new `Baudrate` type with methods for the different modes:

   ```rust
   // Clock is needed because the calculations needs to take core clock into account
   pub struct Baudrate<CLOCK> {
       pub ubrr: u16,
       pub u2x: bool,
       pub _clock: PhantomData<CLOCK>,
   }

   impl<CLOCK: crate::clock::Clock> Baudrate<CLOCK> {
       pub fn new(baud: u32) -> Baudrate<CLOCK> {
           // current implementation
       }

       pub fn with_exact(u2x: bool, ubrr: u16) -> Baudrate<CLOCK> {
           Baudrate { ubrr, u2x, _clock: PhantomData }
       }
   }
   ```

   The `Usart` constructor now takes `Baudrate<CLOCK>` instead of `u32`.

2. Using an `Into<Baudrate<CLOCK>>` bound in the `Usart` constructor.  We also implement `From<u32> for Baudrate<CLOCK>`.  This would allow backwards compatibility with current code (but I'm not sure if that is a good thing ...).

explicite:
I will try to start this. Probably I will have lot of questions...
1. This is the `generic algorithm` mentioned by you `let brr = CLOCK::FREQ / (16 * baud) - 1;`?
1. `Baundrate` should be defined as distinct file under generic src like Clock?

Rahix:
> This is the generic algorithm mentioned by you let brr = CLOCK::FREQ / (16 * baud) - 1;?

Yes, it's this code:

https://github.com/Rahix/avr-hal/blob/bad92776522bbbedee45d276991b45fb81f253d8/avr-hal-generic/src/serial.rs#L56-L58

In comparison, Arduino upstream uses [this code](https://github.com/arduino/ArduinoCore-avr/blob/3055c1efa3c6980c864f661e6c8cc5d5ac773af4/cores/arduino/HardwareSerial.cpp#L119-L136):

```cpp
  // Try u2x mode first
[...]

PR #88, opened 2020-10-11
[#80] Baudrate type implementation

explicite:
Base impl. for discussion

### DIFF
--- a/avr-hal-generic/src/serial.rs
+++ b/avr-hal-generic/src/serial.rs
@@ -1,5 +1,105 @@
 //! Serial Implementations
 
+use core::cmp::Ordering;
+
+// Clock is needed because the calculations needs to take core clock into account
+#[derive(Debug, Clone, Copy)]
+pub struct Baudrate<CLOCK> {
+    pub ubrr: u16,
+    pub u2x: bool,
+    pub _clock: ::core::marker::PhantomData<CLOCK>,
+}
+
+impl<CLOCK: crate::clock::Clock> PartialEq for Baudrate<CLOCK> {
+    fn eq(&self, other: &Self) -> bool {
+        self.compare_value() == other.compare_value()
+    }
+}
+
+impl<CLOCK: crate::clock::Clock> Eq for Baudrate<CLOCK> {}
+
+impl<CLOCK: crate::clock::Clock> PartialOrd for Baudrate<CLOCK> {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.compare_value().cmp(&other.compare_value()))
+    }
+}
+
+impl<CLOCK: crate::clock::Clock> Ord for Baudrate<CLOCK> {
+    fn cmp(&self, other: &Self) -> Ordering {
+        other.compare_value().cmp(&self.compare_value())
+    }
+}
+
+impl<CLOCK: crate::clock::Clock> From<u32> for Baudrate<CLOCK> {
+    fn from(baud: u32) -> Self {
+        Baudrate::new(baud)
+    }
+}
+
+impl<CLOCK: crate::clock::Clock> Baudrate<CLOCK> {
+    pub fn new(baud: u32) -> Baudrate<CLOCK> {
+        let mut ubrr = (CLOCK::FREQ / 4 / baud - 1) / 2;
+        let mut u2x = true;
+        debug_assert!(ubrr <= u16::MAX as u32);
+        if ubrr > 4095 {
+            u2x = false;
+            ubrr = (CLOCK::FREQ / 8 / baud - 1) / 2;
+        }
+
+        Baudrate {
+            ubrr: ubrr as u16,
+            u2x: u2x,
+            _clock: ::core::marker::PhantomData,
+        }
+    }
+
+    pub fn with_exact(u2x: bool, ubrr: u16) -> Baudrate<CLOCK> {
+        Baudrate {
+            ubrr, u2x, _clock: ::core::marker::PhantomData,
+        }
+    }
+
+    fn compare_value(&self) -> u32 {
+        if self.u2x {
+            return 8 * (self.ubrr as u32 + 1);
+        } else {
+            return 16 * (self.ubrr as u32 + 1);
+        };
+    }
+}
+
+pub trait BaudrateExt {
+    fn into_baudrate<CLOCK: crate::clock::Clock>(self) -> Baudrate<CLOCK>;
+}
+
+impl BaudrateExt for u32 {
+    fn into_baudrate<CLOCK: crate::clock::Clock>(self) -> Baudrate<CLOCK> {
+        Baudrate::new(self)
+    }
+}
+
+pub trait BaudrateArduinoExt {
+    fn into_baudrate<CLOCK: crate::clock::Clock>(self) -> Baudrate<CLOCK>;
+}
+
+impl BaudrateArduinoExt for u32 {
+    fn into_baudrate<CLOCK: crate::clock::Clock>(self) -> Baudrate<CLOCK> {
+        let br = Baudrate::new(self);
+
+        // hardcoded exception for 57600 for compatibility with the bootloader
+        // shipped with the Duemilanove and previous boards and the firmware
+        // on the 8U2 on the Uno and Mega 2560.
+        //
+        // https://github.com/arduino/ArduinoCore-avr/blob/3055c1efa3c6980c864f661e6c8cc5d5ac773af4/cores/arduino/HardwareSerial.cpp#L123-L132
+        if CLOCK::FREQ == 16_000_000 && br.ubrr == 34 && br.u2x {
+            // (CLOCK::FREQ / 8 / 57600 - 1) / 2 == 16
+            Baudrate::with_exact(false, 16)
+        } else {
+            br
+        }
+    }
+}
+
 /// Implement serial traits for a USART peripheral
 #[macro_export]
 macro_rules! impl_usart {
@@ -40,7 +140,7 @@ macro_rules! impl_usart {
                 p: $USART,
                 rx: $rxmod::$RX<$crate::port::mode::Input<RX_MODE>>,
                 tx: $txmod::$TX<$crate::port::mode::Output>,
-                baud: u32,
+                baud: Baudrate<CLOCK>,
             ) -> $Usart<CLOCK, RX_MODE> {
                 let mut usart = $Usart {
                     p,
@@ -52,12 +152,9 @@ macro_rules! impl_usart {
                 usart
             }
 
-            fn initialize(&mut self, baud: u32) {
-                // Value for baudrate register must be calculated based on clock frequency.
-                let brr = CLOCK::FREQ / (16 * baud) - 1;
-                self.p.[<ubrr $n>].write(|w| unsafe { w.bits(brr as u16) });
-
-                self.p.[<ucsr $n a>].reset();
+            fn initialize(&mut self, baud: Baudrate<CLOCK>) {
+                self.p.[<ubrr $n>].write(|w| unsafe { w.bits(baud.ubrr) });
+                self.p.[<ucsr $n a>].write(|w| w.[<u2x $n>]().bit(baud.u2x));
 
                 // Enable receiver and transmitter but leave interrupts disabled.
                 self.p.[<ucsr $n b>].write(|w| w
--- a/boards/arduino-diecimila/src/lib.rs
+++ b/boards/arduino-diecimila/src/lib.rs
@@ -1,20 +1,25 @@
 #![no_std]
 
-pub extern crate atmega168_hal as hal;
+// Expose hal & pac crates
+pub use atmega168_hal as hal;
+pub use crate::hal::pac;
+
 /// See [`avr_device::entry`](https://docs.rs/avr-device/latest/avr_device/attr.entry.html).
 #[cfg(feature = "rt")]
-pub use hal::entry;
+pub use crate::hal::entry;
 
 mod pins;
 
-pub use crate::atmega168::Peripherals;
+pub use crate::pac::Peripherals;
 pub use crate::pins::*;
-pub use atmega168_hal::adc;
-pub use atmega168_hal::atmega168;
-pub use atmega168_hal::prelude;
-pub use atmega168_hal::pwm;
-pub use atmega168_hal::spi;
+pub use crate::hal::adc;
+pub mod prelude {
+    pub use crate::hal::prelude::*;
+    pub use crate::hal::usart::BaudrateArduinoExt as _;
+}
+pub use crate::hal::pwm;
+pub use crate::hal::spi;
 
-pub type Delay = hal::delay::Delay<hal::clock::MHz16>;
-pub type Serial<IMODE> = hal::usart::Usart0<hal::clock::MHz16, IMODE>;
-pub type I2c<M> = hal::i2c::I2c<hal::clock::MHz16, M>;
+pub type Delay = crate::hal::delay::Delay<hal::clock::MHz16>;
+pub type Serial<IMODE> = crate::hal::usart::Usart0<hal::clock::MHz16, IMODE>;
+pub type I2c<M> = crate::hal::i2c::I2c<hal::clock::MHz16, M>;
--- a/boards/arduino-leonardo/examples/leonardo-adc.rs
+++ b/boards/arduino-leonardo/examples/leonardo-adc.rs
@@ -15,7 +15,7 @@ fn main() -> ! {
         dp.USART1,
         pins.d0,
         pins.d1.into_output(&mut pins.ddr),
-        57600,
+        57600.into_baudrate(),
     );
 
     ufmt::uwriteln!(&mut serial, "Reading analog inputs ...\r").void_unwrap();
--- a/boards/arduino-leonardo/examples/leonardo-i2cdetect.rs
+++ b/boards/arduino-leonardo/examples/leonardo-i2cdetect.rs
@@ -13,7 +13,7 @@ fn main() -> ! {
         dp.USART1,
         pins.d0,
         pins.d1.into_output(&mut pins.ddr),
-        57600,
+        57600.into_baudrate(),
     );
     let mut i2c = arduino_leonardo::I2c::new(
         dp.TWI,
--- a/boards/arduino-leonardo/src/lib.rs
+++ b/boards/arduino-leonardo/src/lib.rs
@@ -41,18 +41,24 @@
 
 #![no_std]
 
-pub extern crate atmega32u4_hal as hal;
+// Expose hal & pac crates
+pub use atmega32u4_hal as hal;
+pub use crate::hal::pac;
+
 /// See [`avr_device::entry`](https://docs.rs/avr-device/latest/avr_device/attr.entry.html).
 #[cfg(feature = "rt")]
-pub use hal::entry;
+pub use crate::hal::entry;
+
+pub use crate::pac::Peripherals;
 
 mod pins;
 pub use crate::pins::*;
 
-pub use atmega32u4_hal::atmega32u4;
-pub use crate::atmega32u4::Peripherals;
-pub use atmega32u4_hal::prelude;
-
+pub mod prelude {
+    pub use crate::hal::prelude::*;
+    pub use crate::hal::usart::BaudrateExt as _;
+}
+pub use crate::hal::usart;
 
 /// Busy-Delay
 ///
@@ -185,7 +191,7 @@ pub mod pwm {
 ///     dp.USART1,
 ///     pins.d0,
 ///     pins.d1.into_output(&mut pins.ddr),
-///     57600,
+///     57600.into_baudrate(),
 /// );
 ///
 /// ufmt::uwriteln!(&mut serial, "Hello from Arduino!\r").void_unwrap();
--- a/boards/arduino-mega2560/src/lib.rs
+++ b/boards/arduino-mega2560/src/lib.rs
@@ -1,22 +1,29 @@
 #![no_std]
 
-pub extern crate atmega2560_hal as hal;
+// Expose hal & pac crates
+pub use atmega2560_hal as hal;
+pub use crate::hal::pac;
+
 /// See [`avr_device::entry`](https://docs.rs/avr-device/latest/avr_device/attr.entry.html).
 #[cfg(feature = "rt")]
-pub use hal::entry;
+pub use crate::hal::entry;
+
+pub use crate::pac::Peripherals;
 
 mod pins;
-
-pub use atmega2560_hal::atmega2560;
-pub use crate::atmega2560::Peripherals;
-pub use atmega2560_hal::prelude;
-pub use atmega2560_hal::spi;
-pub use atmega2560_hal::adc;
 pub use crate::pins::*;
 
-pub type Delay = hal::delay::Delay<hal::clock::MHz16>;
-pub type Serial<IMODE> = atmega2560_hal::usart::Usart0<hal::clock::MHz16, IMODE>;
-pub type I2c<M> = hal::i2c::I2c<hal::clock::MHz16, M>;
+pub mod prelude {
+    pub use crate::hal::prelude::*;
+    pub use crate::hal::usart::BaudrateExt as _;
+}
+
+pub use crate::hal::spi;
+pub use crate::hal::adc;
+
+pub type Delay = crate::hal::delay::Delay<hal::clock::MHz16>;
+pub type Serial<IMODE> = crate::hal::usart::Usart0<hal::clock::MHz16, IMODE>;
+pub type I2c<M> = crate::hal::i2c::I2c<hal::clock::MHz16, M>;
 
 /// Support for PWM pins
 ///
@@ -62,5 +69,5 @@ pub type I2c<M> = hal::i2c::I2c<hal::clock::MHz16, M>;
 ///
 /// [ex-pwm]: https://github.com/sepotvin/avr-hal/blob/master/boards/arduino-mega2560/examples/mega2560-pwm.rs
 pub mod pwm {
-    pub use atmega2560_hal::pwm::*;
+    pub use crate::hal::pwm::*;
 }
--- a/boards/arduino-uno/src/lib.rs
+++ b/boards/arduino-uno/src/lib.rs
@@ -55,17 +55,23 @@
 
 #![no_std]
 
-pub extern crate atmega328p_hal as hal;
+// Expose hal & pac crates
+pub use atmega328p_hal as hal;
+pub use crate::hal::pac;
+
 /// See [`avr_device::entry`](https://docs.rs/avr-device/latest/avr_device/attr.entry.html).
 #[cfg(feature = "rt")]
-pub use hal::entry;
+pub use crate::hal::entry;
+
+pub use crate::pac::Peripherals;
 
 mod pins;
 pub use crate::pins::*;
 
-pub use atmega328p_hal::atmega328p;
-pub use crate::atmega328p::Peripherals;
-pub use atmega328p_hal::prelude;
+pub mod prelude {
+    pub use crate::hal::prelude::*;
+    pub use crate::hal::usart::BaudrateArduinoExt as _;
+}
 
 /// Busy-Delay
 ///
@@ -189,7 +195,7 @@ pub mod pwm {
 ///     dp.USART0,
 ///     pins.d0,
 ///     pins.d1.into_output(&mut pins.ddr),
-///     57600,
+///     57600.into_baudrate(),
 /// );
 ///
 /// ufmt::uwriteln!(&mut serial, "Hello from Arduino!\r").void_unwrap();
--- a/chips/atmega32u4-hal/src/lib.rs
+++ b/chips/atmega32u4-hal/src/lib.rs
@@ -1,14 +1,14 @@
 #![no_std]
 
-extern crate avr_hal_generic as avr_hal;
+/// Reexport of `atmega32u4` from `avr-device`
+pub use avr_device::atmega32u4 as pac;
 
-pub use avr_device::atmega32u4;
 /// See [`avr_device::entry`](https://docs.rs/avr-device/latest/avr_device/attr.entry.html).
 #[cfg(feature = "rt")]
 pub use avr_device::entry;
 
-pub use avr_hal::clock;
-pub use avr_hal::delay;
+pub use avr_hal_generic::clock;
+pub use avr_hal_generic::delay;
 
 pub mod adc;
 pub mod port;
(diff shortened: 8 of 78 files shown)


Rahix:
My suggestion would be to start working on this for one board only (preferably some hardware you have and can test things on).  Once discussion about the design has settled and all problems are ironed out, it can be transferred to all other crates as well.  This way, you don't have to make each change for all the crates every time.

avr-hal-generic/src/serial.rs
```diff
@@ -1,5 +1,32 @@
 //! Serial Implementations
 
+// Clock is needed because the calculations needs to take core clock into account
+#[derive(Debug, Clone, Copy)]
```
Rahix, line 4: All the other traits should also be implemented: `PartialEq`, `Eq`, `PartialOrd`, `Ord`, ...

Note that you'll have to implement this manually and calculate which one is bigger because `u2x` doubles the baudrate and this should be taken into account.

avr-hal-generic/src/serial.rs
```diff
@@ -52,10 +79,8 @@ macro_rules! impl_usart {
                 usart
             }
 
-            fn initialize(&mut self, baud: u32) {
-                // Value for baudrate register must be calculated based on clock frequency.
-                let brr = CLOCK::FREQ / (16 * baud) - 1;
-                self.p.[<ubrr $n>].write(|w| unsafe { w.bits(brr as u16) });
+            fn initialize(&mut self, baud: Baudrate<CLOCK>) {
+                self.p.[<ubrr $n>].write(|w| unsafe { w.bits(baud.ubrr) });
```
Rahix, line 83: The `u2x` bit needs to be written as well now.  It's in `UCSRnA`, so replace the `.reset()` with something similar to:

```rust
self.p.[<ucsr $n a>].write(|w| w.[<u2x $n>].bit(baud.u2x));
```

avr-hal-generic/src/serial.rs
explicite: Some of them implemented. I'm new to rust so please review and list other required.

avr-hal-generic/src/serial.rs
explicite: done

explicite:
@Rahix tested default on arduino leonardo. Do you have any specific scenarios for tests?
Right now Baudrate will contain some logic which we test with some units. Do we have any rules for this? Should I define some unit tests?

Rahix:
@explicite, my suggestion was to remove the changes from all board code except `arduino-leonardo` and its example for now.  This way the MR doesn't contain so many changes for now and once everything is looking good, you can readd the changes for the other boards.  Otherwise you'll have to update each file again and again which is a bit cumbersome ...

> Do you have any specific scenarios for tests?

We should test all the well known baudrates, ideally also with different core clock speeds.  Not sure how to easily try other clock speeds though (maybe by setting the clock prescaler?).

> Should I define some unit tests?

I guess this would be the right thing to do but we don't really have a testing infrastructure yet ... I think manual testing is enough for now.

avr-hal-generic/src/serial.rs
Rahix: The right place to look for is the API guidelines:  It has a point [`C-COMMON-TRAITS`](https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits) which lists all the important ones.

Looking at your implementation, I have a few notes:

1. I think you can remove the trait bound on `CLOCK` as you're not touching the type parameter anywhere.  It's only allowed to compare two `Baudrate`s if they have the same `CLOCK` anyway. So this is sufficient:

    ```rust
    impl<CLOCK> PartialEq for Baudrate<CLOCK> {
        fn eq(&self, other: &Self) -> bool {
            self.value() == other.value()
        }
    }
    ```

2. I think larger `ubrr` means smaller baudrate, but we need to recheck the datasheet on that ... If true, you got it the wrong way around (and also `u2x` should then mean _divide by 2_).

3. When `Ord` is implemented, you'd usually write `PartialOrd` like this:

    ```rust
    impl<CLOCK> PartialOrd for Baudrate<CLOCK> {
        fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
            Some(self.cmp(other))
        }
    }
    ```

explicite: 1. Without bound I'm not able to use `value()` fn
1.
32u4br
Looks like u2x decrease divider and increase baudrate.
1. Done
1. What abut asynchronous and synchronous difference? u2x works only on async. On sync always should be 0

Rahix: > Without bound I'm not able to use value() fn

You have to put `value()` into an impl block without the bound as well, then it'll work.  But honestly, it doesn't matter much ...

>  Looks like u2x decrease divider and increase baudrate.

Yes!  As per the formulas in the picture, _baudrate_ is inversely proportional to UBRR, so you need to swap operands in the `fn cmp()` impl (i.e. bigger UBRR then means smaller `Baudrate` value).  I think you'll also need to include the `+1` in your `fn value()` implementation to produce correct results.

> What abut asynchronous and synchronous difference? u2x works only on async. On sync always should be 0

We're always configuring the peripheral for _async_ operation:

https://github.com/Rahix/avr-hal/blob/7337cd76cd96f2d27701b137396d94a06d3a501d/avr-hal-generic/src/serial.rs#L71

avr-hal-generic/src/serial.rs
```diff
@@ -1,5 +1,75 @@
...
+        if ((CLOCK::FREQ == 16_000_000) && (baud == 57_600)) || (baud_setting > 4095) {
+            u2x = false;
+            baud_setting = (CLOCK::FREQ / 8 / baud - 1) / 2;
+        }
+
+        let ubrrh = baud_setting >> 8;
+        let ubrrl = baud_setting;
+
+        Baudrate {
+            ubrr: (ubrrh | ubrrl) as u16,
```
Rahix, line 49: You messed something up here.  I think you copied this from the arduino-core but over there something different is done:  Writing to the `*_ubrrh` and `*_ubrrl` variables _is_ writing to the ubrr register high and low bytes.  Your `baud_setting` already contains the complete 16-bit ubrr value so that's what you should be storing:

```suggestion
        debug_assert!(baud_setting <= u16::MAX);

        Baudrate {
            ubrr: baud_setting as u16,
```

avr-hal-generic/src/serial.rs
```diff
@@ -1,5 +1,75 @@
...
+
+    fn value(&self) -> u16 {
+        let divisor = (16 * (self.ubrr + 1)) as u32;
+        if self.u2x {
+            return (CLOCK::FREQ / (divisor / 2)) as u16;
+        } else {
+            return (CLOCK::FREQ / divisor) as u16;
+        };
+    }
+}
```
Rahix, line 71: I know what you're trying to do here but this isn't going to work well:  Doing these calculations is expensive and we can actually get around them by comparing inverse values and disregarding the `CLOCK` entirely:

```rust
    fn compare_value(&self) -> u32 {
        if self.u2x {
            8 * (self.ubrr as u32 + 1)
        } else {
            16 * (self.ubrr as u32 + 1)
        }
    }
```

and of course you then need to swap the operands in your `Ord` impl:

```rust
impl<CLOCK: crate::clock::Clock> Ord for Baudrate<CLOCK> {
    fn cmp(&self, other: &Self) -> Ordering {
        // Inverse comparison because compare_value() returns Fosc/BAUD (the inverse value).
        other.compare_value().cmp(&self.compare_value())
    }
}
```

boards/arduino-leonardo/src/lib.rs
```diff
@@ -172,20 +171,23 @@ pub mod pwm {
     pub use atmega32u4_hal::pwm::*;
 }
 
+pub type Baudrate = hal::usart::Baudrate<hal::clock::MHz16>;
```
Rahix, line 174: I'd prefer just re-exposing the usart module from the HAL:

```rust
pub use hal::usart;
```

and then importing it in the examples:

```rust
pub use arduino_leonardo::usart;

// ...
usart::Baudrate::new(...);
```

This is what we're doing for all other modules so I would argue it's best to use the same pattern for serial as well.

avr-hal-generic/src/serial.rs
```diff
@@ -40,7 +110,7 @@ macro_rules! impl_usart {
                 p: $USART,
                 rx: $rxmod::$RX<$crate::port::mode::Input<RX_MODE>>,
                 tx: $txmod::$TX<$crate::port::mode::Output>,
-                baud: u32,
+                baud: Baudrate<CLOCK>,
```
Rahix, line 113: I'm not 100% convinced that this is what we want to do.  I think there is value in allowing `u32` as well, which will then just go with the default implementation.  I.e. I'd allow any `Into<Baudrate<CLOCK>>` here and implement

```rust
impl<CLOCK: crate::clock::CLOCK> From<u32> for Baudrate<CLOCK> {
    ...
}
```

avr-hal-generic/src/serial.rs
```diff
@@ -1,5 +1,75 @@
...
+    pub fn new(baud: u32) -> Baudrate<CLOCK> {
+        // Try u2x mode first
+        let mut baud_setting = (CLOCK::FREQ / 4 / baud - 1) / 2;
+        let mut u2x = true;
+
+        // Switch back from u2x if required
+        if ((CLOCK::FREQ == 16_000_000) && (baud == 57_600)) || (baud_setting > 4095) {
+            u2x = false;
+            baud_setting = (CLOCK::FREQ / 8 / baud - 1) / 2;
+        }
```
Rahix, line 43: This is what Arduino is doing but especially that workaround for their bootloader seems suspicious to me.  We have to remember that not everyone using the HAL is using it for Arduino boards so we should be careful about shipping Arduino workarounds everywhere by default.  I'll look deeper into this and share my results.

avr-hal-generic/src/serial.rs
Rahix: An alternative would be to implement an extension trait on `u32` for converting a value into a baudrate.  E.g.

```rust
trait BaudExt {
    fn into_baud<CLOCK: crate::clock::Clock>(self) -> Baudrate<CLOCK>;
}

impl BaudExt for u32 {
    fn into_baud<CLOCK: crate::clock::Clock>(self) -> Baudrate<CLOCK> {
        Baudrate::new(self)
    }
}
```

and in application code

```rust
use arduino_leonardo::prelude::*;

Serial::new(..., 57600.into_baud());
```

explicite: With `Into` I'm getting `the trait `Into` cannot be made into an object`

boards/arduino-leonardo/src/lib.rs
explicite: done, should I fix for other imports?

avr-hal-generic/src/serial.rs
explicite: Thanks, idea wast to do this 1-2-1 with datasheet and then simplify.

avr-hal-generic/src/serial.rs
Rahix: You have to either make it a generic parameter:

```rust
pub fn new<B: Into<Baudrate<CLOCK>>>(..., baudrate: B)
```

or use the `impl Trait` syntax:

```rust
pub fn new(..., baudrate: impl Into<Baudrate<CLOCK>>)
```

---

Though, to be honest I prefer my second alternative suggestion even more.  It would allow us to provide board-specific `into_baud()` implementations so boards that need non-standard calculations can be supported easily.

boards/arduino-leonardo/src/lib.rs
Rahix: Not sure what other imports you're talking about?

avr-hal-generic/src/serial.rs
explicite: Will move to second option

boards/arduino-leonardo/src/lib.rs
explicite: Looks like I made same as `arduino_leonardo::Pins` but is defined under leaonardo not in generic. I review other and they are correct.

Rahix: Ah, `Pins` is the exception :D  The reason is that `Pins` is so fundamental that it's just annoying when hidden behind a module.  The type-alias for e.g. `Serial` is similar though I am less sure if we want to keep that in the long run.

avr-hal-generic/src/serial.rs
explicite: Not sure where put `BaudrateExt` trait. It should be in board lib?

Rahix: `BaudrateExt` goes into the prelude as an anonymous import:

```rust
pub use ...::BaudrateExt as _;
```

explicite: Can you please review if I done this in correct way? Thanks.

Rahix: The import looks good but you put it in the wrong place.  Sorry, I should have explained better:  It's meant to go into the `prelude` module which is a module imported in all the examples as 

```rust
use arduino_leonardo::prelude::*;
```

This pulls all the traits into scope which means all their methods are available to be called (because a trait must be in scope for its methods to be callable).  The anonymous import

```rust
use ...::BaudrateExt as _;
```

means that the 'name' `BaudrateExt` is not actually imported, which would 'pollute' the namespace where `prelude` is imported.  You can think of this as importing `BaudrateExt` under an alternate name like `BaudrateExt32748937248972394792389479` which is unlikely to clash with any names defined in the file importing `prelude`.

Anyway, in practical terms, right now `arduino-leonardo` just reexports the prelude from the HAL:

https://github.com/Rahix/avr-hal/blob/7337cd76cd96f2d27701b137396d94a06d3a501d/boards/arduino-leonardo/src/lib.rs#L54

Instead of this, you want to create a new `prelude` module here:

```rust
pub mod prelude {
    // Expose everything from the HAL prelude
    pub use atmega32u4_hal::prelude::*;

    pub use ...::BaudrateExt as _;
}
```

avr-hal-generic/src/serial.rs
Rahix: In general, enabling U2X consistently yields clock rates with equal or less error than with disabled U2X so I'd say it should be the default where possible.  I.e. it makes sense to ship this algorithm in the `Baudrate::new()` method.

The workaround (`(CLOCK::FREQ == 16_000_000) && (baud == 57_600))`) is really not something we want to include by default, however.  It actually makes the clock error worse, going from `-0.8%` (with u2x) to `+2.1%` (without u2x).  This is (apparently) necessary because the USB-to-Serial bridge on some arduino boards has a firmware with a similar clock error and the `-0.8%` in the other direction would make the total error `3%` which is too much, I guess.

My suggestion would be to add a custom `BaudrateArduinoExt` trait where the fix is applied.  On the affected boards, `BaudrateArduinoExt` is pulled into the prelude instead of `BaudrateExt`.

```rust
pub trait BaudrateArduinoExt {
    fn into_baudrate<CLOCK: crate::clock::Clock>(self) -> Baudrate<CLOCK>;
}

impl BaudrateArduinoExt for u32 {
    fn into_baudrate<CLOCK: crate::clock::Clock>(self) -> Baudrate<CLOCK> {
        let br = Baudrate::new(self);

        // hardcoded exception for 57600 for compatibility with the bootloader
        // shipped with the Duemilanove and previous boards and the firmware
        // on the 8U2 on the Uno and Mega 2560.
        //
        // https://github.com/arduino/ArduinoCore-avr/blob/3055c1efa3c6980c864f661e6c8cc5d5ac773af4/cores/arduino/HardwareSerial.cpp#L123-L132
        if CLOCK::FREQ == 16_000_000 && br.ubrr == 34 && br.u2x {
            // (CLOCK::FREQ / 8 / 57600 - 1) / 2 == 16
            Baudrate::with_exact(false, 16)
        } else {
            br
        }
    }
}
```

The only problem I see with this is that it's a bit "magical" and makes it quite non-obvious which version is used.  Though the alternative of requiring board users to explicitly do something different depending on the hardware they are using, is against the idea of having a HAL in the first place ...

explicite: I done this. Now I'm testing `leonardo-serial` example. After upload, arduino is not recognized until I will reset. After that I need to cleanup board to works fine without reset.

I'm not sure if it's regression, serial example (need to spend some more time on example itself) or problem with `.elf`

Rahix: Right, you're probably trying to access the serial console directly via the USB cable which you use to flash code to the board?  That's not going to work on Leonardo as it is different from the other Arduino boards:  On other Arduinos there is a separate USB-to-Serial converter between the USB port and the actual MCU.  On Leonardo, the USB port is directly connected to the MCU (ATmega32U4) and for Arduino C code, the MCU directly presents itself to the host as a USB-to-Serial converter.  In Rust we don't do that (at this time at least).  Instead, what you're working with right now is the real USART peripheral connected to the pins D0 (RX) and D1 (TX).  So to access it you'll need a separate USB-to-Serial converter to connect to those pins.

explicite: Nice, I have some usb to ttl form my old ESP project. After quick test I was able to send some chars with `let baudrate = 9600_u32.into_baudrate();` and receive correct response: Will test below combination - with Ext. impl and with exact ubrr
baund| UBRR|tested|
-- | -- |--|
300 | 767 | no|
600 | 383 | no|
1200 | 191 |no|
2400 | 95 | no|
4800 | 47 | no|
9600 | 23 | x |
14400 | 15 | no|
19200 | 11 | no|
28800 | 7 | no|
38400 | 5 | no|
57600 | 3 | no|
76800 | 2 | no|
115200 | 1|no|

avr-hal-generic/src/serial.rs
```diff
@@ -1,5 +1,102 @@
...
+impl<CLOCK: crate::clock::Clock> Baudrate<CLOCK> {
+    pub fn new(baud: u32) -> Baudrate<CLOCK> {
+        let brr = (CLOCK::FREQ / 4 / baud - 1) / 2;
+        debug_assert!(brr <= u16::MAX as u32);
+
+        Baudrate {
+            ubrr: brr as u16,
+            u2x: true,
+            _clock: ::core::marker::PhantomData,
+        }
```
Rahix, line 48: Hmm, now you've removed too much:  If the baudrate is very low, the `ubrr` you're calculating here could go too high (maximum allowed value is 4095).  For this situation, you should fall back to non-u2x mode:

```suggestion
        let mut ubrr = (CLOCK::FREQ / 4 / baud - 1) / 2;
        let mut u2x = true;
        debug_assert!(brr <= u16::MAX as u32);
        
        if ubrr > 4095 {
            u2x = false;
            ubrr = (CLOCK::FREQ / 8 / baud - 1) / 2;
        }

        Baudrate {
            ubrr: ubrr as u16,
            u2x: u2x,
            _clock: ::core::marker::PhantomData,
        }
```

avr-hal-generic/src/serial.rs
explicite: For now I have three failing baunds `14400`, `28800` and `76800`

Rahix: You might want to check whether the chip on your USB-to-Serial converter properly supports those rates before diving into the AVR side.  Those three don't look like common values to me, so maybe there is some problem ...

Rahix: That aside, we should also run a test on an Arduino Uno with the special implementation to check whether that's working as intended.  If you don't have a device on hand, I can take care of that :)

explicite: I will be able to test with uno in the end of this week. So if with your help we can be faster.

PL-2303HX should support some of the bauds. http://www.prolific.com.tw/UserFiles/files/ds_pl2303HXD_v1_4_4.pdf
Not sure about linux drivers. Right now I'm on 5.4.0-52-generic and I can see it. Not sure if supports all bauds.

Rahix: Linux will just pass the baudrate to the CDC-ACM endpoint exposed by your PL-2303HX.  What the device makes of it is the important part, some just select the closest baudrate, others properly configure their clock for the exact selected speed.

From the datasheet you linked, at least 14400 and 28800 should work though ... You don't by any chance have an oscilloscope at hand to check the signal?  I'm a bit short on time this and the next week.  Though after that I could run some intrusive tests to see.

That said, I wouldn't block this PR's progress on these few baudrates though.  As long as all common rates are well supported, we are good I'd say.

avr-hal-generic/src/serial.rs
```diff
@@ -1,5 +1,107 @@
...
+    }
+
+    pub fn with_exact(u2x: bool, ubrr: u16) -> Baudrate<CLOCK> {
+        Baudrate {
+            ubrr: ubrr,
+            u2x,
+            _clock: ::core::marker::PhantomData,
+        }
```
Rahix, line 61: Can be written a bit more concise:

```suggestion
        Baudrate {
            ubrr, u2x, _clock: ::core::marker::PhantomData,
        }
```

boards/arduino-leonardo/examples/leonardo-adc.rs
```diff
@@ -10,12 +10,13 @@ fn main() -> ! {
...
     let mut pins = arduino_leonardo::Pins::new(dp.PORTB, dp.PORTC, dp.PORTD, dp.PORTE, dp.PORTF);
+    let baudrate = 57600.into_baudrate();
 
     let mut serial = arduino_leonardo::Serial::new(
         dp.USART1,
         pins.d0,
         pins.d1.into_output(&mut pins.ddr),
-        57600,
+        baudrate,
     );
```
Rahix, line 20: The variable here is superfluous, just inline the expression into the function call:

```suggestion
    let mut serial = arduino_leonardo::Serial::new(
        dp.USART1,
        pins.d0,
        pins.d1.into_output(&mut pins.ddr),
        57600.into_baudrate(),
    );
```

The same goes for all the other examples.

boards/arduino-leonardo/examples/leonardo-i2cdetect.rs
```diff
@@ -2,18 +2,20 @@
 #![no_main]
 
 use arduino_leonardo::prelude::*;
+use arduino_leonardo::usart;
```
Rahix, line 5: Unnecessary import, as of the latest version.

boards/arduino-leonardo/examples/leonardo-i2cdetect.rs
```diff
@@ -2,18 +2,20 @@
...
 
     let mut pins = arduino_leonardo::Pins::new(dp.PORTB, dp.PORTC, dp.PORTD, dp.PORTE, dp.PORTF);
+    let baudrate = 57600.into_baudrate;
     let mut serial = arduino_leonardo::Serial::new(
         dp.USART1,
         pins.d0,
         pins.d1.into_output(&mut pins.ddr),
-        57600,
+        baudrate,
     );
```
Rahix, line 19: As above, please just inline the `.into_baudrate()` call:

```suggestion
    let mut serial = arduino_leonardo::Serial::new(
        dp.USART1,
        pins.d0,
        pins.d1.into_output(&mut pins.ddr),
        57600.into_baudrate(),
    );
```

boards/arduino-leonardo/src/lib.rs
```diff
@@ -45,15 +45,20 @@ pub extern crate atmega32u4_hal as hal;
...
 pub use crate::atmega32u4::Peripherals;
-pub use atmega32u4_hal::prelude;
+pub use atmega32u4_hal::atmega32u4;
 
+pub mod prelude {
+    // Expose everything from the HAL prelude
+    pub use atmega32u4_hal::prelude::*;
 
+    pub use hal::usart::BaudrateArduinoExt as _;
+}
```
Rahix, line 61: Please rebase this branch/merge upstream changes into it, as the imports here have slightly changed in `master`.  The prelude module looks good, though!

boards/arduino-leonardo/src/lib.rs
```diff
@@ -172,20 +177,23 @@ pub mod pwm {
     pub use atmega32u4_hal::pwm::*;
 }
 
+pub type Baudrate = hal::usart::Baudrate<hal::clock::MHz16>;
+
```
Rahix, line 181: Please drop this type alias, it is not needed due to the `*Ext` traits you added.

```suggestion
```

boards/arduino-leonardo/src/lib.rs
```diff
@@ -172,20 +177,23 @@ pub mod pwm {
...
 ///
+/// let baudrate = 57600.into_baudrate;
 /// let mut pins = arduino_leonardo::Pins::new(dp.PORTB, dp.PORTC, dp.PORTD, dp.PORTE);
 ///
 /// let mut serial = arduino_leonardo::Serial::new(
 ///     dp.USART1,
 ///     pins.d0,
 ///     pins.d1.into_output(&mut pins.ddr),
-///     57600,
+///     baudrate,
```
Rahix, line 196: Don't create an extra variable here:

```suggestion
/// let mut pins = arduino_leonardo::Pins::new(dp.PORTB, dp.PORTC, dp.PORTD, dp.PORTE);
///
/// let mut serial = arduino_leonardo::Serial::new(
///     dp.USART1,
///     pins.d0,
///     pins.d1.into_output(&mut pins.ddr),
///     57600.into_baudrate(),
```

chips/atmega32u4-hal/src/lib.rs
```diff
@@ -94,6 +94,7 @@ pub mod spi {
 /// Serial interface using USART
 pub mod usart {
     use crate::port::portd;
+    pub use avr_hal::serial::BaudrateExt as _;
```
Rahix, line 97: Not necessary here, the trait is included in the `prelude`.

Rahix (commented):
From the technical side, I think we're all good now!  I've left a few comments related to cleanup, please take care of them.  Afterwards, let's do some more tests to make sure it's working correctly now and then fixup all the other HAL & Board crates where necessary.  Thanks for working on this :)

boards/arduino-leonardo/src/lib.rs
explicite: done, not sure if in correct way

Rahix: Looks good to me!

Rahix (changes requested):
I have some more minor style nit-picks, but apart from that it's looking really good!

Once you've got those applied, please go ahead and apply the same changes as in `boards/arduino-leonardo/src/lib.rs` to all the other board crates.  We'll have to check which ones need the fixed `..Ext` trait and which don't.  AFAIK at least `arduino-uno` and `arduino-diecimila` will need it.  Then, of course all the examples need an update as well.

Rahix:
Oh, and feel free to remove the "Draft" state from this PR :)

explicite:
@Rahix  should I implement same changes for other chips and boards?

Rahix (approved):
Yeah, I think it looks good now!  Please do the changes from `boards/arduino-leonardo/src/lib.rs` for all the other boards as well now.  That means, apply the following changeset:

```diff
@@ -54,7 +54,11 @@ pub use crate::pac::Peripherals;
 mod pins;
 pub use crate::pins::*;
 
-pub use crate::hal::prelude;
+pub mod prelude {
+    pub use crate::hal::prelude::*;
+    pub use crate::hal::usart::BaudrateExt as _;
+}
+pub use crate::hal::usart;
 
 /// Busy-Delay
 ///
@@ -187,7 +191,7 @@ pub mod pwm {
 ///     dp.USART1,
 ///     pins.d0,
 ///     pins.d1.into_output(&mut pins.ddr),
-///     57600,
+///     57600.into_baudrate(),
 /// );
 ///
 /// ufmt::uwriteln!(&mut serial, "Hello from Arduino!\r").void_unwrap();
```

Of course you'll also have to fix all the relevant examples alongside it:

```diff
     let mut serial = <board>::Serial::new(
         dp.USART1,
         pins.d0,
         pins.d1.into_output(&mut pins.ddr),
-        57600,
+        57600.into_baudrate(),
     );
```

explicite:
@Rahix  I think all fixed as required. Should I made any changes/testing?

boards/arduino-uno/src/lib.rs
```diff
@@ -68,7 +68,10 @@ pub use crate::pac::Peripherals;
 mod pins;
 pub use crate::pins::*;
 
-pub use crate::hal::prelude;
+pub mod prelude {
+    pub use crate::hal::prelude::*;
+    pub use crate::hal::usart::BaudrateExt as _;
```
Rahix, line 73: I think (but we need to test!) for the uno we need the special implementation:

```suggestion
    pub use crate::hal::usart::BaudrateArduinoExt as _;
```

boards/arduino-mega2560/src/lib.rs
```diff
@@ -13,7 +13,10 @@ pub use crate::pac::Peripherals;
 mod pins;
 pub use crate::pins::*;
 
-pub use crate::hal::prelude;
+pub mod prelude {
+    pub use crate::hal::prelude::*;
+    pub use crate::hal::usart::BaudrateExt as _;
```
Rahix, line 18: Also need to check whether this is affected by the special implementation.  @couchand, maybe you can give a try of this PR at 57600 baud?

boards/arduino-diecimila/src/lib.rs
```diff
@@ -13,7 +13,10 @@ mod pins;
 pub use crate::pac::Peripherals;
 pub use crate::pins::*;
 pub use crate::hal::adc;
-pub use crate::hal::prelude;
+pub mod prelude {
+    pub use crate::hal::prelude::*;
+    pub use crate::hal::usart::BaudrateExt as _;
```
Rahix, line 18: Pretty sure diecimila needs the fix:

```suggestion
    pub use crate::hal::usart::BaudrateArduinoExt as _;
```

Rahix (changes requested):
Looking pretty good!  Only thing we need to change is making use of the special implementation on affected boards.  We should test where it applies, I've marked the potential candidates with comments:

boards/arduino-mega2560/src/lib.rs
explicite: In worst case I think will be able to test this in next week.

explicite: For mega2560 same results. I will check now Uno rev3 for same baunds

baund| UBRR|tested|
-- | -- |--|
300 | 767 |ok |
600 | 383 |ok|
1200 | 191 |ok|
2400 | 95 |ok |
4800 | 47 |ok |
9600 | 23 |ok  |
14400 | 15 |failed |
19200 | 11 |ok |
28800 | 7 |failed |
38400 | 5 |ok |
57600 | 3 | ok|
76800 | 2 |failed |
115200 | 1|ok|

avr-hal-generic/src/serial.rs
explicite: I do not have now any oscilloscope now. With time I will buy some. Now I don't want to invest to much now.

boards/arduino-mega2560/src/lib.rs
Rahix: This is with the `BaudrateExt` trait, right?  So it seems 57600 and 115200 are working fine with the default implementation which means we don't need to switch for this board!

explicite: Yes. Looks like works fine. I'm able to test today uno rev3 too. Then I will need to reupload bootloader for mega2560...

Rahix: If you don't have any objections, I'm content with the results you provided and would go ahead with merging your PR now.  Thank you so much for you work here!  You were a lot more thorough than most people would be which I appreciate a lot. :)

avr-hal-generic/src/serial.rs
Rahix: Let's open an issue for this an resolve it at a later time: #99

Rahix:
Looks like you'll need to fix merge conflicts/rebase your branch once more, then I can merge it :)

Rahix (approved):

PR #89, opened 2020-10-18
Refactor hal crates

Rahix:
I'm working on cleaning up the import and module structure to make the hal crates look more similar.  This is hopefully a first step towards making them more maintainable.  In the future we should explore combining some of them to reduce the hellish amount of code duplication.

**Note**: This might break some downstream code as public items are being moved around/renamed.

### DIFF
--- a/avr-hal-generic/src/adc.rs
+++ b/avr-hal-generic/src/adc.rs
@@ -60,7 +60,7 @@ macro_rules! impl_adc {
         use $crate::hal::adc::{Channel, OneShot};
         use $crate::nb;
         use $crate::port::mode::Analog;
-        pub use avr_hal::adc::*;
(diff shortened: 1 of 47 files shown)


Rahix:
Looking for some feedback before I merge this.  I'd mostly be interested how much this breaks current downstream code and whether the changes decrease downstream code readability.  You can try it by simply pointing all `avr-hal` depedencies towards `rev = "9df51b16a7893fbdc4085d6d0647a9a84010bbd1 "`.  For example:

```toml
[dependencies.arduino-leonardo]
git = "https://github.com/Rahix/avr-hal"
rev = "9df51b16a7893fbdc4085d6d0647a9a84010bbd1"
```

Issue #75, opened 2020-09-25
Need examples showing how to use timer interrupts : ISR : OCR: interrupt service routine: output compare register  


ToddG:
## BACKGROUND

I'm attempting to follow various examples for writing timer driven interrupt service routines in rust targeting the AVR...similar to these examples in `arduino c`:

https://www.instructables.com/id/Arduino-Timer-Interrupts/
https://circuitdigest.com/microcontroller-projects/arduino-timer-tutorial

### EXAMPLE `c` CODE

In these examples, there is a stanza for setting the various registers to set this up. Here's an extract:

```c
//set timer1 interrupt at 1Hz
  TCCR1A = 0;// set entire TCCR1A register to 0
  TCCR1B = 0;// same for TCCR1B
  TCNT1  = 0;//initialize counter value to 0
  // set compare match register for 1hz increments
  OCR1A = 15624;// = (16*10^6) / (1*1024) - 1 (must be <65536)
  // turn on CTC mode
  TCCR1B |= (1 << WGM12);
  // Set CS10 and CS12 bits for 1024 prescaler
  TCCR1B |= (1 << CS12) | (1 << CS10);  
  // enable timer compare interrupt
  TIMSK1 |= (1 << OCIE1A);
```

Then these c examples show how to wire up an ISR:

```c
ISR(TIMER1_COMPA_vect){...}
```

### AVR-HAL `rust` CODE

In the `avr-hal` repo's various examples, I only see reference to the timer registers of `tccr1b` and `tcnt1`.  Is this an oversight or is this because of some limitation on the platform? It would be really useful to see an example that included the other registers such as compare `ocr1a` and especially an example showing the use of an interrupt service routine like the common `c` ISR above.

Links:
Here's the example code I am following, though I've looked at all the `avr-hal` examples that include a timer:

1. [Create the timer](https://github.com/Rahix/avr-hal/blob/2a59d441a5b9506b3cbfac21e532639733a6314a/boards/arduino-uno/examples/uno-hc-sr04.rs#L22)
2. [Configure prescaling (TCCR1B)](https://github.com/Rahix/avr-hal/blob/2a59d441a5b9506b3cbfac21e532639733a6314a/boards/arduino-uno/examples/uno-hc-sr04.rs#L23)
[...]

PR #93, opened 2020-10-25
[#90] I2C slave driver

### LINKED ISSUE #90: I2C Slave Driver
Currently, we only support [TWI/I2c in master mode](https://github.com/Rahix/avr-hal/blob/master/avr-hal-generic/src/i2c.rs).  In addition to that, let's add a driver to operate the TWI peripheral in slave mode.

Cc @ToddG

ToddG:
Draft PR.

Closes #90.

### DIFF
--- a/avr-hal-generic/src/i2c.rs
+++ b/avr-hal-generic/src/i2c.rs
@@ -210,14 +210,14 @@ macro_rules! impl_twi_i2c {
     ) => {$crate::paste::paste! {
         /// I2C Master
         $(#[$i2c_attr])*
-        pub struct $I2c<CLOCK: $crate::clock::Clock, M> {
+        pub struct [<$I2c Master>]<CLOCK: $crate::clock::Clock, M> {
             p: $I2C,
             _clock: ::core::marker::PhantomData<CLOCK>,
             sda: $sdamod::$SDA<M>,
             scl: $sclmod::$SCL<M>,
         }
 
-        impl<CLOCK> $I2c<CLOCK, $crate::i2c::I2cPullUp>
+        impl<CLOCK> [<$I2c Master>]<CLOCK, $crate::i2c::I2cPullUp>
         where
             CLOCK: $crate::clock::Clock,
         {
@@ -231,7 +231,7 @@ macro_rules! impl_twi_i2c {
                 sda: $sdamod::$SDA<$crate::port::mode::Input<$crate::port::mode::PullUp>>,
                 scl: $sclmod::$SCL<$crate::port::mode::Input<$crate::port::mode::PullUp>>,
                 speed: u32,
-            ) -> $I2c<CLOCK, $crate::i2c::I2cPullUp> {
+            ) -> [<$I2c Master>]<CLOCK, $crate::i2c::I2cPullUp> {
                 // Calculate TWBR
                 let twbr = ((CLOCK::FREQ / speed) - 16) / 2;
                 p.$twbr.write(|w| unsafe { w.bits(twbr as u8) });
@@ -247,7 +247,7 @@ macro_rules! impl_twi_i2c {
             }
         }
 
-        impl<CLOCK> $I2c<CLOCK, $crate::i2c::I2cFloating>
+        impl<CLOCK> [<$I2c Master>]<CLOCK, $crate::i2c::I2cFloating>
         where
             CLOCK: $crate::clock::Clock,
         {
@@ -261,7 +261,7 @@ macro_rules! impl_twi_i2c {
                 sda: $sdamod::$SDA<$crate::port::mode::Input<$crate::port::mode::Floating>>,
                 scl: $sclmod::$SCL<$crate::port::mode::Input<$crate::port::mode::Floating>>,
                 speed: u32,
-            ) -> $I2c<CLOCK, $crate::i2c::I2cFloating> {
+            ) -> [<$I2c Master>]<CLOCK, $crate::i2c::I2cFloating> {
                 // Calculate TWBR
                 let twbr = ((CLOCK::FREQ / speed) - 16) / 2;
                 p.$twbr.write(|w| unsafe { w.bits(twbr as u8) });
@@ -277,7 +277,7 @@ macro_rules! impl_twi_i2c {
             }
         }
 
-        impl<CLOCK, M> $I2c<CLOCK, M>
+        impl<CLOCK, M> [<$I2c Master>]<CLOCK, M>
         where
             CLOCK: $crate::clock::Clock,
         {
@@ -434,7 +434,7 @@ macro_rules! impl_twi_i2c {
             }
         }
 
-        impl<CLOCK, M> $I2c<CLOCK, M>
+        impl<CLOCK, M> [<$I2c Master>]<CLOCK, M>
         where
             CLOCK: $crate::clock::Clock,
             $crate::delay::Delay<CLOCK>: $crate::hal::blocking::delay::DelayMs<u16>,
@@ -458,7 +458,7 @@ macro_rules! impl_twi_i2c {
         }
 
 
-        impl<CLOCK, M> $crate::hal::blocking::i2c::Write for $I2c<CLOCK, M>
+        impl<CLOCK, M> $crate::hal::blocking::i2c::Write for [<$I2c Master>]<CLOCK, M>
         where
             CLOCK: $crate::clock::Clock,
         {
@@ -472,7 +472,7 @@ macro_rules! impl_twi_i2c {
             }
         }
 
-        impl<CLOCK, M> $crate::hal::blocking::i2c::Read for $I2c<CLOCK, M>
+        impl<CLOCK, M> $crate::hal::blocking::i2c::Read for [<$I2c Master>]<CLOCK, M>
         where
             CLOCK: $crate::clock::Clock,
         {
@@ -486,7 +486,7 @@ macro_rules! impl_twi_i2c {
             }
         }
 
-        impl<CLOCK, M> $crate::hal::blocking::i2c::WriteRead for $I2c<CLOCK, M>
+        impl<CLOCK, M> $crate::hal::blocking::i2c::WriteRead for [<$I2c Master>]<CLOCK, M>
         where
             CLOCK: $crate::clock::Clock,
         {
--- a/avr-hal-generic/src/i2c.rs
+++ b/avr-hal-generic/src/i2c.rs
@@ -505,6 +505,12 @@ macro_rules! impl_twi_i2c {
         /// do not need to pass the speed/bitrate as that is determined by the Master. However, the
         /// Slave constructors do need to pass the slave address and the flag to enable/disable the
         /// General Call Address.
+        ///
+        /// Example (pseudocode):
+        ///
+        ///     let s = [<$I2c Slave>]::new(...)
+        ///     s.start()
+        ///     TODO: need feedback from Rahix on how they want this API to look
         $(#[$i2c_attr])*
         pub struct [<$I2c Slave>]<M> {
             /// Peripherals
@@ -568,7 +574,12 @@ macro_rules! impl_twi_i2c {
         }
 
         impl <M>[<$I2c Slave>]<M>{
-            pub fn ack(self) -> [<$I2c Slave>]<M>{
+            /// Start the slave listening for messages on the I2C bus
+            pub fun start(self) -> Result<[<$I2c SlaveStateUninitialized>]<M>, $crate::i2c::Error>{
+                [<$I2c SlaveState>]::new(self)
+            }
+
+            fn ack(self) -> [<$I2c Slave>]<M>{
                 self.p.twcr.write(|w| w
                     .twea().set_bit()
                     .twsto().clear_bit()
@@ -578,7 +589,7 @@ macro_rules! impl_twi_i2c {
                 self
             }
 
-            pub fn nack(self) -> [<$I2c Slave>]<M> {
+            fn nack(self) -> [<$I2c Slave>]<M> {
                 self.p.twcr.write(|w| w
                     .twea().set_bit()
                     .twsto().clear_bit()
--- a/avr-hal-generic/src/i2c.rs
+++ b/avr-hal-generic/src/i2c.rs
@@ -232,10 +232,27 @@ macro_rules! impl_twi_i2c {
                 scl: $sclmod::$SCL<$crate::port::mode::Input<$crate::port::mode::PullUp>>,
                 speed: u32,
             ) -> [<$I2c Master>]<CLOCK, $crate::i2c::I2cPullUp> {
-                // Calculate TWBR
+                // Calculate the TWI Bit Rate Register (TWBR)
+                //
+                // (datasheet) 22.5.2 Bit Rate Generator Unit
+                //
+                //      This unit controls the period of SCL when operating in a Master mode. The SCL period is controlled by settings
+                //      in the TWI Bit Rate Register (TWBR) and the Prescaler bits in the TWI Status Register (TWSR). Slave operation
+                //      does not depend on Bit Rate or Prescaler settings, but the CPU clock frequency in the Slave must be at least 16
+                //      times higher than the SCL frequency. Note that slaves may prolong the SCL low period, thereby reducing the
+                //      average TWI bus clock period.
                 let twbr = ((CLOCK::FREQ / speed) - 16) / 2;
                 p.$twbr.write(|w| unsafe { w.bits(twbr as u8) });
+
                 // Disable prescaler
+                //
+                // (datasheet) 22.5.2 Bit Rate Generator Unit
+                //
+                //      SCL frequency = CPU Clock Frequency
+                //                     --------------------------------
+                //                     16 + 2(TWBR) * (Prescaler Value)
+                //
+                // Setting the prescaler to 1 makes the math easy.
                 p.$twsr.write(|w| w.$twps().prescaler_1());
 
                 [<$I2c Master>] {
@@ -263,9 +280,11 @@ macro_rules! impl_twi_i2c {
                 speed: u32,
             ) -> [<$I2c Master>]<CLOCK, $crate::i2c::I2cFloating> {
                 // Calculate TWBR
+                // Same comments in new() above.
                 let twbr = ((CLOCK::FREQ / speed) - 16) / 2;
                 p.$twbr.write(|w| unsafe { w.bits(twbr as u8) });
                 // Disable prescaler
+                // Same comments in new() above.
                 p.$twsr.write(|w| w.$twps().prescaler_1());
 
                 [<$I2c Master>] {
--- a/avr-hal-generic/src/i2c.rs
+++ b/avr-hal-generic/src/i2c.rs
@@ -525,41 +525,6 @@ macro_rules! impl_twi_i2c {
             twgce: bool,
         }
 
-        use core::sync::atomic;
-
-        /// I2C_SLAVE_TRIGGER: true == set; false == clear
-        static I2C_SLAVE_TRIGGER: atomic::AtomicBool = atomic::AtomicBool::new(false);
-
-        fn is_i2c_slave_trigger_set()->bool{
-           avr_device::interrupt::free(|_cs| {
-                if I2C_SLAVE_TRIGGER.load(atomic::Ordering::SeqCst) {
-                    true
-                }else{
-                    false
-                }
-            })
-        }
-
-        fn is_i2c_slave_trigger_clear()->bool{
-            !is_i2c_slave_trigger_set()
-        }
-
-        fn clear_i2c_slave_trigger(){
-           avr_device::interrupt::free(|_cs| {
-                if I2C_SLAVE_TRIGGER.load(atomic::Ordering::SeqCst) {
-                    I2C_SLAVE_TRIGGER.store(false, atomic::Ordering::SeqCst)
-                }
-            });
-        }
-
-        fn set_i2c_slave_trigger(){
-           avr_device::interrupt::free(|_cs| {
-                if !I2C_SLAVE_TRIGGER.load(atomic::Ordering::SeqCst) {
-                    I2C_SLAVE_TRIGGER.store(true, atomic::Ordering::SeqCst);
-                }
-            })
-        }
-
         impl [<$I2c Slave>]<$crate::i2c::I2cPullUp>
         {
             /// Initialize the I2C bus
@@ -717,6 +682,7 @@ macro_rules! impl_twi_i2c {
 
         /// Transitions from Initialized to AddressMatched
         impl <M>[<$I2c SlaveStateInitialized>]<M> {
+
             /// Wait until TWINT has been triggered. TWINT is triggered if a message appears on
             /// the I2C bus and the address matches this slave's address, or if a message appears
             /// on the I2C bus and this slave has TWGCE enabled.
@@ -800,11 +766,9 @@ macro_rules! impl_twi_i2c {
                 match slave.p.twsr.read().tws().bits() {
                 |   $crate::i2c::twi_status::TW_SR_DATA_ACK
                 |   $crate::i2c::twi_status::TW_SR_GCALL_DATA_ACK
-                    => {
-                        set_i2c_slave_trigger();
-                        return Ok([<$I2c SlaveState>]::RxReady([<$I2c SlaveStateRxReady>]::<M> {
-                            slave: slave,
-                            }))},
+                    => return Ok([<$I2c SlaveState>]::RxReady([<$I2c SlaveStateRxReady>]::<M> {
+                        slave: slave,
+                        })),
                 |   $crate::i2c::twi_status::TW_SR_DATA_NACK
                 |   $crate::i2c::twi_status::TW_SR_GCALL_DATA_NACK
                 |   $crate::i2c::twi_status::TW_SR_STOP
@@ -840,11 +804,9 @@ macro_rules! impl_twi_i2c {
                 |   $crate::i2c::twi_status::TW_ST_SLA_ACK
                 |   $crate::i2c::twi_status::TW_ST_ARB_LOST_SLA_ACK
                 |   $crate::i2c::twi_status::TW_ST_DATA_ACK
-                    => {
-                        set_i2c_slave_trigger();
-                        return Ok([<$I2c SlaveState>]::TxReady([<$I2c SlaveStateTxReady>]::<M> {
-                            slave: slave,
-                            }))},
+                    => return Ok([<$I2c SlaveState>]::TxReady([<$I2c SlaveStateTxReady>]::<M> {
+                        slave: slave,
+                        })),
                 |   $crate::i2c::twi_status::TW_ST_DATA_NACK
                 |   $crate::i2c::twi_status::TW_ST_LAST_DATA
                     => return Ok([<$I2c SlaveState>]::Initialized([<$I2c SlaveStateInitialized>]::<M> {
--- a/avr-hal-generic/src/i2c.rs
+++ b/avr-hal-generic/src/i2c.rs
@@ -503,7 +503,8 @@ macro_rules! impl_twi_i2c {
         /// I2C Slave is similar to the I2C Master. The similarities are that both structs have the
         /// `new()` and `new_with_external_pullup()` constructors. However the Slave constructors
         /// do not need to pass the speed/bitrate as that is determined by the Master. However, the
-        /// Slave constructors do need to pass the slave address.
+        /// Slave constructors do need to pass the slave address and the flag to enable/disable the
+        /// General Call Address.
         $(#[$i2c_attr])*
         pub struct [<$I2c Slave>]<M> {
             /// Peripherals
@@ -651,27 +652,49 @@ macro_rules! impl_twi_i2c {
 
         /// Transitions from Initialized to AddressMatched
         impl <M>[<$I2c SlaveStateInitialized>]<M> {
+
+            /// Wait until TWINT has been triggered. TWINT is triggered if a message appears on
+            /// the I2C bus and the address matches this slave's address, or if a message appears
+            /// on the I2C bus and this slave has TWGCE enabled.
             pub fn wait(self
-            ) -> Result<[<$I2c SlaveStateAddressMatched>]<M>, $crate::i2c::Error>{
-                while self.slave.p.twcr.read().twint().bit_is_clear() { }
-                // TWINT has been triggered, meaning the address matched or
-                // we are responding to a general call address
-                Ok([<$I2c SlaveStateAddressMatched>]::<M> {
-                    slave: self.slave,
-                })
+            ) -> Result<[<$I2c SlaveState>]<M>, $crate::i2c::Error>{
+                while self.slave.p.twcr.read().twint().bit_is_clear() { };
+                match self.slave.p.twsr.read().tws().bits() {
+                    $crate::i2c::twi_status::TW_SR_SLA_ACK
+                |   $crate::i2c::twi_status::TW_SR_ARB_LOST_SLA_ACK
+                |   $crate::i2c::twi_status::TW_SR_ARB_LOST_GCALL_ACK
+                    => return Ok([<$I2c SlaveState>]::AddressMatched([<$I2c SlaveStateAddressMatched>]::<M> {
+                        slave: self.slave,
+                        })),
+                    $crate::i2c::twi_status::TW_SR_GCALL_ACK
+                    => {
+                        return if self.slave.twgce {
+                            Ok([<$I2c SlaveState>]::AddressMatched([<$I2c SlaveStateAddressMatched>]::<M> {
+                                slave: self.slave,
+                            }))
+                        }else{
+                            Ok([<$I2c SlaveState>]::Initialized([<$I2c SlaveStateInitialized>]::<M> {
+                                slave: self.slave,
+                            }))
+                        }
+                    },
+                    _ => {
+                        return Err($crate::i2c::Error::Unknown);
+                    },
+                }
             }
         }
 
         /// Transitions from AddressMatched to RxReady | TxReady
         impl <M>[<$I2c SlaveStateAddressMatched>]<M> {
+
+            /// Transition to the next state based on the data direction bit
             pub fn next(self
             ) -> Result<[<$I2c SlaveState>]<M>, $crate::i2c::Error>{
                 match self.slave.p.twsr.read().tws().bits() {
-                // TODO: impl
-                //     $crate::i2c::twi_status::TW_SR_SLA_ACK
-                // |   $crate::i2c::twi_status::TW_SR_ARB_LOST_SLA_ACK
-                // |   $crate::i2c::twi_status::TW_SR_GCALL_ACK
-                // |   $crate::i2c::twi_status::TW_SR_ARB_LOST_GCALL_ACK =>
+                // |   $crate::i2c::twi_status::TW_SR_DATA_ACK
+                // |   $crate::i2c::twi_status::TW_SR_GCALL_DATA_ACK
+                // =>
                 //         self.p.twcr.write(|w| w
                 //             .twsto().clear_bit()
                 //             .twint().set_bit();
--- a/avr-hal-generic/src/i2c.rs
+++ b/avr-hal-generic/src/i2c.rs
@@ -1,20 +1,11 @@
 //! I2C Implementations
 
-// datasheet : ATmega48A-PA-88A-PA-168A-PA-328-P-DS-DS40002061B.pdf
-//
 /// TWI Status Codes
 pub mod twi_status {
     // The status codes defined in the C header are meant to be used with the
     // masked status value: (TWSR & TW_STATUS_MASK).  In our case, svd2rust
     // already added code to shift it to just the status value, so all status
     // codes need to be shifted to the right as well.
-    //
-    // See also:
-    //
-    // (datasheet)  Table 22-2. Status codes for Master Transmitter Mode
-    //              Table 22-3. Status codes for Master Receiver Mode
-    //              Table 22-4. Status Codes for Slave Receiver Mode
-    //              Table 22-5. Status Codes for Slave Transmitter Mode
 
     /// Start condition transmitted
     pub const TW_START: u8 = 0x08 >> 3;
@@ -231,29 +222,11 @@ macro_rules! impl_twi_i2c {
         impl<CLOCK, M> [<$I2c Master>]<CLOCK, M>
             where CLOCK: $crate::clock::Clock, {
             fn initialize_i2c( p: &$I2C, speed: &u32,){
-                // Calculate the TWI Bit Rate Register (TWBR)
-                //
-                // (datasheet) 22.5.2 Bit Rate Generator Unit
-                //
-                //      This unit controls the period of SCL when operating in a Master mode. The
-                //      SCL period is controlled by settings in the TWI Bit Rate Register (TWBR)
-                //      and the Prescaler bits in the TWI Status Register (TWSR). Slave operation
-                //      does not depend on Bit Rate or Prescaler settings, but the CPU clock
-                //      frequency in the Slave must be at least 16 times higher than the SCL
-                //      frequency. Note that slaves may prolong the SCL low period, thereby
-                //      reducing the average TWI bus clock period.
+                // calculate the bit rate
                 let twbr = ((CLOCK::FREQ / speed) - 16) / 2;
                 p.$twbr.write(|w| unsafe { w.bits(twbr as u8) });
 
-                // Disable prescaler
-                //
-                // (datasheet) 22.5.2 Bit Rate Generator Unit
-                //
-                //      SCL frequency = CPU Clock Frequency
-                //                     --------------------------------
-                //                     16 + 2(TWBR) * (Prescaler Value)
-                //
-                // Setting the prescaler to 1 makes the math easy.
+                // disable the prescaler
                 p.$twsr.write(|w| w.$twps().prescaler_1());
             }
         }
--- a/avr-hal-generic/src/i2c.rs
+++ b/avr-hal-generic/src/i2c.rs
@@ -687,12 +687,12 @@ macro_rules! impl_twi_i2c {
        }
        pub struct [<$I2c SlaveStateError>];
 
-        // // Slave State Machine
-        // pub struct [<$I2c SlaveStateMachine>]<M>{
-        //    slave: [<$I2c Slave>]<M>,
-        //    state: [<$I2c SlaveState>],
-        // }
-        //
+        // Slave State Machine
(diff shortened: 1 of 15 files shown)


avr-hal-generic/src/i2c.rs
```diff
@@ -499,5 +499,82 @@ macro_rules! impl_twi_i2c {
                 Ok(())
             }
         }
+
+        $(#[$i2c_attr])*
```
ToddG, line 503: Need clarification on what `$(#[$i2c_attr])*` does. Looks like an attribute that expands?

avr-hal-generic/src/i2c.rs
```diff
@@ -499,5 +499,82 @@ macro_rules! impl_twi_i2c {
...
+        pub struct Slave<CLOCK: $crate::clock::Clock, M> {
```
ToddG, line 504: The previous suggestion of `$I2cSlave` for a name was causing conflicts, so I used `Slave` temporarily...

avr-hal-generic/src/i2c.rs
```diff
@@ -499,5 +499,82 @@ macro_rules! impl_twi_i2c {
...
+            pub fn new(
+                p: $I2C,
+                sda: $sdamod::$SDA<$crate::port::mode::Input<$crate::port::mode::PullUp>>,
+                scl: $sclmod::$SCL<$crate::port::mode::Input<$crate::port::mode::PullUp>>,
+                speed: u32,
+                address: u8,
+            ) -> Slave<CLOCK, $crate::i2c::I2cPullUp> {
+                // Calculate TWBR
+                let twbr = ((CLOCK::FREQ / speed) - 16) / 2;
+                p.$twbr.write(|w| unsafe { w.bits(twbr as u8) });
```
ToddG, line 532: I need to research what TWBR does, esp. in the context of an `i2c slave`.

Rahix: TWBR is the `baudrate` or clock speed of the I2c bus.  The slave is not making such decisions as the clock is driven by the master thus I don't think TWBR is used for slave operation (Hate the master/slave teminology here but until a different naming scheme for I2C becomes popular we'll have to stick with it I guess ...).

avr-hal-generic/src/i2c.rs
Rahix: As explained in [my comment](https://github.com/Rahix/avr-hal/issues/90#issuecomment-716125402) this is needed for the doc-string and other type attributes.  This is also recommended by the rust API guidelines as [`C-MACRO-ATTR`](https://rust-lang.github.io/api-guidelines/macros.html#item-macros-compose-well-with-attributes-c-macro-attr).

avr-hal-generic/src/i2c.rs
```diff
@@ -499,5 +499,82 @@ macro_rules! impl_twi_i2c {
...
+            /// `new()` will enable the internal pull-ups to comply with the I2C
+            /// specification.  If you have external pull-ups connected, please
+            /// use `new_with_external_pullup()` instead.
+            ///
+            /// TODO: verify whether we need the speed param (bitrate)
+            pub fn new(
+                p: $I2C,
+                sda: $sdamod::$SDA<$crate::port::mode::Input<$crate::port::mode::PullUp>>,
+                scl: $sclmod::$SCL<$crate::port::mode::Input<$crate::port::mode::PullUp>>,
+                speed: u32,
```
Rahix, line 527: I think the speed parameter can be dropped here as the bus master is responsible for the clock.

avr-hal-generic/src/i2c.rs
```diff
@@ -201,7 +206,8 @@ macro_rules! impl_twi_i2c {
                 data: $twdr:ident,
             },
         }
-    ) => {
+    ) => {$crate::paste::paste! {
+        /// I2C Master
         $(#[$i2c_attr])*
         pub struct $I2c<CLOCK: $crate::clock::Clock, M> {
```
Rahix, line 212: While we're at it, should we rename this one to `[< $I2c Master>]`?  Maybe we should check what other HALs that support both modes, are naming them ...

ToddG: Cool, I was thinking the same thing. I'll look around and see what's common.

avr-hal-generic/src/i2c.rs
```diff
@@ -570,5 +570,16 @@ macro_rules! impl_twi_i2c {
...
             }
         }
+
+        impl<CLOCK, M> [<$I2c Slave>]<CLOCK, M>
+        where
+            CLOCK: $crate::clock::Clock,
+        {
+            pub fn start(&mut self) {
+                // this does not compile
+                // self.p.$twar.write(|w| unsafe {w.bits(self.address)});
```
Rahix, line 580: Does not compile because the macro token `$twar` is not defined anywhere.  You'll need to add it in the macro pattern , somewhere e.g.:

```diff
 registers: {
     ...
     bitrate: $twbr:ident,
+    address: $twar:ident,
     data: $twdr:ident,
 }
```

And of course add this to the macro invocations sites.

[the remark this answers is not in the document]
Rahix: let's go through the line step by step:

1. `p` is the peripheral.  It's type is passed in from the macro-invocation and e.g. for `ATmega328P` will end up being [`avr_device::atmega328p::TWI`](https://docs.rs/avr-device/0.2.3/avr_device/atmega328p/struct.TWI.html).  This type "derefs" to [`&avr_device::atmega328p::twi::RegisterBlock`](https://docs.rs/avr-device/0.2.3/avr_device/atmega328p/twi/struct.RegisterBlock.html) which is where fields for all the registers are found.
2. `$twsr` is a macro token that the macro-invocation gives a proper register name for.  This exists because the registers might have different names on different MCUs.  For `ATmega328P`, it will just be `twsr`.  The same story with the `$twps` field name for a bit-field in the register.  The line will end up being expanded to
   ```rust
   p.twsr.write(|w| w.twps().prescaler_1()); 
   ```
   for `ATmega328P`.
3. The `.write()` method is part of the `svd2rust` peripheral API.  Read more about it in [its documentation](https://docs.rs/svd2rust/0.17.0/svd2rust/#read--modify--write-api).  Essentially, this line says: Write the whole register to its default values _and_ set the prescaler bits to the bit-pattern for `PRESCALER1`.
4. `w.$twps()` means "take the TWPS field in this register"
5. and `$twps().prescaler_1()` means "write the bit-pattern for `PRESCALER1` to it".  This is also part of the autogenerated API ([link 1](https://docs.rs/avr-device/0.2.3/avr_device/atmega328p/twi/twsr/struct.W.html), [link 2](https://docs.rs/avr-device/0.2.3/avr_device/atmega328p/twi/twsr/struct.TWPS_W.html)).

Hope this helps ...

ToddG: Did I delete your comments? If so, sorry! Re-pasting back in here.

----
From Rahix
----

In avr-hal-generic/src/i2c.rs:

> @@ -570,5 +570,16 @@ macro_rules! impl_twi_i2c {
                 }
             }
         }
+
+        impl<CLOCK, M> [<$I2c Slave>]<CLOCK, M>
+        where
+            CLOCK: $crate::clock::Clock,
+        {
+            pub fn start(&mut self) {
+                // this does not compile
+                // self.p.$twar.write(|w| unsafe {w.bits(self.address)});

let's go through the line step by step:

    p is the peripheral. It's type is passed in from the macro-invocation and e.g. for ATmega328P will end up being avr_device::atmega328p::TWI. This type "derefs" to &avr_device::atmega328p::twi::RegisterBlock which is where fields for all the registers are found.
    $twsr is a macro token that the macro-invocation gives a proper register name for. This exists because the registers might have different names on different MCUs. For ATmega328P, it will just be twsr. The same story with the $twps field name for a bit-field in the register. The line will end up being expanded to

    p.twsr.write(|w| w.twps().prescaler_1()); 

    for ATmega328P.
    The .write() method is part of the svd2rust peripheral API. Read more about it in its documentation. Essentially, this line says: Write the whole register to its default values and set the prescaler bits to the bit-pattern for PRESCALER1.
    w.$twps() means "take the TWPS field in this register"
    and $twps().prescaler_1() means "write the bit-pattern for PRESCALER1 to it". This is also part of the autogenerated API (link 1, link 2).

Hope this helps ...

Rahix: Wow, didn't even know it was possible to delete other peoples comments :D

ToddG: So here's what I see so far...

1. `peripheral` -> `crate::pac::TWI` -> `RegisterBlock`
1. `RegisterBlock` contains all the TWI thingies, like `TWSR`

And in this generated file `avr-device/src/devices/atmega328p/twi.rs`, the TWI thingies are defined, like this:

```rust
pub type TWSR = crate::Reg<u8, _TWSR>;
#[allow(missing_docs)]
#[doc(hidden)]
pub struct _TWSR;
#[doc = "`read()` method returns [twsr::R](twsr::R) reader structure"]
impl crate::Readable for TWSR {}
#[doc = "`write(|w| ..)` method takes [twsr::W](twsr::W) writer structure"]
impl crate::Writable for TWSR {}
#[doc = "TWI Status Register"]
pub mod twsr;
```

And the part that I was missing was the implementation of the `crate::Readable` and `crate::Writeable`.

I think I'm starting to see how this is wired together now.

ToddG: I'm still a little confused as to this wiring...how does `crate::pac::TWI` deref to become `RegisterBlock`?

Yeah, the message that popped up was, 'the repository owner has deleted a comment'. I thought I had double posted a comment and was deleting the duplicate. But it turned out to be your comment. Totally weird that I could do that.

Rahix: > how does crate::pac::TWI deref to become RegisterBlock?

There's an

```rust
impl Deref for TWI {
	type Target = twi::RegisterBlock;

	...
}
```

somewhere in the auto-generated code.

avr-hal-generic/src/i2c.rs
```diff
@@ -210,14 +210,14 @@ macro_rules! impl_twi_i2c {
     ) => {$crate::paste::paste! {
         /// I2C Master
         $(#[$i2c_attr])*
-        pub struct $I2c<CLOCK: $crate::clock::Clock, M> {
+        pub struct [<$I2c Master>]<CLOCK: $crate::clock::Clock, M> {
```
Rahix: > Error:
>
> ```text
>    error[E0422]: cannot find struct, variant or union type `I2c` in this scope
>      --> chips/atmega328p-hal/src/lib.rs:31:20
>       |
>    31 |         pub struct I2c {
>       |                    ^^^ not found in this scope
> ```
>
> I don't understand something. I thought that the chips were defining the `I2c` struct and then passing it into the macro for expansion. The macro would then match against the struct elements, and then expand them in the macro expansion. If the chips defined the I2c struct, then this error would not be here. So where is the I2c struct defined? Here? Or in the macro?

Macro errors are often very confusing, be very careful about interpreting them as is.

The syntax inside the macro, that is, everything between `impl_twi_i2c!{` and `}` is _not_ interpreted by the rust compiler at all, so nothing is defined here.  The tokens are passed to the macro which then generates _something_ from them.  This result is the only thing the rust compiler sees.

Next, the `I2c` is a macro token that gets used in multiple places in the expansion.  What the error tells you is that somewhere the token is used but no type of that name is found.  That makes sense, because you've changed the macro to only generate a struct named `$I2c Master` and `$I2c Slave` (= `I2cMaster` & `I2cSlave`).  So you've probably forgot to change some reference to the name somewhere ...

For example, in line 241 and 271, you still reference a type `$I2c` instead of `[<$I2c Master>]`.  Maybe take another look if there's more places that need to be changed.

ToddG: Ah, that really helped. Thx!

Rahix: Pretty much.  One more step for completeness:

`atf2svd` generates -> `svd` files -> **`svdtools` patches** -> `svd2rust` generates

The supplied ATDF files are ... substandard to say the least and there's a lot of registers/fields that need to be fixed up before we can feed them to svd2rust.  That's all found in [`avr-device/path`](https://github.com/Rahix/avr-device/tree/master/patch) if you're curious.

avr-hal-generic/src/i2c.rs
```diff
@@ -1,11 +1,22 @@
 //! I2C Implementations
 
+// datasheet : ATmega48A-PA-88A-PA-168A-PA-328-P-DS-DS40002061B.pdf
```
ToddG, line 3: I added a bunch of datasheet excerpts as comments as a way to help me understand what's going on. Personally, I like lot's of explanatory comments, but of course, it's up to you to decide if you want to keep them in the final PR.

avr-hal-generic/src/i2c.rs
```diff
@@ -217,23 +228,19 @@ macro_rules! impl_twi_i2c {
             scl: $sclmod::$SCL<M>,
         }
 
-        impl<CLOCK, M> [<$I2c Master>]<CLOCK, M>
-        where
-            CLOCK: $crate::clock::Clock,
-        {
-            fn initialize_i2c(
```
ToddG: Added a static method for initializing TWI.

avr-hal-generic/src/i2c.rs
```diff
@@ -329,51 +336,110 @@ macro_rules! impl_twi_i2c {
...
-                //    condition to claim the bus Master status. TWSTA must be cleared by software when the START condition has
-                //    been transmitted.
+                //    The application writes the TWSTA bit to one when it desires to become a
+                //    Master on the 2-wire Serial Bus. The TWI hardware checks if the bus is
+                //    available, and generates a START condition on the bus if it is free. However,
+                //    if the bus is not free, the TWI waits until a STOP condition is detected, and
+                //    then generates a new START condition to claim the bus Master status. TWSTA
+                //    must be cleared by software when the START condition has been transmitted.
                 //
                 //    NOTE: `twstart` macro token is defined as `twsta` in chips/[CHIP]/src/lib.rs
```
ToddG, line 375: The naming difference tripped me up here until I realized twstart === twsta.

avr-hal-generic/src/i2c.rs
```diff
@@ -329,51 +336,110 @@ macro_rules! impl_twi_i2c {
...
                 self.p.$twcr.write(|w| w
                     .$twen().set_bit()
                     .$twint().set_bit()
                     .$twstart().set_bit()
                 );
-                // NOTE: why do we need to wait here? `wait()` is just reading twcr.twint to see if
-                // the bit is set...but it was just set in the line above...
+
+                // QUESTION: why do we need to wait here? `wait()` is just reading twcr.twint to see if
```
ToddG, line 382: Why do we need the wait() here?

avr-hal-generic/src/i2c.rs
```diff
@@ -329,51 +336,110 @@ macro_rules! impl_twi_i2c {
...
+                //
+                //    TWDR In Transmit mode, TWDR contains the next byte to be transmitted. In
+                //    Receive mode, the TWDR contains the last byte received. It is writable while
+                //    the TWI is not in the process of shifting a byte. This occurs when the TWI
+                //    Interrupt Flag (TWINT) is set by hardware.
                 self.wait();
 
                 // Validate status
+                //
+                // NOTE: here's what this impl looks like in lpc8xx-hal, modeled as a state
```
ToddG, line 399: I like the class state machine pattern in rust. Just showing what lpc8xx does here. Perhaps we can adopt this pattern?

avr-hal-generic/src/i2c.rs
```diff
@@ -389,9 +455,32 @@ macro_rules! impl_twi_i2c {
...
+                //    Interrupt Flag (TWINT) is set by hardware. Note that the Data Register cannot
+                //    be initialized by the user before the first interrupt occurs. The data in
+                //    TWDR remains stable as long as TWINT is set. While data is shifted out, data
+                //    on the bus is simultaneously shifted in. TWDR always contains the last byte
+                //    present on the bus, except after a wake up from a sleep mode by the TWI
+                //    interrupt. In this case, the contents of TWDR is undefined. In the case of a
+                //    lost bus arbitration, no data is lost in the transition from Master to Slave.
+                //    Handling of the ACK bit is controlled automatically by the TWI logic, the CPU
+                //    cannot access the ACK bit directly.
                 self.p.$twdr.write(|w| unsafe { w.bits(rawaddr) });
```
ToddG, line 484: This code makes sense to me, though I have questions about `transact()`

avr-hal-generic/src/i2c.rs
```diff
@@ -425,51 +514,57 @@ macro_rules! impl_twi_i2c {
...
-                //        }
-                //        /// Returns `true` if the bit is set (1).
-                //        #[inline(always)]
-                //        pub fn bit_is_set(&self) -> bool {
-                //            self.bit()
-                //        }
-                //    }
+                // And the FieldReader also explicitly uses 'clear' to refer to 0 and 'set' to
+                // refer to 1:
                 //
```
ToddG, line 538: So basically, I was just trying to understand this weird over-use of the terms `clear` and `set`.

avr-hal-generic/src/i2c.rs
```diff
@@ -425,51 +514,57 @@ macro_rules! impl_twi_i2c {
...
+                //    if the bit is clear (0).  #[inline(always)] pub fn bit_is_clear(&self) ->
+                //    bool { !self.bit() } /// Returns `true` if the bit is set (1).
+                //    #[inline(always)] pub fn bit_is_set(&self) -> bool { self.bit() } }
+                //
+                // QUESTION: so is what I'm seeing a distinction between a _HARDWARE FLAG_ being 'cleared',
+                // (perhaps 'reset' is a better name?) vs. setting/clearing a _bit_ in a bitfield?
+                while self.p.$twcr.read().$twint().bit_is_clear() { } }
 
             fn transact(&mut self) {
+                // QUESTION: why are we setting TWEN here? It has already been set (above). Does it
```
ToddG, line 550: We set `$twen` with every other bit write. I suppose it doesn't hurt, but I'm curious why it is necessary. Is this to guard against downstream code setting this flag and breaking this library?

avr-hal-generic/src/i2c.rs
```diff
@@ -425,51 +514,57 @@ macro_rules! impl_twi_i2c {
...
+                // need to be set ? Why would it be unset? I looked through the datasheet and did
+                // not see any mention of this being turned off by hardware.
                 self.p.$twcr.write(|w| w.$twen().set_bit().$twint().set_bit());
+                // QUESTION: how do tight loops work on microcontrollers? Does this prevent the
+                // cpu from shutting down?
                 while self.p.$twcr.read().$twint().bit_is_clear() { }
```
ToddG, line 556: Do tight loops like this turn the CPU into a heater?

avr-hal-generic/src/i2c.rs
```diff
@@ -497,9 +592,27 @@ macro_rules! impl_twi_i2c {
...
+                        //    pulse is generated on the TWI bus if the following conditions are met:
+                        //    1. The device’s own slave address has been received.
+                        //    2. A general call has been received, while the TWGCE bit in the TWAR is set.
+                        //    3. A data byte has been received in Master Receiver or Slave Receiver mode.
+                        //    By writing the TWEA bit to zero, the device can be virtually disconnected from the 2-wire Serial Bus temporarily.
+                        //    Address recognition can then be resumed by writing the TWEA bit to one again.
                         self.p.$twcr.write(|w| w.$twint().set_bit().$twen().set_bit().$twea().set_bit());
                         self.wait();
                     } else {
+                        // QUESTION: Why is this behaviour different for the last byte?
```
ToddG, line 609: Curious as to the bifurcation when writing the byte array. Why is `$twea` not set for the final byte? I thought the final byte required an ACK as well.

avr-hal-generic/src/i2c.rs
```diff
@@ -497,9 +592,27 @@ macro_rules! impl_twi_i2c {
                 let last = buffer.len() - 1;
                 for (i, byte) in buffer.iter_mut().enumerate() {
```
ToddG, line 593: Perhaps I'm blind, but I do not see where the buffer or the byte are being used...isn't the idea that we are reading the bytes out of the TWI and placing into the buffer? I don't see where that is happening. It _looks_ like we are just cycling through the buffer and just signalling that we can consume the next byte...

ToddG (commented):
@Rahix -- I left a bunch of random questions in here as to the current I2C Master implementation.

avr-hal-generic/src/i2c.rs
Rahix: Register bits don't necessarily act like memory cells.  The following things which you might assume from a memory cell do _not_ always apply for register bits:

- Reading a register twice without any writes in between will yield the same value.

  _Counter Examples_:
  - A FIFO register will return the next byte from the internal buffer on each read.  I.e. the first read will yield the first byte, the second read will yield the second byte, etc.
  - Sometimes reading a bit will reset it back to zero.

- Writing a value to a register and reading it back will give back the same value.

  _Counter Examples_: In hardware, the logic that receives data written to a register is entirely disjunct from the logic presenting data when the register is read.  There are many different ways in which this shows:

  - Sometimes only a subset of bits in the register are writeable (=read-only).  All other bits from the written value will be ignored.
  - Some bits are write-only and reading will always return 0.
  - Sometimes writing a 1 to a bit will reset it back to zero.
  - Sometimes a bits value is entirely dependent on hardware state and writing it has a completely different effect than what reading is for (bits are "reused" for two different purposes).

---

In this specific case, we write `twint` to 1 which will clear the flag:

> The TWINT Flag must be cleared by software by writing a logic one to it.

Then we `wait()` until hardware sets it again.

avr-hal-generic/src/i2c.rs
Rahix: This is pretty standard terminology in the bare-metal world :D

Not sure how familiar you are with digital logic, but think about the `TWINT` bit like a logic buffer/flip-flop.  Different signals will do different things to the flip-flop:

- Reading the bit in the register just presents the flip-flop output signal (either high or low = 1 or 0).
- If the CPU writes a 1 to the bit in the register, the flip-flop will be reset -> set to 0 -> it will be _cleared_.
- Certain hardware conditions will set the flip-flop to 1 -> _set_ the bit.
- While the flip-flop is _set_ (logic high on its output), SCL is forced low (no clock pulses).
- On falling edge of the flip-flop output (transition from high to low), TWI will start operation (depending on some other circumstances).

Also see my other comment above about hardware registers in general.

avr-hal-generic/src/i2c.rs
Rahix: Simplifying a bit, you can only ever write a whole register, in this case, all 8 bits at once.  For toggling only certain bits, you'd have to first read the entire register, modify its bits with some logic and and or operations and then write the modified value back.  This is also called a _read-modify-write_ loop.

The `svd2rust` API specifies that a `.write(...)` will write the _whole_ register.  This means all 8 bits are written.  Any bit that is not explicitly specified will be set to its "reset"-value which is usually just 0.

Thus, if we were to not set `TWEN` to 1 every time, we'd actually be turning off the TWI peripheral.

---

Now, you could of course use `.modify(...)` which is `svd2rust`'s version of the _read-modify-write_ loop.  Then we'd only need to care about the bits we're actually interacting with.

This does, however, have a number of disadvantages:

1. Its at least 3 more instructions per write, which can amount to quite a large overhead in the big picture.
2. Because we leave some bits untouched, it can quickly lead to hard-to-discover bugs because some bit was still set where it shouldn't.  By keeping it explicit which bits are set and which are cleared for every write, it is much easier to see such circumstances.
3. For a single 8-bit register, the `.write()` operation is atomic.  A _read-modify-write_ loop however is not, which means it might need a critical section to prevent an interrupt firing in the middle.

avr-hal-generic/src/i2c.rs
Rahix: Short answer: no

- First of all, this loop will only run for a very short time, not milliseconds or seconds.
- Microcontrollers generally support continuous execution of instructions for long periods of time because that's what they are usually meant to do.  There are some power-optimizations where you can put the MCU "to sleep" until an interrupt occurs but those are only used when you expect to get at least a few milliseconds of sleep time before the next interrupt.
- The reason you hear the fan spin up on a desktop machine when running such code for long periods of time is that your OS includes such optimizations and puts CPUs into idle-states when not in use.  Then, when someone is using 100% CPU, this no longer happens and the CPU runs continuously -> it heats up.  Even more, the OS might have a power governor that ramps up the clock speed when more CPU resources are used, leading to even more thermal.

In embedded/bare-metal, these tight loops waiting for hardware are very common.  They are by far the most efficient way to wait for a hardware even that is expected to arrive very soon.  They are even used to wait for software in some cases, when latency is critical (so called spin-locks).

avr-hal-generic/src/i2c.rs
Rahix: Last line of the loop, the received byte is written into the buffer:

https://github.com/Rahix/avr-hal/blob/9f9d6972e4ac7c149e2644bbc52c47cca81838f4/avr-hal-generic/src/i2c.rs#L415

avr-hal-generic/src/i2c.rs
Rahix: The datasheet text you insterted below explains it:  After the last byte, master send NACK and then a stop condition.

avr-hal-generic/src/i2c.rs
ToddG: Great comment. Do you mind if I capture this in a Newbie FAQ?

avr-hal-generic/src/i2c.rs
ToddG: Thx!

avr-hal-generic/src/i2c.rs
ToddG: This is like an intro course into micro-controllers. Thx!

avr-hal-generic/src/i2c.rs
ToddG: I see why I was confused now.

So we have read the last byte in `read_data()`, and we set TWINT and TWEN high:

https://github.com/Rahix/avr-hal/blob/9f9d6972e4ac7c149e2644bbc52c47cca81838f4/avr-hal-generic/src/i2c.rs#L397

Then call `wait()` which waits for TWINT to be cleared:

https://github.com/Rahix/avr-hal/blob/9f9d6972e4ac7c149e2644bbc52c47cca81838f4/avr-hal-generic/src/i2c.rs#L356

Then wherever `read_data()` was invoked from can invoke a STOP, like here in `blocking_read()`:

https://github.com/Rahix/avr-hal/blob/9f9d6972e4ac7c149e2644bbc52c47cca81838f4/avr-hal-generic/src/i2c.rs#L477

The `stop()` is here:

https://github.com/Rahix/avr-hal/blob/9f9d6972e4ac7c149e2644bbc52c47cca81838f4/avr-hal-generic/src/i2c.rs#L420

My question has evolved to:

1. (minor question) I don't see a `restart` method on the API... is that intentional?
1. The datasheet defines the STOP as being invoked exactly as in the `stop()` method shown. My question is why the intervening writes? Here's what I see:

a) write TWINT TWEN high
b) wait for TWINT to clear
c) write TWINT TWEN TWSTO high

Why do we need `a` and `b`?

avr-hal-generic/src/i2c.rs
ToddG: Ah, I _was_ blind. Thx :-)

ToddG:
Additional question:

I have read up on PhantomData [here](https://doc.rust-lang.org/core/marker/struct.PhantomData.html) and [here](https://doc.rust-lang.org/nomicon/phantom-data.html), but I'm still curious how it is being used here:

https://github.com/Rahix/avr-hal/blob/9f9d6972e4ac7c149e2644bbc52c47cca81838f4/avr-hal-generic/src/i2c.rs#L208

* Why is there a phantomdata marked type here at all?
* Why is the phantom field named `_clock`?

Rahix:
> Why is there a phantomdata marked type here at all?

Because you cannot have a generic type parameter (`CLOCK`) which is not used.  The `PhantomData<CLOCK>` is a zero-sized type which counts as a "use" of the type paramter, thus it is legal.

> Why is the phantom field named `_clock`?

To indicate that the field does not really hold any use, I've prefixed it with an underscore.

ToddG:
Regarding the CLOCK phantom... yeah, I've been playing around with how it's used and re-reading the docs, etc. I think I understand it now.

Thx!

avr-hal-generic/src/i2c.rs
```diff
@@ -548,21 +548,18 @@ macro_rules! impl_twi_i2c {
         /// I2C Slave is similar to the I2C Master. The similarities are that both structs have the
         /// `new()` and `new_with_external_pullup()` constructors. However the Slave constructors
         /// do not need to pass the speed/bitrate as that is determined by the Master. However, the
-        /// Slave constructors do need to pass the slave address. 
+        /// Slave constructors do need to pass the slave address.
         ///
         /// Note: Only the 7 LSB bits of the address field are used.
         $(#[$i2c_attr])*
-        pub struct [<$I2c Slave>]<CLOCK: $crate::clock::Clock, M> {
```
ToddG: Slaves don't need a clock, so removed this here and in consumer code.

avr-hal-generic/src/i2c.rs
```diff
@@ -599,73 +593,137 @@ macro_rules! impl_twi_i2c {
...
-                self.p.$twar.write(|w| unsafe {w.bits(self.address)});
-                self.p.$twcr.write(|w| w
-                    .$twen().set_bit()
-                    .$twea().set_bit()
-                    .$twsta().clear_bit()
-                    .$twsto().clear_bit()
-                    .$twint().set_bit()
-                );
+
+        // Slave State Machine
```
ToddG, line 607: I wrote a simplified version of what `lpc8xx-hal` has. I may eventually end up with exactly what they wrote, but I'm just trying to build up from scratch with the simplest API possible. I researched other popular rust state machine patterns, and settled on this.

I like the idea of using the compiler and type checker to validate the state transitions, but this looks ugly to me:

https://hoverbear.org/blog/rust-state-machine-pattern/

I like the all transitions listed in one place as here:

https://dev.to/werner/practical-rust-web-development-state-machine-2jlc

But I prefer to have use types with explicit methods on them as here:

     TypeA.fooMethod() -> TypeB
     TypeB.barMethod() -> TypeC

etc.

I'm just stubbing the methods out here to get your thoughts before I flesh out the implementation.

avr-hal-generic/src/i2c.rs
```diff
@@ -599,73 +593,137 @@ macro_rules! impl_twi_i2c {
...
+       impl [<$I2c SlaveStateTxReady>]{
+           pub fn write(_: u8) -> [<$I2c SlaveState>] {
+                // returns [<$I2c SlaveStateTxReady>] | [<$I2c SlaveStateError>]
+                [<$I2c SlaveState>]::Error([<$I2c SlaveStateError>])
+           }
+       }
+       pub struct [<$I2c SlaveStateError>];
+
+
+//       impl [<$I2c Slave>]
```
ToddG, line 681: Couple ways to go here...

    Slave.init() -> the SlaveState::Initialized
 
or

    SlaveState::Unitialized.init(slave) -> SlaveState::Initialized

ToddG (commented):
Looking for API feedback.

avr-hal-generic/src/i2c.rs
Rahix: I quite like what the kernel has to say about commenting: https://www.kernel.org/doc/html/latest/process/coding-style.html#commenting

I.e. the HOW should be obvious from the code and shouldn't need a datasheet excerpt alongside it.  For tricky stuff, mentioning the relevant section of the datasheet might make sense but I'd never copy the contents into the code -> if I need them, I can just open the data-sheet.

avr-hal-generic/src/i2c.rs
Rahix: This is a good idea!  Though maybe put it into the existing impl block further down instead of adding a second one up top?  You could also take the pattern from the `serial` module and make the initializer a method instead of a static function:

https://github.com/Rahix/avr-hal/blob/7d851d50b2485257a2cbede991e8de4c734053d8/avr-hal-generic/src/serial.rs#L139-L173

avr-hal-generic/src/i2c.rs
Rahix: Well, this is true for the entirety of embedded software development so I don't think `avr-hal` is the right place for it.  But if you want to post it somewhere else, please go a head :)

avr-hal-generic/src/i2c.rs
Rahix: Ah, this is all a bit messy ... I think we can get away with removing the register name macro tokens entirely like we did for the SPI driver.  Do you want me to do it and then you rebase your code ontop or do you want to take a stab at it yourself?

avr-hal-generic/src/i2c.rs
ToddG: Funny, I was coming around to this opinion myself in the past few days. Adding in the comments helped me think about the problem for awhile, but then they started getting in the way. It _would_ be nice if the datasheets were deep-linkable, so that I could reference the line number of the spec. I'll check in my latest changes in a few minutes.

avr-hal-generic/src/i2c.rs
ToddG: Cool, will do.

avr-hal-generic/src/i2c.rs
ToddG: Gotcha. I'll probably post a summary of this experience on my blog. No one reads it, but perhaps it'll help someone someday...probably me in 6 months :-)

avr-hal-generic/src/i2c.rs
ToddG: I'll take a run at it. I think I understand the pattern in SPI driver.

avr-hal-generic/src/i2c.rs
Rahix: > It would be nice if the datasheets were deep-linkable

Yeah, it really would ... Best one can do is name the section _and_ revision of the datasheet (because some lunatics think it is a good idea to renumber the sections from time to time ...)

> Adding in the comments helped me think about the problem for awhile, but then they started getting in the way.

In the C world some people have adopted a nice method: Use `//` comments for any temporary in-development comments while thinking/developing/debugging.  Use `/* ... */` comments for "production" comments that are meant to stay.  Then, before submitting changes, remove all `//` comments.

In Rust, this doesn't really work because people just use `//` for all comments everywhere.  So I personally have started to just use some comment prefix to mark 'dev-comments', e.g. `// WIP: ...`.

avr-hal-generic/src/i2c.rs
Rahix: Essentially, remove this block:

```diff
 #[macro_export]
 macro_rules! impl_twi_i2c {
     (
         $(#[$i2c_attr:meta])*
         pub struct $I2c:ident {
             peripheral: $I2C:ty,
             pins: {
                 sda: $sdamod:ident::$SDA:ident,
                 scl: $sclmod:ident::$SCL:ident,
             },
-            registers: {
-                control: $twcr:ident {
-                    enable: $twen:ident,
-                    ack: $twea:ident,
-                    int: $twint:ident,
-                    start: $twstart:ident,
-                    stop: $twstop:ident,
-                },
-                status: $twsr:ident {
-                    prescaler: $twps:ident,
-                    status: $tws:ident,
-                },
-                bitrate: $twbr:ident,
-                data: $twdr:ident,
-            },
         }
     ) => {
```

and then replace all of the `$...` register/field identifiers with the real ones.  E.g.

```diff
 fn start(
     &mut self,
     addr: u8,
     dir: $crate::i2c::Direction,
 ) -> Result<(), $crate::i2c::Error> {
     // Write start condition
-    self.p.$twcr.write(|w| w
+    self.p.twcr.write(|w| w
-        .$twen().set_bit()
+        .twen().set_bit()
-        .$twint().set_bit()
+        .twint().set_bit()
-        .$twstart().set_bit()
+        .twsta().set_bit()
     );
     self.wait();
```

avr-hal-generic/src/i2c.rs
Rahix: > a) write TWINT TWEN high

this clears the TWINT bit, as I described in the other comments

>  b) wait for TWINT to clear

actually, the line

https://github.com/Rahix/avr-hal/blob/9f9d6972e4ac7c149e2644bbc52c47cca81838f4/avr-hal-generic/src/i2c.rs#L357

waits for it to become set again ("while bit is cleared, loop").

avr-hal-generic/src/i2c.rs
ToddG: done.

avr-hal-generic/src/i2c.rs
ToddG: +1 on the `// WIP : ...` I typically add either `// TODO ...` or `// BUGBUG ...` but I'll adopt `WIP` for temp in dev comments going forward.

ToddG:
Ok, I'm getting close. I've encapsulated the data sheet status codes in a series of state transitions.

I need some feedback regarding the client experience/api? You mentioned signaling/polling an atomic variable at one point.

Rahix:
I'm going to take a closer look later today!

One thing we need to take care of is that this PR now contains many changes that are so to speak "orthogonal" to each other (e.g. "removing the $macro-tokens", "renaming `$I2c` to `$I2c Master`", and "adding the I2C slave implementation").  Ideally, each one should go into a single separate commit so that is becomes easy to trace what is changed each time.

I know that splitting out a changeset into such cleanly separated commits is quite cumbersome so I can offer to take care of this by pulling out all the "undisputed" changes (i.e. everything except the slave implementation itself, I think) into commits that I'll merge manually prematurely (of course attributing them to you!).  Once that is done, you can rebase this PR ontop of them and it will only contain the I2c slave driver changeset which is then much easier to look at in more detail.

Would that be okay for you?

ToddG:
@Rahix Somehow I missed your last message. My apologies for taking 6 days to respond. Yes, I was getting concerned about orthogonal changes drifting into this PR, too. Sure thing, splitting them out and rebasing sounds great, thank you for taking care of that.

ToddG:
@Rahix Ok, so I've added some triggers and helper methods based on your suggestions [here](https://github.com/Rahix/avr-hal/issues/75).

Couple of thoughts...

* A fully fleshed example using the `I2C Master API` would really help me understand the api. The [i2cdetect](https://github.com/Rahix/avr-hal/blob/master/boards/arduino-uno/examples/uno-i2cdetect.rs) just doesn't give my brain enough to work with. 
* Once I understand the pattern for master, I think this will unblock me. I'll wrap up by writing an example using the `I2C Slave API` as I finish writing that api.
* My final goal is to write a `master` <-> `slave` example of two arduinos talking to each other over the I2C bus

Rahix:
> Sure thing, splitting them out and rebasing sounds great, thank you for taking care of that.

Okay, will do as soon as I have time!

> A fully fleshed example using the I2C Master API would really help me understand the api. The i2cdetect just doesn't give my brain enough to work with.

Hmm, in the end, any embedded-hal I2C code will look the same.  So just take any existing I2C device driver (from [awesome embedded rust](https://github.com/rust-embedded/awesome-embedded-rust#driver-crates)) to see how the bus would be used.

> My final goal is to write a master <-> slave example of two arduinos talking to each other over the I2C bus

That makes a lot of sense!  Let's include the code for those two examples here as well!

Rahix:
> Ok, so I've added some triggers and helper methods based on your suggestions here.

Hmm, I would say these changes are going in the wrong direction.  A HAL driver should not contain any `static`s as this will mess with composability.  Such things should be done solely on application level.   I'm also not quite sure what you added them for?  If it's interrupts: they are irrelevant here, they are also something to be managed on application level (note how none of the other HAL drivers have any interrupt-specific code).

ToddG:
Ok, I'll back that static stuff out and examine the I2C driver in awesome embedded rust and see how that's implemented and used and then go from there.

ToddG:
Ok, so this example doesn't compile yet, but this is my take on the API for the `i2c slave`:

```rust
#![no_std]
#![no_main]

use arduino_uno::prelude::*;
use panic_halt as _;

#[arduino_uno::entry]
fn main() -> ! {
    let dp = arduino_uno::Peripherals::take().unwrap();

    let mut pins = arduino_uno::Pins::new(dp.PORTB, dp.PORTC, dp.PORTD);
    let mut serial = arduino_uno::Serial::new(
        dp.USART0,
        pins.d0,
        57600,
    );
    let mut i2c = arduino_uno::I2c::new(
        dp.TWI,
        pins.a4.into_pull_up_input(&mut pins.ddr),
        pins.a5.into_pull_up_input(&mut pins.ddr),
        50000,
    );
    let mut slave = arduino_uno::I2cSlave::new(
        i2c,
        pins.a1.into_pull_up_input(&mut pins.ddr),
        pins.a2.into_pull_up_input(&mut pins.ddr),
        b01111111,
        false,
    );

    // statemachine
    let mut state = slave.start()?;

    // data to write
    let mut data = 0;

    loop {
        ufmt::uwriteln!(&mut serial, "I2C slave starting...\r").void_unwrap();
        state = match(state){
            Ok(arduino_uno::I2c::I2cSlaveState::Uninitialized)
                =>{ state.init() }
            Ok(arduino_uno::I2c::I2cSlaveState::Initialized)
                =>{ state.wait() }
            Ok(arduino_uno::I2c::I2cSlaveState::AddressMatched)
                =>{
                    ufmt::uwriteln!(&mut serial, "I2C slave address: {}\r", state.address()).void_unwrap();
                    state.next()
            }
            Ok(arduino_uno::I2c::I2cSlaveState::RxReady)
                =>{
                    ufmt::uwriteln!(&mut serial, "I2C slave received: {}\r", state.read()).void_unwrap();
                    state.ack().next()
            }
            Ok(arduino_uno::I2c::I2cSlaveState::TxReady)
                =>{
                    ufmt::uwriteln!(&mut serial, "I2C slave writing: {}\r", data).void_unwrap();
                    state.write(data);
                    state.next()
            }
            Err(arduino_uno::I2c::Error)
                => {
                    ufmt::uwriteln!(&mut serial, "I2C slave error\r").void_unwrap();
                    state.init()
            }
        }
    }
}

```

Rahix:
@ToddG, can you quickly review #102 so I can move forward with that?

ToddG:
@Rahix #102 approved. I'm not sure what's going on with github notifications. I'm online, logged into github, all day every day for work...yet somehow I'm not getting notifications on PR's and such. I'll look at my settings closely...my apologies for the delays.

Rahix:
No problem, thanks for the reply! I've now merged #102. Please rebase this branch ontop of latest master now and make sure only the actual I2C slave driver changes remain here.

PR #95, opened 2020-10-26
Remove type defines

jkristell:
Removes the type defines from the macro. This is in preparation for merging support for the atmega328pb which has two Spi peripherals and the type statements causes conflicts. Also fixes changes `Spi` to `$Spi` in the`new` method.

Compile tested


Rahix (approved):
LGTM!

Issue #94, opened 2020-10-26
Consider merging (some) HAL crates


Rahix:
From my perspective as the maintainer of `avr-hal`, the HAL crates are becoming more and more of a burden to maintain because any global change needs to be reflected in each one.  There is a huge amount of code-duplication across the HAL crates which is not helping with this.

I think it makes sense to look into merging at least some of the HAL crates, for chips where the differences are small enough that `#[cfg()]`s can be used to differentiate between them.

The big question is where to draw the lines, what MCUs should go into which HAL crate:

- The nuclear solution would be to put all devices of `ATmega`/`ATtiny` into a crate.  This would leave us with the minimum number of crates but a mess of `#[cfg()]`s to manage them.
- On the other end of the spectrum, we could just go by the families as described by Microchip in their [Overview (PDF)](http://ww1.microchip.com/downloads/en/DeviceDoc/30010135E.pdf).  For the current set of supported MCUs this would not give us much advantage over the current situation, though.

IMO we need to larger grouping than the "Microchip Families" but just forcing all `ATmega`/`ATtiny` devices into a single crate also looks scary.  The latter might still be manageable now, but looking into the future, with more and more devices being supported, I can see it becoming a large mess.

Not yet sure what to do ...

jaxter184:
[This wikipedia article](https://en.wikipedia.org/wiki/ATtiny_microcontroller_comparison_chart) groups the ATtinys into 4 groups (not including the newer generation, which is pretty clear about its grouping). I'm not sure what exactly the process is for that, but it seems to be mostly by architecture? I'm not sure how useful it would be to use that in this case, but if the goal is just to reduce the number of crates, it seems like a reasonable metric. Otherwise, I'm a fan of splitting by `ATmega`/`ATtiny`

MarkuBu:
In theory this HAL should cover all available AVR devices. Not sure if it is possible to find a lowest common denominator for that. Maybe it is a better idea to look at the available peripherals and combine them to a device

Rahix:
Hmm, I also found this [Stackoverflow Answer](https://electronics.stackexchange.com/a/390541) which might give a good base:

> "4" The last digit marks the series of chips. Within a series they are often pin-compatible and share a similar set of features. Some prominent series are:
>
> - 'no digit' - these are the first generation chips with 8 to 128 kiB Flash
> - '8' - a series from 4 to 32 kiB Flash, all in the same housing. More or less an improved version of the original chips
[...]

PR #96, opened 2020-10-26
Add atmega328pb as a extension to atmega328p

jkristell:
I think it makes sense to make atmega328pb a "feature" of atmega328p. And it looks okayish this way. We could tweak the the device a bit so that the first i2c and spi device doesn't have a 0-suffix so that the second device would just be an addition, like for the Usart.

Remarks and questions
- [x] This depends on the Spi macro fix in #95 
- [x] And I noticed I have a couple of todos left in the timer setup I need to take a look at.
- [x] Is there a clever way to select just one of the devices from avr-device?

### DIFF
--- a/chips/atmega328p-hal/Cargo.toml
+++ b/chips/atmega328p-hal/Cargo.toml
@@ -8,10 +8,12 @@ edition = "2018"
 rt = ["avr-device/rt"]
 # Package exposes the ADC6 and ADC7 pins (only 32TQFP, 32MLF, 32UFBGA)
 adc-pins = []
+device-selected = []
+atmega328p = ["avr-device/atmega328p", "device-selected"]
+atmega328pb = ["avr-device/atmega328pb", "device-selected"]
 
 [dependencies]
 avr-hal-generic = { path = "../../avr-hal-generic/" }
 
 [dependencies.avr-device]
 version = "0.2.2"
-features = ["atmega328p"]
--- a/chips/atmega328p-hal/src/lib.rs
+++ b/chips/atmega328p-hal/src/lib.rs
@@ -1,7 +1,23 @@
 #![no_std]
 
+#[cfg(not(feature = "device-selected"))]
+compile_error!(
+    "This crate requires you to specify your target chip as a feature.
+
+    Please select one of the following
+
+    *   atmega328p
+    *   atmega328pb
+	"
+);
+
+
 /// Reexport of `atmega328p` from `avr-device`
+#[cfg(feature = "atmega328p")]
 pub use avr_device::atmega328p as pac;
+/// Reexport of `atmega328pb` from `avr-device`
+#[cfg(feature = "atmega328pb")]
+pub use avr_device::atmega328pb as pac;
 
 /// See [`avr_device::entry`](https://docs.rs/avr-device/latest/avr_device/attr.entry.html).
 #[cfg(feature = "rt")]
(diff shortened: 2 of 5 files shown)


chips/atmega328p-hal/Cargo.toml
```diff
@@ -8,10 +8,12 @@ edition = "2018"
 rt = ["avr-device/rt"]
 # Package exposes the ADC6 and ADC7 pins (only 32TQFP, 32MLF, 32UFBGA)
 adc-pins = []
+# Compile for the atmega328pb version of the chip
+pb = []
```
Rahix, line 12: I quite like the pattern the STM32 crates are using here (as I am also looking into merging even more of the HAL crates, see #94).  That would mean:

```toml
device-selected = []
atmega328p = ["avr-device/atmega328p", "device-selected"]
atmega328pb = ["avr-device/atmega328pb", "device-selected"]
```

```rust
#[cfg(not(feature = "device-selected"))]
compile_error!(
    "This crate requires you to specify your target chip as a feature.

    Please select one of the following

    *   atmega328p
    *   atmega328pb
	"
);

#[cfg(feature = "atmega328p")]
pub use avr_device::atmega328p as pac;
#[cfg(feature = "atmega328pb")]
pub use avr_device::atmega328pb as pac;

#[cfg(feature = "device-selected")]
pub mod ... {}
#[cfg(feature = "device-selected")]
pub mod ... {}
```

jkristell: That's a good way to do it. Thanks for the suggestion.

jkristell:
As always, Only compile tested :-)

chips/atmega328p-hal/src/lib.rs
```diff
@@ -93,9 +169,62 @@ pub mod spi {
...
+                sclk: portc::PC1,
+                mosi: porte::PE3,
+                miso: portc::PC0,
+            }
+        }
+    }
+}
+
 /// Serial interface using USART
 pub mod usart {
```
Rahix, line 225: Please feature-gate all modules that aren't already feature-gated by your changes behind `device-selected`.  Otherwise, when forgetting to select a device, you'd get a ton of compiler errors which just hide the one interesting one generated by the `compile_error!()` above.

```suggestion
#[cfg(feature = "device-selected")]
pub mod usart {
```

Rahix: Ah, I maybe didn't make it clear what I meant:  The other modules need the same
feature-guard:

```rust
#[cfg(feature = "device-selected")]
pub mod port;
#[cfg(feature = "device-selected")]
pub mod adc;
#[cfg(feature = "device-selected")]
pub mod pwm;
#[cfg(feature = "device-selected")]
pub mod wdt;

#[cfg(feature = "device-selected")]
pub mod prelude {
    pub use avr_hal_generic::prelude::*;
    pub use crate::port::PortExt as _;
}
```

jkristell: Your right. Added the gates and added missing pac documentation

Rahix (approved):
Looks good now!  Thanks for working on this!

jkristell:
Did a rebase to clean up the history. I think this could be merged now. Thanks for all your patience with this unexpectedly complicated add of a device :)

Rahix:
> Did a rebase to clean up the history.

Much appreciated!  I've given up asking people to do it by now ...

jkristell:
Thanks for the merge. Github should add the squash feature that gitlab has, that's really handy.

Also want you to know that this is actually tested on hardware now, and the things I tested so far works great. You have done a fantastic job with all of the avr-crates. Thanks!

PR #98, opened 2020-11-05
Reexport avr-hal-generic

jkristell:
Reexport avr_hal_generic so that its functionality can be used by hals outside of this crate.

I'm creating a custom HAL for a custom board at https://github.com/jkristell/rustyremote and want to be able to do so outside of this crate. One showstopper is the need for avr-hal-generic as it's not possible to use it as a git dependency. I guess the plan is to release it to crates.io as well. But in the mean time this seems to work.

### DIFF
--- a/chips/atmega328p-hal/src/lib.rs
+++ b/chips/atmega328p-hal/src/lib.rs
@@ -25,6 +25,7 @@ pub use avr_device::entry;
 
 pub use avr_hal_generic::clock;
 pub use avr_hal_generic::delay;
+pub use avr_hal_generic;
 
 #[cfg(feature = "device-selected")]
 pub mod port;


Rahix:
Does this not work?

```toml
[dependencies.atmega328p-hal]
git = "https://github.com/Rahix/avr-hal.git"
rev = "9f9d6972e4ac7c149e2644bbc52c47cca81838f4"
features = ["atmega328pb"]

[dependencies.avr-hal-generic]
git = "https://github.com/Rahix/avr-hal.git"
rev = "9f9d6972e4ac7c149e2644bbc52c47cca81838f4"
```

That said, what exactly do you need from `avr-hal-generic`?  Ideally, all items that are useful for downstream users should be reexported somewhere in the HAL crates.

jkristell:
That works! I didn't know you could do that.

I need avr-hal-generic for the impl_board_pins macro to name the routed pins: https://github.com/jkristell/rustyremote/blob/master/src/pins.rs

So perhaps that should be exported instead?

Rahix:
ah, that makes sense.  Hmm, I'm not sure to be honest, I'm leaning towards counting this as an exception as it is something that is truly meant to be generic ... Though I don't really have good arguments for this position so I'd be open to changing it if it shows to be better as re-exported.

jkristell:
Perhaps some kind of board support-support package then :). This is sort related to #94, and the issue with maintainer burden. If boards support packages can be moved out of this crate (when the hal is ready for it) others could step in more easily and maintain them. The KontroLIR I'm working is so niche that I don't think it makes sense to have it in this repo.

Closing this as I can use avr-hal-generic as a git dependency for now.

Issue #99, opened 2020-11-06
Some uncommon baudrates might be unstable


Rahix:
Following PR #88, some (uncommon!) baudrates seem to be unstable:  Namely 14400, 28800, and 76800 seem to be problematic.  Further analysis of the generated waveform is needed to determine what is the problem.

[Relevant comment from the PR](https://github.com/Rahix/avr-hal/pull/88#discussion_r511619365).

Cc: @explicite

explicite:
@Rahix I want to create test for this. I'm not sure if it's possible. Let me explain my idea.

I want to connect **mega2560** pins 18 and 19 to **Uno** on with problematic baudrates**. Uno** will only echo char from from **mega2560**.
**mega2560** will be connected via usb to ttl to my **PC** on standard baudrate like 56700.  

I will type some char and send to **mega2560**. **mega2560** will route this with problematic baudrate to **Uno**. **Uno** will echo to **mega2560** and to PC. If char will be same then we can prove that USB to TTL is not able to support this baudrates.

Please let me know if this is good idea of we will need to debug this with oscilloscope.

explicite:
@Rahix looks like for mega2560 other than USART 0 required to lib update for board. There is type alias for
https://github.com/Rahix/avr-hal/blob/master/boards/arduino-mega2560/src/lib.rs#L25

Maybe we can try to  create USART file with all definitions in board?

Rahix:
Right, we can add a module to the `arduino-mega2560` crate with aliases for all the serial stuff:

```rust
pub mod usart {
    pub use avr_hal_generic::serial::*;
    pub type Usart0<IMODE> = ...;
    pub type Usart1<IMODE> = ...;
}
```

Rahix:
@explicite, just a quick heads up: I've pushed commits 493546530eb87d47ae1eee507e9309be590688d4 and 6b05d48b4e43a15b3610a0e852b261e61ab0cade which might affect you a bit.

explicite:
@Rahix, thanks, looks like I was able to apply this changes.
`14400` works fine with my test [mega2560](https://github.com/explicite/avr-hal/blob/0184ce3a15ca114bc78d23f8c8ab28b64f7096a8/boards/arduino-mega2560/examples/mega2560-serial.rs#L19-L41)
[...]

PR #100, opened 2020-11-09
arduino-mega2560 usart's definition for board

explicite:
USART's definition for arduino-mega2560 to help in #99

### DIFF
--- a/boards/arduino-mega2560/src/lib.rs
+++ b/boards/arduino-mega2560/src/lib.rs
@@ -22,7 +22,17 @@ pub use crate::hal::spi;
 pub use crate::hal::adc;
 
 pub type Delay = crate::hal::delay::Delay<hal::clock::MHz16>;
-pub type Serial<IMODE> = crate::hal::usart::Usart0<hal::clock::MHz16, IMODE>;
+pub type Serial<IMODE> = crate::usart::Usart0<IMODE>;
+
+pub mod usart {
+    pub use avr_hal_generic::serial::*;
+
+    pub type Usart0<IMODE> = crate::hal::usart::Usart0<crate::hal::clock::MHz16, IMODE>;
+    pub type Usart1<IMODE> = crate::hal::usart::Usart1<crate::hal::clock::MHz16, IMODE>;
+    pub type Usart2<IMODE> = crate::hal::usart::Usart2<crate::hal::clock::MHz16, IMODE>;
+    pub type Usart3<IMODE> = crate::hal::usart::Usart3<crate::hal::clock::MHz16, IMODE>;
+}
+
 pub type I2c<M> = crate::hal::i2c::I2c<hal::clock::MHz16, M>;
 
 /// Support for PWM pins


boards/arduino-mega2560/examples/mega2560-adc.rs
```diff
@@ -23,7 +23,7 @@ fn main() -> ! {
     );
     let mut delay = arduino_mega2560::Delay::new();
 
-    let mut serial = arduino_mega2560::Serial::new(
+    let mut serial = arduino_mega2560::usart::Usart0::new(
```
explicite, line 26: @Rahix can we simplify this? Maybe we can redefine new in board to skip first attribute?

boards/arduino-mega2560/src/lib.rs
```diff
@@ -22,7 +22,17 @@ pub use crate::hal::spi;
 pub use crate::hal::adc;
 
 pub type Delay = crate::hal::delay::Delay<hal::clock::MHz16>;
-pub type Serial<IMODE> = crate::hal::usart::Usart0<hal::clock::MHz16, IMODE>;
+
+pub mod usart {
+    pub use atmega2560_hal as hal;
```
explicite, line 27: Not sure why, but I need to reimport.

Rahix: Should not be needed when you change the lines below like this:

```diff
-     pub type Usart0<IMODE> = crate::hal::usart::Usart0<hal::clock::MHz16, IMODE>;
+     pub type Usart0<IMODE> = crate::hal::usart::Usart0<crate::hal::clock::MHz16, IMODE>;
```

boards/arduino-mega2560/src/lib.rs
```diff
@@ -22,7 +22,17 @@ pub use crate::hal::spi;
...
+    pub use avr_hal_generic::serial::*;
+
+    pub type Usart0<IMODE> = crate::hal::usart::Usart0<hal::clock::MHz16, IMODE>;
+    pub type Usart1<IMODE> = crate::hal::usart::Usart1<hal::clock::MHz16, IMODE>;
+    pub type Usart2<IMODE> = crate::hal::usart::Usart1<hal::clock::MHz16, IMODE>;
+    pub type Usart3<IMODE> = crate::hal::usart::Usart1<hal::clock::MHz16, IMODE>;
```
Rahix, line 33: You forgot to update these here:

```suggestion
    pub type Usart2<IMODE> = crate::hal::usart::Usart2<hal::clock::MHz16, IMODE>;
    pub type Usart3<IMODE> = crate::hal::usart::Usart3<hal::clock::MHz16, IMODE>;
```

boards/arduino-mega2560/src/lib.rs
```diff
@@ -22,7 +22,17 @@ pub use crate::hal::spi;
 pub use crate::hal::adc;
 
 pub type Delay = crate::hal::delay::Delay<hal::clock::MHz16>;
-pub type Serial<IMODE> = crate::hal::usart::Usart0<hal::clock::MHz16, IMODE>;
```
Rahix, line 25: I'd really keep this alias and just change it to

```rust
pub type Serial<IMODE> = crate::usart::Usart0<IMODE>;
```

then you also don't need to update all the examples.

The `Serial` is supposed to be the one that is directly connected to the USB-serial-converter, i.e. the one you'll interact with from your PC.  This one is made more accessible than the others because its the one that will be used the most.

boards/arduino-mega2560/examples/mega2560-adc.rs
Rahix: I don't think so.  As I said in the other comment, I'd keep the `Serial` alias so the examples don't need any changing.

Rahix (changes requested):

boards/arduino-mega2560/src/lib.rs
explicite: Done

Rahix (approved):
Looks good now, thanks!

PR #101, opened 2020-11-11
ATmega4809 support

explicite:
Spec for ATmega4809 chip

@Rahix can you please define how we should split this work? I think we can start with spec.

### DIFF
--- a/avr-specs/avr-atmega4809.json
+++ b/avr-specs/avr-atmega4809.json
@@ -1,0 +1,32 @@
+{
+  "arch": "avr",
+  "atomic-cas": false,
+  "cpu": "atmega4809",
+  "data-layout": "e-P1-p:16:8-i8:8-i16:8-i32:8-i64:8-f32:8-f64:8-n8-a:8",
+  "eh-frame-header": false,
+  "env": "",
+  "exe-suffix": ".elf",
+  "executables": true,
+  "late-link-args": {
+    "gcc": [
+      "-lgcc"
+    ]
+  },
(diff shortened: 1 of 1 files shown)


Rahix:
My suggestion would be adding the HAL crate in one PR and then adding the board in a follow-up PR.  So, let's start with the `atmega4809-hal` crate in this PR.  I think you will also need to add support for the MCU in [`avr-device`](https://github.com/Rahix/avr-device) first (See ["Adding a new Chip"](https://github.com/Rahix/avr-device#adding-a-new-chip) for details on that).

explicite:
I'm getting error on atdf to svd conversion:
```
jpaw@node ~/s/avr-device> atdf2svd vendor/atmega4809.atdf svd/atmega4809.svd                                                             
Error: access-mode '"W"' is unsupported in element
    <bitfield mask="0xff" values="EVSYS_STROBE0" rw="W" name="STROBE0" caption="Software event on channels" ...>
```
Probably this line https://github.com/avr-rust/avr-mcu/blob/master/packs/atmega/ATmega4809.atdf#L1097

Not sure how to fix that.

Rahix:
Interesting. That'll need a fix in `atdf2svd` right here: https://github.com/Rahix/atdf2svd/blob/d342cc7f5b15413b547db5a715ed85fa79a17435/src/atdf/field.rs#L43-L50.  Not sure why it is missing in the first place if `R` and `WR` are here ... All the remaining parts for write-only access are implemented it seems, so this is the only change necessary.

PR #102, opened 2020-11-15
Changes from #93, cleaned and rebased on master

Rahix:
Hey @ToddG,

I've isolated the separate changes from your PR (#93) now.  I have taken a bit of liberty in doing so and did some additional changes to make everything work well (for the most part minor fixes and deprecation aliases to ease updating).  Please take a look at the individual commits whether they reflect your original intent and whether it is okay for you to have your name tagged on them in this form :)

If everything is alright for you, I'd merge these as they are.  Then, the only remaining change in your PR should be the actual I2C slave driver.  Please rebase your work ontop of master after this PR here is merged such that only the slave-driver changes remain in there.

### DIFF
--- a/avr-hal-generic/src/i2c.rs
+++ b/avr-hal-generic/src/i2c.rs
@@ -186,16 +186,16 @@ macro_rules! impl_twi_i2c {
                 scl: $sclmod:ident::$SCL:ident,
             },
         }
-    ) => {
+    ) => {$crate::paste::paste! {
         $(#[$i2c_attr])*
-        pub struct $I2c<CLOCK: $crate::clock::Clock, M> {
+        pub struct [<$I2c Master>]<CLOCK: $crate::clock::Clock, M> {
             p: $I2C,
             _clock: ::core::marker::PhantomData<CLOCK>,
             sda: $sdamod::$SDA<M>,
             scl: $sclmod::$SCL<M>,
         }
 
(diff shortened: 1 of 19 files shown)


Rahix:
@ToddG, can you quickly give me an _ok_/_not ok_ for this PR?  I'd like to get those changes in so we can then properly look at your I2C slave driver. :)

avr-hal-generic/src/i2c.rs
```diff
@@ -254,26 +236,32 @@ macro_rules! impl_twi_i2c {
...
         }
 
-        impl<CLOCK, M> $I2c<CLOCK, M>
+        impl<CLOCK, M> [<$I2c Master>]<CLOCK, M>
         where
             CLOCK: $crate::clock::Clock,
         {
+            fn set_speed(&mut self, speed: u32){
+                // Calculate TWBR register value
+                let twbr = ((CLOCK::FREQ / speed) - 16) / 2;
```
explicite, line 258: Is there any idea to extend and use https://github.com/Rahix/avr-hal/blob/master/avr-hal-generic/src/usart.rs#L7

Rahix: Hmm, serial baudrate and I2C clock speed are not really related concepts so re-using your code for USART does not quite make sense here.

ToddG (commented):
Looks good to me.

ToddG (approved):

Rahix:
Thanks @ToddG!

PR #103, opened 2020-11-17
treewide: spi: Enforce proper CS pin usage

### LINKED ISSUE #27: SPI: CS must be configured before SPI peripheral
I'm trying to use the SPI library but it's not working as expected.  The `send` method works fine and returns without incident, but the `read` method blocks forever.

```
    let byte = block!(spi.send(5u8).and_then(|_| spi.read()));
    if let Ok(byte) = byte {
        ufmt::uwriteln!(&mut serial, "wrote read {}\r", byte).unwrap();
    } else {
        ufmt::uwriteln!(&mut serial, "not wrote read\r").unwrap();
    }
```
It never makes it past the `block` macro, it seems that it's returning WouldBlock forever.  All `read` does is try to flush, code here:

https://github.com/Rahix/avr-hal/blob/master/avr-hal-generic/src/spi.rs#L125

The is_write_in_progress is obviously true as expected, since a send was just executed.  But it seems that the write complete status register is not being set.

https://github.com/Rahix/avr-hal/blob/master/avr-hal-generic/src/spi.rs#L127

Documentation for the SPI interrupt flag is in section 18.5.2 of this document:

http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-7810-Automotive-Microcontrollers-ATmega328P_Datasheet.pdf

I'm trying to debug this but any suggestions are welcome.

jonahd-g:
/resolves #27 

Makes the proposed change to the SPI macro and the usage by ATmega328p.  Waiting for review and approval before making an attempt to update all other usages.

### DIFF
--- a/avr-hal-generic/src/spi.rs
+++ b/avr-hal-generic/src/spi.rs
@@ -83,18 +83,31 @@ macro_rules! impl_spi {
         /// Wrapper for the CS pin
         ///
         /// Used to contain the chip-select pin during operation to prevent its mode from being
-        /// changed from Output.  Implements
-        /// [`OutputPin`](https://docs.rs/embedded-hal/latest/embedded_hal/digital/v2/trait.OutputPin.html)
-        /// so that it can still be controlled.
+        /// changed from Output. This is necessary because the SPI state machine would otherwise
+        /// reset itself to SPI slave mode immediately. This wrapper can be used just like an
+        /// output pin, because it implements all the same traits from embedded-hal.
         pub struct ChipSelectPin($csmod::$CS<$crate::port::mode::Output>);
-
         impl $crate::hal::digital::v2::OutputPin for ChipSelectPin {
-            type Error = <$csmod::$CS<$crate::port::mode::Output> as $crate::hal::digital::v2::OutputPin>::Error;
+            type Error = $crate::void::Void;
             fn set_low(&mut self) -> Result<(), Self::Error> {
                 self.0.set_low()
             }
             fn set_high(&mut self) -> Result<(), Self::Error> {
                 self.0.set_high()
+            }
+        }
+        impl $crate::hal::digital::v2::StatefulOutputPin for ChipSelectPin {
+            fn is_set_low(&self) -> Result<bool, Self::Error> {
+                self.0.is_set_low()
+            }
+            fn is_set_high(&self) -> Result<bool, Self::Error> {
+                self.0.is_set_high()
+            }
+        }
+        impl $crate::hal::digital::v2::ToggleableOutputPin for ChipSelectPin {
+            type Error = $crate::void::Void;
+            fn toggle(&mut self) -> Result<(), Self::Error> {
+                self.0.toggle()
             }
         }
 
--- a/avr-hal-generic/src/spi.rs
+++ b/avr-hal-generic/src/spi.rs
@@ -75,9 +75,29 @@ macro_rules! impl_spi {
                 sclk: $sclkmod:ident::$SCLK:ident,
                 mosi: $mosimod:ident::$MOSI:ident,
                 miso: $misomod:ident::$MISO:ident,
+                cs: $csmod:ident::$CS:ident,
             }
         }
     ) => {
+
+        /// Wrapper for the CS pin
+        ///
+        /// Used to contain the chip-select pin during operation to prevent its mode from being
+        /// changed from Output.  Implements
+        /// [`OutputPin`](https://docs.rs/embedded-hal/latest/embedded_hal/digital/v2/trait.OutputPin.html)
+        /// so that it can still be controlled.
+        pub struct ChipSelectPin($csmod::$CS<$crate::port::mode::Output>);
+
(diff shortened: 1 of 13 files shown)


avr-hal-generic/src/spi.rs
```diff
@@ -75,9 +75,29 @@ macro_rules! impl_spi {
...
+
+        /// Wrapper for the CS pin
+        ///
+        /// Used to contain the chip-select pin during operation to prevent its mode from being
+        /// changed from Output.  Implements
+        /// [`OutputPin`](https://docs.rs/embedded-hal/latest/embedded_hal/digital/v2/trait.OutputPin.html)
+        /// so that it can still be controlled.
+        pub struct ChipSelectPin($csmod::$CS<$crate::port::mode::Output>);
+
+        impl $crate::hal::digital::v2::OutputPin for ChipSelectPin {
```
Rahix, line 91: Should also implement `StatefulOutputPin` and `ToggleableOutputPin` from _embedded-hal_ as well.

avr-hal-generic/src/spi.rs
```diff
@@ -75,9 +75,29 @@ macro_rules! impl_spi {
...
+            type Error = <$csmod::$CS<$crate::port::mode::Output> as $crate::hal::digital::v2::OutputPin>::Error;
```
Rahix, line 92: Rust syntax at its finest :D  I would just use the concrete type though, should be `$crate::void::Void` if I'm not mistaken.

avr-hal-generic/src/spi.rs
```diff
@@ -94,19 +114,21 @@ macro_rules! impl_spi {
...
+            /// back so that its output state can be changed as needed.
             pub fn new(
                 peripheral: $SPI,
                 sclk: $sclkmod::$SCLK<$crate::port::mode::Output>,
                 mosi: $mosimod::$MOSI<$crate::port::mode::Output>,
                 miso: $misomod::$MISO<$crate::port::mode::Input<$crate::port::mode::PullUp>>,
+                cs: $csmod::$CS<$crate::port::mode::Output>,
                 settings: Settings
-            ) -> Self {
+            ) -> (Self, ChipSelectPin) {
```
Rahix, line 131: This means we'll also need to update all SPI examples to look like

```rust
let (mut spi, _) = Spi::new();
```

and I'll have to document this change as a breakage for downstream users.

avr-hal-generic/src/spi.rs
```diff
@@ -75,9 +75,29 @@ macro_rules! impl_spi {
...
             }
         }
     ) => {
+
+        /// Wrapper for the CS pin
+        ///
+        /// Used to contain the chip-select pin during operation to prevent its mode from being
+        /// changed from Output.  Implements
+        /// [`OutputPin`](https://docs.rs/embedded-hal/latest/embedded_hal/digital/v2/trait.OutputPin.html)
+        /// so that it can still be controlled.
```
Rahix, line 88: I'd write something like:

_[...] prevent its mode from being changed from Output.  This is necessary because the SPI state machine would otherwise reset itself to SPI slave mode immediately._

_This wrapper can be used just like an output pin, because it implements all the same traits from embedded-hal._

Rahix (changes requested):
Nice, thank you very much!  The design is solid, I've left a few comments but apart from that you can go ahead and update all the other HAL crates and examples :)

avr-hal-generic/src/spi.rs
jonahd-g: All right.  I wasn't sure if those were important, but I'll add it in.

avr-hal-generic/src/spi.rs
jonahd-g: Oh sweet, that makes it much easier.

avr-hal-generic/src/spi.rs
jonahd-g: Yeah, I'll make sure the examples get updated too.  I'll include a description of the breaking change in the changelog

jonahd-g:
> The design is solid

Lol you don't get to say that when I mostly copied from your comment :wink:

Rahix:
> Lol you don't get to say that when I mostly copied from your comment :wink:

Haha, I did not at all remember what I wrote back then, please forgive me :D

boards/arduino-leonardo/src/lib.rs
```diff
@@ -93,14 +93,13 @@ pub fn delay_us(us: u16) {
 ///
 /// let mut pins = arduino_leonardo::Pins::new(dp.PORTB, dp.PORTC, dp.PORTD, dp.PORTE);
 ///
-/// pins.led_rx.into_output(&mut pins.ddr); // Chip Select must be set to output mode.
-///
 /// // Create SPI interface.
 /// let mut spi = arduino_leonardo::spi::Spi::new(
```
Rahix, line 97:
```suggestion
/// let (mut spi, mut cs) = arduino_leonardo::spi::Spi::new(
```

Rahix (approved):
LGTM!  Thanks a lot!

PR #104, opened 2020-11-17
Update pins.rs

realErtugrul:
corrected false naming in the documentation


Rahix (approved):
Ouch, thanks for noticing! :)

PR #105, opened 2020-11-27
Add attiny841 and attiny861

jaxter184:
Preliminary pull request adding attiny841 and attiny861 chips. I tested a blink example on both, and it works in both cases, but I haven't tested much else.

It might also be worth mentioning in the README or something that the attiny841 is not supported by the current version of avrdude, so you need to add a section to the .avrduderc

### DIFF
--- a/avr-specs/avr-attiny841.json
+++ b/avr-specs/avr-attiny841.json
@@ -1,7 +1,7 @@
 {
   "arch": "avr",
   "atomic-cas": false,
-  "cpu": "avr25",
+  "cpu": "attiny841",
   "data-layout": "e-P1-p:16:8-i8:8-i16:8-i32:8-i64:8-f32:8-f64:8-n8-a:8",
   "eh-frame-header": false,
   "env": "",
--- a/avr-specs/avr-attiny841.json
+++ b/avr-specs/avr-attiny841.json
@@ -1,0 +1,32 @@
+{
+  "arch": "avr",
+  "atomic-cas": false,
+  "cpu": "avr25",
+  "data-layout": "e-P1-p:16:8-i8:8-i16:8-i32:8-i64:8-f32:8-f64:8-n8-a:8",
+  "eh-frame-header": false,
+  "env": "",
+  "exe-suffix": ".elf",
+  "executables": true,
+  "late-link-args": {
+    "gcc": [
+      "-lgcc"
+    ]
+  },
+  "linker": "avr-gcc",
+  "linker-flavor": "gcc",
+  "linker-is-gnu": true,
+  "llvm-target": "avr-unknown-unknown",
+  "max-atomic-width": 8,
+  "no-default-libraries": false,
+  "os": "unknown",
+  "pre-link-args": {
+    "gcc": [
+      "-mmcu=attiny841",
+      "-Wl,--as-needed"
+    ]
+  },
+  "target-c-int-width": "16",
+  "target-endian": "little",
+  "target-pointer-width": "16",
+  "vendor": "unknown"
+}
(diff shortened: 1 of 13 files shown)


chips/attiny841-hal/.cargo/config.toml
```diff
@@ -0,0 +1,5 @@
+[build]
+target = "../../avr-specs/avr-attiny85.json"
```
Rahix, line 2: That should be `avr-attiny841.json` I guess?

chips/attiny861-hal/.cargo/config.toml
```diff
@@ -0,0 +1,5 @@
+[build]
+target = "../../avr-specs/avr-attiny85.json"
```
Rahix, line 2: Should be `avr-attiny861.json`?

avr-specs/avr-attiny841.json
```diff
@@ -0,0 +1,32 @@
...
+    "gcc": [
+      "-mmcu=attiny841",
+      "-Wl,--as-needed"
+    ]
+  },
+  "target-c-int-width": "16",
+  "target-endian": "little",
+  "target-pointer-width": "16",
+  "vendor": "unknown"
+}
```
Rahix, line 32: The target specs should be generated via [`sync-from-upstream.py`](https://github.com/Rahix/avr-hal/blob/master/avr-specs/sync-from-upstream.py), can you integrate your two MCUs into that script?

Rahix (commented):
Hey,

thanks for this PR!  How similar are these two MCUs?  Would it make sense to provide a single HAL crate for both of them?  I'm looking into keeping the number of HAL crates as low as possible, see #94 ...

jaxter184:
I've made the requested changes

On combining the crates, these two chips are unfortunately pretty different (different number of pins, different peripherals), so I don't think it would make sense to group them together.

avr-specs/sync-from-upstream.py
```diff
@@ -28,6 +28,12 @@
     "attiny88": {
         "cpu": "attiny88",
     },
+    "attiny841": {
+        "cpu": "attiny841",
```
Rahix, line 32: I think this needs to be `avr25` to recreate your original spec files:

```suggestion
        "cpu": "avr25",
```

avr-specs/sync-from-upstream.py
```diff
@@ -28,6 +28,12 @@
...
+    },
+    "attiny861": {
+        "cpu": "attiny861",
```
Rahix, line 35: Same here:

```suggestion
        "cpu": "avr25",
```

Rahix:
> On combining the crates, these two chips are unfortunately pretty different (different number of pins, different peripherals), so I don't think it would make sense to group them together.

Different pins shouldn't be a problem.  Can you list the peripheral differences?  The key question is whether the peripherals that exist in both devices differ from each other.  If not, I'd be very much in favor of a single HAL.  I know I am pushing pretty strong for the unification because, as outlined in #94, the current state of many crates with duplicated code is absolutely unmaintainable for me ...

jaxter184:
|| t841 | t861 |
|-|-|-|
|gpio|a|a,b|
|timer|1x 8bit, 2x 16bit|1x 8bit, 1x 16/8bit hybrid|
|adc|10bit|10bit|
|ac|2x|1x|
|comms|SPI,I2C (slave only), 2x USART|USI|

The biggest difference is the serial communication options. I'd say the t861 is closer to the t85 than the t841 in terms of peripherals.

jaxter184:
If I do fix this pull request, I'll probably do it by reimplementing it from scratch.

PR #107, opened 2020-12-01
remove register! macro from atmega328pb 

fneddy:
the registers! macro was removed in ecb501aac279b09cfa44059e37eccb88969538de but this two calls where not removed


Rahix (approved):
Ouch, you're totally correct!  This MCU is not currently tested in CI so that allowed it to slip through :(  I should look into changing that ...

Issue #109, opened 2020-12-02
macro impl_spi! creates name collision in atmega328pb


fneddy:
the macro `impl_spi!` creates a struct `ChipSelectPin`
https://github.com/Rahix/avr-hal/blob/3d255df0194ba2cdbbb086be61b9146a37df7963/avr-hal-generic/src/spi.rs#L89

this makes it impossible to declare two spi ports in the same module like `atmega328pb` does:
https://github.com/Rahix/avr-hal/blob/3d255df0194ba2cdbbb086be61b9146a37df7963/chips/atmega328p-hal/src/lib.rs#L162-L184

currently atmega328pb build is broken.

PR #110, opened 2020-12-02
Quickfix broken SPI config for atmega328pb

fneddy:
`avr_hal_generic::impl_spi!` generates a name collision (`ChipSelectPin` is declared multiple times.
this is a quickfix for #109

### DIFF
--- a/chips/atmega328p-hal/src/lib.rs
+++ b/chips/atmega328p-hal/src/lib.rs
@@ -139,7 +139,7 @@ pub mod spi {
     //!
     //! ```
     //! // create SPI interface
-    //! let (mut spi, mut cs) = Spi0::new(
+    //! let (mut spi, mut cs) = port0::Spi0::new(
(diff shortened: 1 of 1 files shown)


Rahix:
Hmm, not really a fan of this approach.  I'd rather fix the macro directly, to emit differently named CS pin structs.  Eg. change the invocation to look like

```rust
		avr_hal_generic::impl_spi! {
            pub struct Spi0 {
                peripheral: crate::pac::SPI0,
                pins: {
                    sclk: portb::PB5,
                    mosi: portb::PB3,
                    miso: portb::PB4,
                    cs: portb::PB2,
                }
            }

            pub struct ChipSelectPin0;
        }
```

and use that name in the macro.

fneddy:
Ok, I'll try to implement it... but my macro skills are not very good.

Rahix:
I'm happy to help if you need anything!

PR #111, opened 2020-12-03
generic: spi: Fix CS pin name collision in impl_spi!()

### LINKED ISSUE #109: macro impl_spi! creates name collision in atmega328pb

fneddy:
add a wrapper for the `impl_spi!` macro to allow setting a dedicated name for `pub struct ChipSelectPin`

### DIFF
--- a/avr-hal-generic/src/spi.rs
+++ b/avr-hal-generic/src/spi.rs
@@ -66,9 +66,38 @@ impl Default for Settings {
 
 
 /// Implement traits for a SPI interface
-#[doc(hidden)]
 #[macro_export]
-macro_rules! impl_spi_internal {
+macro_rules! impl_spi {
+    (
+        $(#[$spi_attr:meta])*
+        pub struct $Spi:ident {
+            peripheral: $SPI:ty,
+            pins: {
+                sclk: $sclkmod:ident::$SCLK:ident,
+                mosi: $mosimod:ident::$MOSI:ident,
+                miso: $misomod:ident::$MISO:ident,
+                cs: $csmod:ident::$CS:ident,
+            }
+        }
+    ) => {
+        /// First match was without a 'ChipSelectPin' 
+        /// we set it here to a default name and then 
+        /// recusiv expand to the real inplementation 
+        $crate::impl_spi! {
+            pub struct $Spi {
+                peripheral: $SPI,
+                pins: {
+                    sclk: $sclkmod::$SCLK,
+                    mosi: $mosimod::$MOSI,
+                    miso: $misomod::$MISO,
+                    cs: $csmod::$CS,
+                }
+            }
+            pub struct ChipSelectPin;
+        }
+    };
+    
+
     (
         $(#[$spi_attr:meta])*
         pub struct $Spi:ident {
@@ -81,8 +110,7 @@ macro_rules! impl_spi_internal {
             }
         }
         pub struct $ChipSelectPin:ident;
-    )
-    => {
+    ) => {
 
         /// Wrapper for the CS pin
         ///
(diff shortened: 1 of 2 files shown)


avr-hal-generic/src/spi.rs
```diff
@@ -132,7 +136,7 @@ macro_rules! impl_spi {
             ///
             /// The pins are not actually used directly, but they are moved into the struct in
             /// order to enforce that they are in the correct mode, and cannot be used by anyone
-            /// else while SPI is active.  CS is placed into a `ChipSelectPin` instance and given
+            /// else while SPI is active.  CS is placed into a `$ChipSelectPin` instance and given
```
Rahix, line 139: I don't think macros will expand doc-comment content.

avr-hal-generic/src/spi.rs
```diff
@@ -161,7 +165,7 @@ macro_rules! impl_spi {
...
```
Rahix, line 168: Same here.

avr-hal-generic/src/spi.rs
```diff
@@ -299,3 +303,63 @@ macro_rules! impl_spi {
         }
     };
 }
+
+
+/// Implement traits for a SPI interface
+#[macro_export]
+macro_rules! impl_spi {
```
Rahix, line 310: I think it should be possible to do this without introducing a second macro if `impl_spi!` just invokes itself:

```rust
#[macro_export]
macro_rules! impl_spi {
    (
        $(#[$spi_attr:meta])*
        pub struct $Spi:ident {
            peripheral: $SPI:ty,
            pins: {
                sclk: $sclkmod:ident::$SCLK:ident,
                mosi: $mosimod:ident::$MOSI:ident,
                miso: $misomod:ident::$MISO:ident,
                cs: $csmod:ident::$CS:ident,
            }
        }
    ) => {
        $crate::impl_spi! {
            pub struct $Spi {
                peripheral: $SPI,
                pins: {
                    sclk: $sclkmod::$SCLK,
                    mosi: $mosimod::$MOSI,
                    miso: $misomod::$MISO,
                    cs: $csmod::$CS,
                }
            }
            pub struct ChipSelectPin;
        }
    };
    (
        $(#[$spi_attr:meta])*
        pub struct $Spi:ident {
            peripheral: $SPI:ty,
            pins: {
                sclk: $sclkmod:ident::$SCLK:ident,
                mosi: $mosimod:ident::$MOSI:ident,
                miso: $misomod:ident::$MISO:ident,
                cs: $csmod:ident::$CS:ident,
            }
        }
        pub struct $ChipSelectPin:ident;
    )
    => {
        // actual implementation ...
    };
}
```

chips/atmega328p-hal/src/lib.rs
```diff
@@ -178,9 +179,10 @@ pub mod spi {
                 sclk: portc::PC1,
                 mosi: porte::PE3,
                 miso: portc::PC0,
-                cs: portc::PE2,
+                cs: porte::PE2,
```
Rahix, line 182: Nice catch, I wonder how this ever worked ...

Rahix (changes requested):
Thanks a lot, this approach looks like the way to go!  I've left a few comments below.

chips/atmega328p-hal/src/lib.rs
jkristell: I'm also surprised. Thanks for fixing this.

avr-hal-generic/src/spi.rs
fneddy: wow, this is clever. I'll try it.

Rahix (approved):
Looks good to me!  CI is passing so it seems to work fine.  I suppose you tested it for `ATmega328PB` as well?

fneddy:
it compiles and a blinktest is working, but i currently have no SPI setup here. So you may declare it as "compiling-ok"

Rahix:
Good enough for me, I don't see this change breaking anything runtime related ;)

Issue #42, opened 2020-08-09
Add ability to re-configure SPI.


davidlenfesty:
I'm trying to use SPI to read off an SD card, and part of the requirements of an SD card is that it needs to initialize at 100-400kHz, but then you can run it faster afterwards.

I'm using `embedded-sdmmc`, and it allows you to reconfigure the peripheral after initializing through a function. However as this HAL currently exists, there is no way to do that other than releasing the existing instance and creating a new one.

I have tested the following and it seems to work (haven't built the hardware to test yet):

``` rust
impl $Spi {
    [...]
    /// Reconfigure SPI peripheral without releasing pins.
    pub fn reconfigure(&mut self, settings: Settings) {
        self.settings = settings;
        self.setup();
    }
    [...]
}
```

Rahix:
cc @jonahbron, what do you think?

couchand:
Presumably you'd want to wait for any in-flight writes to complete, so perhaps the implementation would be:

```
pub fn reconfigure(&mut self, settings: Settings) -> nb::Result<(), Void> {
    self.flush()?;
    self.settings = settings;
    self.setup();
    Ok(())
}
```

PR #112, opened 2020-12-04
generic: spi: Add method to reconfigure the peripheral

explicite:
Draft for #42

@Rahix can you please review if I made change in correct place? I will try to find some test case to prove this.

### DIFF
--- a/avr-hal-generic/src/spi.rs
+++ b/avr-hal-generic/src/spi.rs
@@ -266,6 +266,14 @@ macro_rules! impl_spi {
                     SerialClockRate::OscfOver128 => w.spi2x().clear_bit(),
                 });
             }
+            // to reconfigure the peripheral after initializing
+            pub fn reconfigure(&mut self, settings: Settings) -> $crate::nb::Result<(), $crate::void::Void> {
+                // wait for any in-flight writes to complete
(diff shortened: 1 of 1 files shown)


Rahix (commented):
The code looks good to me, but you're right, we need a way to test this ...

explicite:
Test case with two Arduino: Mega2560(**Master**), Uno(**Slave**). Master connected via UART to PC (**57600**). 
1. `char` from PC go Mega
2. Mega fwd. `char` to Uno via SPI
3. Uno echo `char` to Mega
4. Mega update  reconfigure SPI (serial clock rate)
5. Mega echo `char` to Uno
6. Uno echo `char` to Mega
7. Mega fwd `char` to PC

If received char will be same then test pass. Will review signals via logic analyzer between Uno and Mega

@Rahix should I change data order and mode to?

Rahix (approved):
I think this is good enough and would have caught any obvious problems.  Thanks a lot!

explicite:
Looks like I'm not able to setup SPI to communicate between Arduino. I review with logic analyzer and I cannot get any output from Uno and Mega. @Rahix can you please review https://github.com/Rahix/avr-hal/pull/118

PR #118, opened 2020-12-12
test/topic 42

explicite:
Test for #42


Rahix:
Ah, I think I see the issue: Both devices are configured as SPI controllers and will try to drive the clock.  One of the devices (the Uno) needs to be configured as SPI peripheral which we don't currently support AFAIK. :/

So I guess you're better off just looking at the signals the Mega 2560 sends with a logic analyzer and verify correct operation of reconfiguration this way.

explicite:
Ok, this simplified flow tested and works fine (serial clock). Do we have any plans to support SPI as peripheral?

Rahix:
> Do we have any plans to support SPI as peripheral?

Not at the moment but we can open an issue for keeping track if you want :)

PR #126, opened 2021-01-31
Fix board runners to correctly detect dependencies

crclark96:
the `&>` redirection literal is undefined behavior in `sh`, leading to
incorrect dependency detection --
https://github.com/koalaman/shellcheck/wiki/SC2039\#redirect-both-stdout-and-stderr

before:

after:


Rahix (approved):
True, thanks for catching this!

Issue #117, opened 2020-12-12
Unified arduino-hal


Rahix:
I think it makes sense to move towards merging board crates similar to what we're doing with HAL crates in #94.  In specific, I want to create a single `arduino-hal` crate with features for the specific devices.  The goal should be to provide an API that allows very easily switching from one board to another.  This means we should abstract all board-specific details into a common layer, as far as that is reasonably possible.

Issue #114, opened 2020-12-09
Support IO-cell of newer AVR devices


Rahix:
Some newer generations of AVR MCUs use a different IO-cell:  There are 10 registers instead of 3 and one additional reg for each pin.  E.g.

- `DIR`, `DIRSET`, `DIRCLR`, `DIRTGL`
- `OUT`, `OUTSET`, `OUTCLR`, `OUTTGL`
- `IN`
- `INTFLAGS`
- `PIN0CTRL`, `PIN1CTRL`, `PIN2CTRL`, `PIN3CTRL`, `PIN4CTRL`, `PIN5CTRL`, `PIN6CTRL`, `PIN7CTRL`

We need a common abstraction that allows both the old and the new IO-cell to be used with the exact same API.  While integrating that, we can also rethink the entire IO HAL design.  My proposal is (in general) to move away from the all macro based approach and do as much directly in `avr-hal-generic` with only the PAC-specific parts being implemented by a macro.  For the IO HAL in particular, this could mean:

```rust
// in avr-hal-generic
pub struct Pin<PIN: PinOps, MODE: PinMode> {
    p: PIN,
    _mode: core::marker::PhantomData<MODE>,
}

pub trait PinOps {
    unsafe fn out_set(&mut self, value: bool);
    unsafe fn out_toggle(&mut self);
    unsafe fn set_direction(&mut self, output: bool);
    unsafe fn in(&self) -> bool;
}

// the macro then implements `PinOps` for concrete (zero-sized) types for each pin:
    pub struct PA0;

    impl PinOps for PA0 {
        #[inline(always)]
        unsafe fn out_set(&mut self, value: bool) {
            if value {
                (*<$PORTX>::ptr()).OUTSET.write(|w| 1 << $i);
            } else {
                (*<$PORTX>::ptr()).OUTCLR.write(|w| 1 << $i);
            }
        }

        // ...
    }

    // and similarly for the "old" IO-cell ...
```

Rahix:
Cc: @chris-ricketts

chris-ricketts:
> Cc: @chris-ricketts

@Rahix Thanks for linking those issues. I'll take those into consideration.

I'm still getting my head around the current ports API/macros and experimenting.

I think a single macro in a form similar to what @fneddy suggested would be possible for defining a chip's ports.
```rust
// Something like the following could be done, maybe
// if we use features to select pin impl macro
port!(PortA, [PA0, PA1, PA2, PA3]) 

[...]

PR #129, opened 2021-02-07
Start work on the architectural refactor

Rahix:
So, it's finally time:  I've gotten around to starting work on the big refactor that got more and more inevitable from issues like #114, #94, and #117.

Right now, the state is just barely enough to blink an LED on Arduino Uno.  I wanted to get this out as quickly as possible to initiate discussions on the design.  As this work will take a while to complete, we will go about it like this:  `master` stays with the current architecture for now until the new design reaches feature-parity.  Until then, we will use the `next` branch to get it there.  Thus, this PR will be merged into `next` instead of `master`.

#### TODO
- [x] Re-write documentation for the `port` module.
- [x] Test integration with the already refactored USART.
- [x] Add basic Arduino Leonardo support to see how that works out.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - next
       - wip
   pull_request:
 
@@ -16,11 +17,6 @@ jobs:
         board:
           - arduino-uno
           - arduino-leonardo
-
-          - arduino-mega2560
-          - bigavr6
-          - trinket
-          - sparkfun-pro-micro
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
@@ -35,4 +31,4 @@ jobs:
       - name: Install avr-gcc, binutils, and libc
         run: sudo apt-get install -y avr-libc binutils-avr gcc-avr
       - name: Compile board crate and examples
-        run: cd "boards/${{ matrix.board }}" && cargo build --examples
+        run: cd "examples/${{ matrix.board }}" && cargo build --bins
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,9 @@ members = [
     "chips/attiny85-hal",
     "chips/attiny88-hal",
 
+    # MCU HAL crates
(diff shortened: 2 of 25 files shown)


Rahix:
As a short breakdown for anyone interested:

- `mcu/atmega-hal/` is the new unified HAL for ATmega MCUs
- `arduino-hal/` is the unified board support crate for Arduino boards
- `examples/arduino-uno/` is the new place for examples (for Arduino Uno)

I.e. the POC example is now [`examples/arduino-uno/src/bin/uno-blink.rs`](https://github.com/Rahix/avr-hal/blob/the-big-refactor/examples/arduino-uno/src/bin/uno-blink.rs) and you can run it by

```bash
cd examples/arduino-uno
cargo run --bin uno-blink
```

Rahix:
Added support for basic USART, to see how it works out... I ran into one issue which popped up due to the new pins architecture: The current USART code was running into the orphan rule because `UsartOps` is implemented in `avr-hal-generic` and `USART0` is from `avr-device` (and the `Pin` is now from `avr-hal-generic` while it was previously part of the HAL crate).

I've mitigated this by wrapping the "raw peripheral" from the PAC in a `RawPeripheral` wrapper.

Rahix:
All basic features for two boards are there now, looking okay so far.  One big thing I am still unsure about:  Right now I have each HAL and board-crate redefine the `Peripherals` struct.  This is super-convenient but has one big drawback:

Either, only peripherals with a HAL implementation are available or we'd have to explicitly name all not-yet-supported peripherals for each supported MCU individually.  I think option 1 would make the HAL pretty much unusable, but option 2 is a _lot_ of manual code-writing...

Anyone has ideas for an alternative solution?

Rahix:
@explicite, if you want to work on Arduino Mega2560 support, these commits can probably serve as a good basis:

- 356090d155e2 ("atmega-hal: Add support for ATmega32U4")
- 5304fd0d923c ("arduino-hal: Add support for Arduino Leonardo")
- 8f6928cbfd26 ("Add example crate for Arduino Leonardo")

Issue #132, opened 2021-02-09
Feature request: add mode enum for WGMxn (TimernPwm atmega328-hal)


koutoftimer:
It is not handy to use `Timer1Pwm` if you need mod different from default one. 

If I understand right, creating `Timer1Pwm` launches [`init`][init] which sets up `CS2:0` bits and starts timer. Now, if you want to change mode you have to stop timer, by reseting `CS2:0` bits, set right `WGM3:0` bits and reset prescaler because it already have changed it's state. It is impossible to reset prescaler if you need to launch `Timer1Pwm` when some other timers are running because the all share single physical prescaler.

Easiest solution I can see is to implement enum with modes from [datasheet][datasheet] "Table 15-5. Waveform Generation Mode Bit Description (1)" on page 109 and add 3rd parameter to constructor.

[init]: https://github.com/Rahix/avr-hal/blob/master/chips/atmega328p-hal/src/pwm.rs#L95
[datasheet]: https://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-7810-Automotive-Microcontrollers-ATmega328P_Datasheet.pdf

Rahix:
*Meta*: Please refrain from opening this many issues at once.  It comes off as super agressive and just produces noise in everyone's inbox.  Instead please try to explain concisely in a single issue what you're trying to do, what's not working, what you think causes it and what we could do to fix it.  We can then later decide whether this warrants separate issues to track progress on multiple topics.

Anyway, I am certain you acted in good faith here, just please keep this in mind for the future.

---

Regarding your actual issue:

The current `Timer#Pwm` abstraction is not meant and wasn't designed to support other modes than "Fast PWM".  It is supposed to just cover the most simple use-case of a fixed-frequency PWM signal with a duty-cycle that can be varied with 256 steps.  This is enough for e.g. super basic servo control or controlling LED brightness.

I guess the documentation should be much more clear about this, though.

For any more complex usecase I'm afraid the best way right now is to build your own custom abstraction for the time being or just interacting with the timer peripheral directly (like e.g. in [`uno-hc-sr04.rs`](https://github.com/Rahix/avr-hal/blob/master/boards/arduino-uno/examples/uno-hc-sr04.rs)).

In the future I definitely want to have HAL abstractions supporting the more complex use-cases as well, see #44 for example.  Those should be built differently than what we have right now, though, as a part of the refactoring effort in #130 probably.

koutoftimer:
[...]

Issue #44, opened 2020-08-10
Accurate PWM


Rahix:
The existing PWM implementation is fixed to 1kHz and uses the _Fast PWM_ mode which isn't phase- nor frequency-correct.  This is rather useless for applications like controlling a Servo where an exact frequency and duty-cycle are needed.

So, in addition to the _Fast PWM_ implementation a seconds "correct" PWM implementation should be added which allows specifying the frequency + duty cycle exactly.  Ideally, this should be based on the _rate_ and _duration_ primitives from [`embedded-time`](https://docs.rs/embedded-time/0.9.1/embedded_time/index.html).

Open Questions:
- How to select the prescaling?  We could devise an algorithm to use the smallest clock-divider which still fits the frequency.  This would, however make it entirely unobvious what precision the duty-cycle will have.  Would good documentation help here?  Or should the prescaler be selected manually instead?
- Should the `Duty` type be a percentage or a _duration_?

explicite:
1. Looks like prescaler is better option. It's better to be explicite
1. [Duration](https://docs.rs/embedded-time/0.9.1/embedded_time/duration/trait.Duration.html)? Percentage seems not very precise. Percentage can help in some use cases?  
1. Are we able to define this on avr-device?

Rahix:
> Looks like prescaler is better option. It's better to be explicite

Not sure what you mean?  We'll use the prescaler either way, the question is how it is initialized.  Either we have an algorithm which takes a frequency and selects a fitting prescaler, or we leave it up to the user to decide manually.

Actually, while thinking about it: Why not both?  We could provide two separate constructors for 'automatic' and 'manual' configuration ...

> Duration? Percentage seems not very precise.

In terms of precision, the two solutions don't differ too much.  It's always a bit up to luck whether the integer values cleanly divide the timer steps.  But we could, again, build a solution where all possibilities are accounted for and provide multiple methods for setting the duty-cycle based on exact timer counts, based on duration, or based on percentage.

> Aret there any usecases where percentage is common?

Yes, definitely.  PWM is a cheap way to create power output that scales linearly and for those usecases you mostly care about a percentage of the maximum power (= 100% duty cycle).

> Are we able to define this in avr-device?

No, this belongs into the HAL only.

koutoftimer:
IMHO, hardware abstraction is creative process, it is a search of something new, something that didn't exist. I would like to stick with reality and make a handy interface, at least in the beginning.

[...]

PR #136, opened 2021-02-10
Adds parametrized duty for `impl_pwm` macro

koutoftimer:
32u4 require closer look. Not sure what to do with10 bit counter. Other chips verified with datasheets, at least timer1 on 328p works like a charm with 16 bit counter.

### DIFF
--- a/avr-hal-generic/src/pwm.rs
+++ b/avr-hal-generic/src/pwm.rs
@@ -36,6 +36,7 @@ macro_rules! impl_pwm {
     (
         $(#[$timer_pwm_attr:meta])*
         pub struct $TimerPwm:ident {
+            type Duty: $Duty:ty,
             timer: $TIMER:ty,
             init: |$init_timer:ident, $prescaler:ident| $init_block:block,
             pins: {$(
@@ -76,7 +77,7 @@ macro_rules! impl_pwm {
             }
 
             impl $crate::hal::PwmPin for $port::$PXi<$crate::port::mode::Pwm<$TimerPwm>> {
-                type Duty = u8;
+                type Duty = $Duty;
 
                 fn enable(&mut self) {
                     // SAFETY: This block will usually result in a read-modify-write sequence which
@@ -103,7 +104,7 @@ macro_rules! impl_pwm {
                 }
 
                 fn get_max_duty(&self) -> Self::Duty {
-                    u8::MAX
+                    Self::Duty::MAX
                 }
 
                 fn set_duty(&mut self, duty: Self::Duty) {
--- a/chips/atmega168-hal/src/pwm.rs
+++ b/chips/atmega168-hal/src/pwm.rs
@@ -43,6 +43,7 @@ avr_hal_generic::impl_pwm! {
     /// pd5.enable();
     /// ```
     pub struct Timer0Pwm {
+        type Duty: u8,
         timer: crate::pac::TC0,
         init: |tim, prescaler| {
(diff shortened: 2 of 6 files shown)


Rahix:
As I wrote in the other issue, this goes against what the `TimerPwm` abstraction is currently trying to archive:  It is supposed to provide a unified interface for using any compatble pin for PWM without caring too much about the details.

That is the reason why all timers are configured for a `MAX` of 255.  This way, no matter what pin you're using, the `set_duty()` method always gets a value from 0 to 255 representing 0 to full duty cycle.

If we were to change this as you propose here, downstream users would need to know what timer they are using for a specific pin and whether this timer is 8 or 16 (or 10) bits wide.  I think this is not acceptable for an abstraction which tries to hide exactly those details and allow homogeneously treating all PWM pins the same.

As I mentioned over in #132, the way forwards is a separate abstraction (or even multiple) for the different applications of the timer peripherals.  Also ref. #44.

---

> at least timer1 on 328p works like a charm with 16 bit counter.

Are you sure?  The timer is still configured for mode 5 (= "Fast PWM, 8-bit") and will wrap at 0xFF:

https://github.com/Rahix/avr-hal/blob/b1aedf824fcb59078f8022d4464b2835ae4bd83a/chips/atmega328p-hal/src/pwm.rs#L95-L97

Thus, calling `set_duty()` with a value above 255 will just lead to saturating behavior.

koutoftimer:
Code I was using
<details>

```rust
#![feature(llvm_asm)]

#![no_std]
#![no_main]
#![feature(abi_avr_interrupt)]

use arduino_uno::{
    delay_ms, prelude::*,
};
use panic_halt as _;

const MAX_CYCLES: u16 = 40_000; // 50 Hz

// SG 90 servo -90..90 angle mapping to 50..250 byte values
fn duty(value: u8) -> u16 {
    20 * value as u16
}

#[arduino_uno::entry]
fn main() -> ! {
    let pd = arduino_uno::Peripherals::take().unwrap();
    let mut pins = arduino_uno::Pins::new(pd.PORTB, pd.PORTC, pd.PORTD);
    let mut serial = arduino_uno::Serial::new(
        pd.USART0, pins.d0, pins.d1.into_output(&mut pins.ddr),
        9600.into_baudrate(),
    );
    let mut timer1 = arduino_uno::pwm::Timer1Pwm::new(
        pd.TC1, arduino_uno::pwm::Prescaler::Prescale8);
    // FIXME: can't export `avr_device::interrrupt::free`
    {
        let timer = unsafe { &*arduino_uno::pac::TC1::ptr() };
        timer.tccr1a.modify(|_, w| w.wgm1().bits(0b10));
        timer.tccr1b.modify(|_, w| w.wgm1().bits(0b11));
        timer.icr1.modify(|_, w| unsafe { w.bits(MAX_CYCLES) });
    }

    let mut servo = pins.d9.into_output(&mut pins.ddr).into_pwm(&mut timer1);
    servo.set_duty(duty(150));
    servo.enable();

    ufmt::uwriteln!(&mut serial, "Hello from Arduino!\r").void_unwrap();

    loop {
        let b: u8 = (serial.read_byte() - '0' as u8) * 100 +
                (serial.read_byte() - '0' as u8) * 10 +
                (serial.read_byte() - '0' as u8);
        servo.set_duty(duty(b));
        ufmt::uwriteln!(&mut serial, "duty set to {}\r", b).void_unwrap();
    }
}
```

</details>

Moving discussion to appropriate issues.

PR #140, opened 2021-02-16
 arduino-hal: Add support for Arduino Mega 

explicite:
Arduino support like in: https://github.com/Rahix/avr-hal/commit/5304fd0d923cb996b3274c75e3cd4bb0a3ffadfb


explicite:
@Rahix compiled with `cargo build --features arduino-mega` in correct branch ;)

arduino-hal/Cargo.toml
```diff
@@ -12,6 +12,7 @@ board-selected = []
 mcu-atmega=[]
 arduino-uno = ["mcu-atmega", "atmega-hal/atmega328p", "board-selected"]
 arduino-leonardo = ["mcu-atmega", "atmega-hal/atmega32u4", "board-selected"]
+arduino-mega = ["mcu-atmega", "atmega-hal/atmega2560", "board-selected"]
```
Rahix, line 15: The official name is _"Arduino Mega 2560"_ so I think we should reflect this by calling the feature-flag and assorted things `arduino-mega2560`... Just to prevent confusion in the future, should a different version of a "Mega" board ever be released.

arduino-hal/src/clock.rs
```diff
@@ -1,8 +1,10 @@
...
 pub(crate) mod default {
-    #[cfg(feature = "arduino-uno")]
-    pub type DefaultClock = avr_hal_generic::clock::MHz16;
-    #[cfg(feature = "arduino-leonardo")]
+    #[cfg(any(
+        feature = "arduino-uno",
+        feature = "arduino-leonardo",
+        feature = "arduino-mega"
+    ))]
     pub type DefaultClock = avr_hal_generic::clock::MHz16;
```
Rahix, line 9: I know that I've been preaching to use "unified" cfgs whereever possible but I'd make this an exception.  I think it is much easier to see which clock rate associates with which board if they are explicitly separate here.  But YMMV...

Rahix (changes requested):
Thanks, looking really good already!  I've got a few minor nits, apart from that I'm happy :)  I'll do a test-build myself later today as CI doesn't cover this code yet without an examples crate, but I guess with that we're good to go.

Thanks again for helping with this work, I'm quite enjoying having you on board :)

arduino-hal/src/clock.rs
explicite: Splitted.

arduino-hal/Cargo.toml
explicite: Renamed

Rahix (approved):
Looks good, thanks a lot!

PR #147, opened 2021-02-26
add missing bullets on arduino uno pin a5 docs

crclark96:


stappersg:
The actual diff is fine,  at least it looks good to me.

Thing I would like to see in future merge request is that there is a commit message, not just the commit summary.

In this example

```
Added bullets to arduino uno pin doc

Pin a5 of arduino uno now also has enumerating bullets
similar to pins a4, d0 and others.
```

Rahix (approved):
@crclark96, thanks for the PR!  Looks good to me.

@stappersg, I disagree here.  This is overly verbose for a commit that just fixes documentation style.  Detailed commit message do make sense where the change is tricky or complicated but not for situations where the diff is self-explanatory.

stappersg:
On Fri, Feb 26, 2021 at 04:48:15AM -0800, Rahix wrote:
> @stappersg, I disagree here.  This is overly verbose for a commit that
> just fixes documentation style.  Detailed commit message do make sense
> where the change is tricky or complicated but not for situations where
> the diff is self-explanatory.


True


(to express the common ground we have)

PR #149, opened 2021-02-27
atmega-hal: atmega328pb

explicite:
atmega328pb mcu support like https://github.com/Rahix/avr-hal/commit/356090d155e2ab02599b61a0f0cd2f330c8cd4 and avr-spec


avr-specs/avr-atmega328pb.json
```diff
@@ -0,0 +1,28 @@
+{
+    "arch": "avr",
+    "atomic-cas": false,
+    "cpu": "atmega328pb",
```
explicite, line 4: @Rahix looks like I cannot with this spec. Not sure how to solve this problem

Rahix: @jkristell, you originally added ATmega328PB support.  What target spec were you compiling with?

jkristell: I used the atmega328p spec.

Rahix: Thanks!  @explicite, in that case just drop the custom spec and use `avr-atmega328p.json` in the CI.

Rahix (approved):
Looks good to me, thanks!

PR #156, opened 2021-03-06
Add a new tool for deploying binaries onto an MCU: ravedude

Rahix:
ravedude is a wrapper around avrdude to make Rust development for AVR microcontrollers as easy as possible.  ravedude is meant to replace the current hackish runner scripts with something that works out of the box.

In addition to managing the avrdude invocation, ravedude also comes with a super basic terminal emulator to make accessing the serial console easy.  In a similar spirit to `probe-run`, this should provide an embedded development experience which is as close to "desktop development" as possible.

It looks like this:

ravedude


Rahix:
@explicite, I do not have a Mega 2560 lying around at the moment, so I didn't add support for it yet.  If you're interested, you can take a look at 135eef3b84d9 ("ravedude: Add support for Arduino Leonardo") and try doing the same for Mega 2560.

PR #157, opened 2021-03-07
atmega-hal:attiny85 support

explicite:
atmega-hal: ATtiny85 support

### DIFF
--- a/mcu/attiny-hal/src/lib.rs
+++ b/mcu/attiny-hal/src/lib.rs
@@ -1,0 +1,43 @@
+#![no_std]
+
+#[cfg(all(
+    not(feature = "device-selected"),
+    not(feature = "disable-device-selection-error")
+))]
+compile_error!(
+    "This crate requires you to specify your target chip as a feature.
+
+    Please select one of the following
+
+    * attiny85
+    "
+);
+
+/// Reexport of `attiny85` from `avr-device`
+#[cfg(feature = "attiny85")]
+pub use avr_device::attiny85 as pac;
+
+/// See [`avr_device::entry`](https://docs.rs/avr-device/latest/avr_device/attr.entry.html).
+#[cfg(feature = "rt")]
+pub use avr_device::entry;
+
+#[cfg(feature = "device-selected")]
+pub use pac::Peripherals;
+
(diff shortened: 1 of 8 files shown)


Rahix:
Ah, sorry, the ATtiny's shouldn't go into `atmega-hal`, they should have their own `attiny-hal` crate... :/

Rahix:
Do you want to have a go at creating it or should I build up the base for you to work with?  Essentially, it should be a mirror copy of `atmega-hal` without any major differences.

mcu/attiny-hal/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+pub use avr_hal_generic::delay;
+
+#[cfg(feature = "device-selected")]
+pub mod port;
+#[cfg(feature = "device-selected")]
+pub use port::Pins;
+
+pub struct ATtiny;
+
+#[cfg(feature = "attiny85")]
```
explicite, line 37: here we are not using any I2C; removed.
when I will define SPI after attinuy88 then will add definition

mcu/attiny-hal/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+
+pub use avr_hal_generic::clock;
+pub use avr_hal_generic::delay;
+
+#[cfg(feature = "device-selected")]
+pub mod port;
+#[cfg(feature = "device-selected")]
+pub use port::Pins;
+
+pub struct ATtiny;
```
Rahix, line 35: Minor nit-pick: To be consistent with `atmega-hal`, this should be named `Attiny` with a small `t`:

```suggestion
pub struct Attiny;
```

mcu/attiny-hal/src/lib.rs
Rahix: Ack.  Most ATtinies don't have TWI or SPI peripherals, AFAIK.  They all use this `USI` "Universal Serial Interface" which we don't yet support.  So yeah, you are right to remove it here.

Rahix (commented):
Got one super minor nit-pick apart from that it looks great!  Thanks, btw, for helping out so much, I really appreciate it :)

PR #158, opened 2021-03-07
atmega-hal ATtiny88 support

explicite:
Draft, #157 should be merged as first


explicite:
Will move to `attiny-hal` after #157

PR #160, opened 2021-03-09
add the atmega8u2 chip

drmorr0:
I confirmed that this compiles.  Once the `avr-device` version is bumped and a new crate is built, we can remove the revision pins in the various `Cargo.toml`s and this "should" be ready to ship.

### DIFF
--- a/ravedude/src/main.rs
+++ b/ravedude/src/main.rs
@@ -85,7 +85,7 @@ fn ravedude() -> anyhow::Result<()> {
         Ok,
     )?;
 
-    if let Some(bin) = args.bin {
+    if let Some(bin) = args.bin.as_ref() {
         task_message!(
             "Programming",
             "{} {} {}",
@@ -94,7 +94,7 @@ fn ravedude() -> anyhow::Result<()> {
             port.display()
         );
 
-        let mut avrdude = avrdude::Avrdude::run(&board.avrdude_options(), &port, &bin)?;
+        let mut avrdude = avrdude::Avrdude::run(&board.avrdude_options(), &port, bin)?;
         avrdude.wait()?;
 
         task_message!("Programmed", "{}", bin.display());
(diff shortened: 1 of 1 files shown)


arduino-hal/Cargo.toml
```diff
@@ -37,4 +37,5 @@ optional = true
 features = ["disable-device-selection-error"]
 
 [dependencies.avr-device]
-version = "0.3"
+git = "https://github.com/Rahix/avr-device"
+rev = "92e4d5f"
```
Rahix, line 41: Tip: The best way to substitute `avr-device` with a local version for all crates in `avr-hal` is via a patch in the top-level `Cargo.toml` file (where the workspace is defined):

```toml
[patch.crates-io]
avr-device = { git = "https://github.com/Rahix/avr-device", rev = "92e4d5f" }
```

Then you won't have to touch every single Cargo.toml across the repository and it is also less work to undo once everything is ready.

mcu/atmega-hal/src/port.rs
```diff
@@ -1,4 +1,9 @@
-#[cfg(any(feature = "atmega48p", feature = "atmega168", feature = "atmega328p"))]
+#[cfg(any(
+    feature = "atmega8u2",
```
Rahix, line 2: I don't think this is the right place.  ATmega8U2 has a pin PC7 which is not present on the other chips listed here.  The `impl_port_traditional!` macro sadly doesn not support conditional compilation for single pins yet so you'll have to duplicate the block for the time being (or check if one of the other impls looks better).

arduino-hal/Cargo.toml
drmorr0: Ah, TIL.

mcu/atmega-hal/src/port.rs
drmorr0: Oh good catch, I missed this.  Thanks!

Rahix:
Hm, why did you close this PR?  Was is by accident?  You seem to have removed all changes from this PR by a force-push?

I'd still go ahead with it once the next release of `avr-device` hits.

drmorr0:
Ugh.  Yea, my local git repo got messed up and I guess I screwed this up.  I re-opened it in #170

Issue #130, opened 2021-02-07
Tracking issue for the architectural refactor


Rahix:
Starting with PR #129, we will refactor the crate structure of `avr-hal` to better fit the different requirements.  This issue will serve to track the progress of the refactor up to the point where it can be merged into the `avr-hal` main branch.  Progress will be happening on the [`next` branch].

[`next` branch]: https://github.com/Rahix/avr-hal/tree/next

#### Goals
- Reduce code-duplication as much as possible.  Currently each MCU HAL is essentially a copy of the others with minor details changed.  This is horrible on maintainability and hinders fast evolution of APIs.  Two issues related to this are #94 and #117.
- Support the newer generation AVR devices alongside the old ones without much effort.  The user-facing APIs should be exactly the same (modulo intrinsic hardware differences).  Also see #114 and #116.
- Provide a single `arduino-hal` which allows to write code that can be effortlessly used on any supported board (as long as all necessary hardware features are available).  See #117.

#### Tasks
- [ ] Build basic crate layout for the new design (*Pull Request*: #129).
- [ ] Redesign HAL APIs/internals for more flexibility towards supported devices
  - [ ] Design a new port API that also works for the newer IO cell (*Pull Request*: #129).
    - [ ] Implement this API for at least one device with the new IO cell.
  - [ ] Overhaul USART HAL.
  - [ ] Overhaul SPI HAL.
  - [ ] Overhaul I2C HAL.
  - [ ] Overhaul ADC HAL.
  - [ ] Overhaul PWM HAL.
- [ ] Reintroduce support for all MCUs
  - [ ] `ATmega1280`
  - [ ] `ATmega168`
  - [ ] `ATmega2560`
  - [ ] `ATmega328PB`
  - [ ] `ATmega328P` (*Pull Request*: #129)
  - [ ] `ATmega32U4`
  - [ ] `ATmega48P`
  - [ ] `ATtiny85`
  - [ ] `ATtiny88`
- [ ] Implement support for `ATmega4809` as the first newer generation device.
- [ ] Reimplement board-support for all previously supported boards
  - [ ] Arduino Diecimila
  - [ ] Arduino Leonardo
  - [ ] Arduino Mega2560
  - [ ] Arduino Uno (*Pull Request*: #129)
  - [ ] Bigavr6
  - [ ] Sparkfun Pro Micro
  - [ ] Trinket
- [ ] Implement support for the _Arduino Nano Every_.
- [ ] Anything else?

[...]

Issue #54, opened 2020-08-18
Create a project template


Rahix:
Similar to [`cortex-m-quickstart`](https://github.com/rust-embedded/cortex-m-quickstart), we should provide a [`cargo-generate`](https://crates.io/crates/cargo-generate) template for AVR.  I'm not sure how exactly to design that to make it easy to start with any device from the same template but I'm sure there is a way to make this work.  Otherwise, several templates for different popular boards could also work.

ToddG:
I wrote this for my own use a few days ago: 

https://github.com/ToddG/arduino-rust-seed

Basically it uses python + cookiecutter to generate an application from templates.

koutoftimer:
I vote for orphan `template-{board name}` branches. Despite of branching support in liquid templates it will be a mess to write common template for every supported board. Separate branches are harder to support but easier to understand what is going on and easier to get involved for newcomers.

Rahix:
I like the idea of having the templates as branches in this repo!  We just need to make it super obvious where they are, because otherwise people will not find them...

I wouldn't go with a separate template for each board, but I do agree that we might need multiple templates for very diverging designs.  But I would say that, for example, every board supported by the new `arduino-hal` from #117 can be supported from a single template.

I saw that `cargo-generate` is making progress on their issue https://github.com/cargo-generate/cargo-generate/issues/17, so I think we can make use of that in the near future :)

Rahix:
For reference: The template should contain a `rust-toolchain` file (https://github.com/Rahix/avr-hal/issues/124#issuecomment-795585670 and [the toolchain file docs](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file)).

Rahix:
<https://github.com/Rahix/avr-hal-template>

PR #162, opened 2021-03-09
Add cargo generate template for start project

HKalbasi:
I create a [template](https://github.com/HKalbasi/rust-arduino-generator) via [cargo generate](https://github.com/cargo-generate/cargo-generate) for bootstrap new projects.

Currently it supports `arduino-uno` and `arduino-leonardo` (leonardo is untested, I just have one uno at home) but adding boards is very easy.

I think it is useful for beginners as well as professionals. But there is valid reasons for rejecting this PR (like cargo generate is slow to install)

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,23 @@ avr-objcopy -S -j .text -j .data -O ihex uno-blink.elf uno-blink.hex
 ```
 
 ## Starting your own project
+
+### Automated way
+1. Install [cargo-generate](https://github.com/ashleygwilliams/cargo-generate)
+2. Init new project:
+   ```
+   cargo generate --git https://github.com/rustwasm/wasm-pack-template.git
+   <answer questions>
+   cd <your-project-name>
+   ```
+3. If you're using rustup, you probably want to set an override for this directory, to use the nightly toolchain (as mentioned above, use `nightly-2021-01-07` for the time being):
+   ```bash
+   rustup override set nightly-2021-01-07
+   ```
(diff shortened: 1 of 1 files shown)


Rahix:
Funnily enough just yesterday I randomly got to chat with one of the `cargo-generate` maintainers about using it for projects in the embedded space, like `avr-hal`.

So yes, a template is very much a thing we want, to lower the entry barrier and to prevent the load of issues we get from people accidentally skipping steps in the instructions.  There is also an issue about it already: #54 :)  The plan is to track the template in a separate branch of this repository for now.

As we're currently doing a big refactor/rewrite of `avr-hal` (see #130), I think we should wait with providing the template until that's completed.  So I'd leave this PR open until we're done and then pull in your work as a basis, if that's okay with you :)

HKalbasi:
:)) I did it simply because no one had done it before. (I was not aware of #54) I saw this project today (Searching "how I can program my arduino using rust") and found it unnecessarily hard to start with (relative to arduino ide and C) so I tried to make a step forward.

There PR is just for replacing initial steps in `Readme.md` and template is in [this repo](https://github.com/HKalbasi/rust-arduino-generator). It already works and it should be easy to update it when that great upgrade become ready. But if you see benefit in delaying it until then, that is ok.

I don't think track the template in a branch is a good idea. Specially if `cargo-generate` clone whole repo it will be slow to download. Keeping template in the tree make sense because of syncing template with code but `cargo-generate` can't handle this AFAIK. But if you decide to do this or making a separate repository, feel free to use my code and tell me to deprecate my template in favor of yours.

HKalbasi:
Seems it is fixed in next branch. Thanks for great work!

PR #168, opened 2021-03-22
Implement initial state for output pins

Javier-varez:
Currently there is no way to set the initial state of an output pin, which can be very problematic for critical signals, as by default they could be in either high or low depending on the corresponding bit of the port register.

This change provides an alternative API for users who want to set the initial state of a pin when switching to output.

### DIFF
--- a/avr-hal-generic/src/port.rs
+++ b/avr-hal-generic/src/port.rs
@@ -107,8 +107,21 @@ impl<PIN: PinOps> Pin<mode::Input<mode::Floating>, PIN> {
 /// does it have the mode-relevant methods availailable (e.g. `set_high()` is only available for
 /// `Output` pins).
 impl<PIN: PinOps, MODE: mode::Io> Pin<MODE, PIN> {
-    /// Convert this pin into an output pin.  See [Digital Output](#digital-output).
+    /// Convert this pin into an output pin, setting the state to low.
+    /// See [Digital Output](#digital-output).
     pub fn into_output(mut self) -> Pin<mode::Output, PIN> {
+        unsafe { self.pin.out_clear() };
+        unsafe { self.pin.make_output() };
+        Pin {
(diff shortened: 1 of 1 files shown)


Rahix:
Hi, thanks a lot for the PR!  This is a very sensible addition, I agree.

However, we're currently reworking the design of `avr-hal`, so the version you added this to is not going to exist for much longer... You can check out the new crates on the `next` branch.  This is the relevant part:

https://github.com/Rahix/avr-hal/blob/12c28ae64217e8a1f508657aa099740cffe7f5ef/avr-hal-generic/src/port.rs#L110-L117

There was a previous comment by someone else about this also leaving the pin in a non-deterministic state depending on what was configured beforehand.  I think the best course of action is to make `into_output()` always put the pin into a low state and adding a second method `into_output_high()` which always puts it high.  E.g.

```rust
    pub fn into_output(mut self) -> Pin<mode::Output, PIN> {
        unsafe {
			self.pin.out_clear();
			self.pin.make_output();
		}
        Pin {
            pin: self.pin,
            _mode: PhantomData,
        }
    }
```

What do you think? Would you be interested on reimplementing this based on the new code in the `next` branch?

stappersg:
On Mon, Mar 22, 2021 at 02:05:16PM -0700, Rahix wrote:
> However, we're currently reworking the design of `avr-hal`

More information about that work in #130

Javier-varez:
Hi! Thanks for the quick reply! I agree that it would be a good idea as well to configure the port as low when into_output is called. I will move this PR to the `next` branch!

Rahix (approved):
Neat, thanks a lot! :)

I guess at some point we might want to document why one would want to use this method, maybe with an example, but I'll be doing a sweep over the entire codebase and add all such things once #130 is wrapped up.

PR #169, opened 2021-03-25
Add explicit support for arduino-nano

drmorr0:
1. Add the arduino-nano feature flag to arduino-hal.
2. Make the ADC6 and ADC7 pins only available for nano, not for uno.
3. Add an example crate for arduino-nano with just a blink and ADC example.

### DIFF
--- a/avr-hal-generic/src/port.rs
+++ b/avr-hal-generic/src/port.rs
@@ -107,8 +107,21 @@ impl<PIN: PinOps> Pin<mode::Input<mode::Floating>, PIN> {
 /// does it have the mode-relevant methods availailable (e.g. `set_high()` is only available for
 /// `Output` pins).
 impl<PIN: PinOps, MODE: mode::Io> Pin<MODE, PIN> {
-    /// Convert this pin into an output pin.  See [Digital Output](#digital-output).
+    /// Convert this pin into an output pin, setting the state to low.
+    /// See [Digital Output](#digital-output).
     pub fn into_output(mut self) -> Pin<mode::Output, PIN> {
+        unsafe { self.pin.out_clear() };
+        unsafe { self.pin.make_output() };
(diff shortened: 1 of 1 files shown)


examples/arduino-nano/.cargo/config.toml
```diff
@@ -0,0 +1,8 @@
+[build]
+target = "../../avr-specs/avr-atmega328p.json"
+
+[target.'cfg(target_arch = "avr")']
+runner = "ravedude uno -cb 57600"
```
Rahix, line 5:
```suggestion
runner = "ravedude nano -cb 57600"
```

mcu/atmega-hal/Cargo.toml
```diff
@@ -7,6 +7,7 @@ edition = "2018"
 [features]
 rt = ["avr-device/rt"]
 device-selected = []
+disable-unused-adc = []
```
Rahix, line 10: In general, features should always be opt-in instead of opt-out, because features are selected based on the _union_ of all features requested by downstream crates.  I.e. suppose crate A sets `disable-unused-adc` while crate B expects the pins to be available - this wouldn't compile.  If, OTOH you'd have an `additional-adc-pins` feature which crate A ignores and crate B sets explicitly - all is well.

This also means you can get rid of the `not()` you have to use everywhere.

Rahix (changes requested):
Hey, thanks a lot for the PR!  I've got one thing I'd like to see differently: Instead of the "negative" opt-out feature I'd prefer a "positive" opt-in feature because this composes better with the way cargo merges features.

mcu/atmega-hal/Cargo.toml
drmorr0: yep, this makes sense, thanks!

PR #171, opened 2021-04-04
Arduino Uno - example HC-SR04 - Handling panic due to overflow

peter-simon:
Some HC-SR04 sensor builds keep their echo pin in high state for long time, instead of keeping it in low state when no object is detected. This is mentioned here for example: https://forum.arduino.cc/index.php?topic=55119.0

When this happens the multiplication of the TIMER1 value with 4 goes into overflow, resulting panic, crashing the program execution.

The proposed change solves this issue by handling the saturating u16 value the same way as if the echo pin never had gone high.


Rahix (approved):
Ah, good to know.  Thanks a lot for the fix!

Issue #40, opened 2020-08-09
Add usb-device support to atmega32u4


TeXitoi:
Do you plan to support https://github.com/mvirkkunen/usb-device on atmega32u4? It would validate the usb-device abstraction and would allow to use rust to implement USB keyboard on most QMK compatible keyboards.

Rahix:
Yes, definitely!  Though I don't have much experience with low-level USB so it'll take some time ... If anyone else wants to take a stab at it, please go for it!

mogenson:
Hey, I'm trying to bring up the USB hardware, based on the LUFA and Teensy libraries. However, I can't seem to get the `USB_GEN()` interrupt to trigger from an end of reset event. Here's an example below. Do you see the same behavior?

```
#![no_std]
#![no_main]
#![feature(abi_avr_interrupt)]

use arduino_leonardo::prelude::*;
use panic_halt as _;

use arduino_leonardo::hal::port;
static mut PIN: Option<port::portc::PC7<port::mode::Output>> = None;
static mut USB: Option<arduino_leonardo::atmega32u4::USB_DEVICE> = None;

#[arduino_leonardo::entry]
fn main() -> ! {
    let dp = arduino_leonardo::Peripherals::take().unwrap();

    let mut pins = arduino_leonardo::Pins::new(dp.PORTB, dp.PORTC, dp.PORTD, dp.PORTE);

    let mut led0 = pins.led_rx.into_output(&mut pins.ddr);
    let mut led1 = pins.led_tx.into_output(&mut pins.ddr);

    let mut led = pins.d13.into_output(&mut pins.ddr);

    led0.set_high().void_unwrap();
    led1.set_low().void_unwrap();
    led.set_low().void_unwrap();

    unsafe {
        PIN = Some(led);
    }

    dp.USB_DEVICE.usbcon.reset();
    dp.USB_DEVICE.uhwcon.reset();
    dp.USB_DEVICE.udcon.reset();
    dp.USB_DEVICE.udien.reset();

    dp.USB_DEVICE.uhwcon.write(|w| w.uvrege().set_bit()); // enable pad regulator
    dp.USB_DEVICE.usbcon.write(|w| w.frzclk().set_bit()); // freeze usb clock
    dp.PLL.pllcsr.write(|w| w.pindiv().set_bit()); // usb 16 mhz clock source
    dp.PLL.pllcsr.write(|w| w.plle().set_bit()); // enable PLL
[...]

PR #174, opened 2021-04-09
Add Arduino Micro support

jnnksdev:
This PR will slowly, but steadily add support for ATMega32u4 based Arduino Micro starting with ravedude.


Rahix:
Hey @jnnksdev, thanks a lot for adding Arduino Micro support.  Is there anything you need help with regarding the USART example?  It should look exactly like the one for Arduino Leonardo.  Keep in mind that the ATmega32u4's USART is **not** connected to the host USB like on Arduino Uno for example, but is only available via pins D1 and D2, if those are connected to an external USB-to-serial converter.  So you will not see any TTY device over the standard USB until we have USB serial support working (see issue #40).

Also, don't forget to mark this PR as "ready for review" once you want me to take a look!

jnnksdev:
Hi @Rahix, yes I came to realize that the serial is not working as straight forward and will experiment with an Arduino Uno to validate it working.

In regards to Serial-to-Host I was investigating here: [ArduinoCore-avr/.../HardwareSerial.cpp](https://github.com/arduino/ArduinoCore-avr/blob/9f8d27f09f3bbd1da1374b5549a82bda55d45d44/cores/arduino/HardwareSerial.cpp#L117), but lack the experience to understand what is going on.

Rahix:
The difference between Uno and Leonardo/Micro is this:

- On _Arduino Uno_, the hardware serial (USART) is connected to a secondary AVR microcontroller which bridges it to USB.  That's where you connect your host computer to and because of the onboard USB-to-Serial converter it just works straight away with the hardware serial.
- On _Arduino Leonardo_ and _Arduino Micro_, this secondary AVR microcontroller does not exist because the ATmega32U4 has a USB peripheral built-in.  The idea is that this allows you to do more than plain USB-Serial over USB, for example the board can present itself as a keyboard or mouse.

  The hardware serial still exists and is connected to pins D0 and D1.  If you have an external USB-to-Serial device, you could connect it to those pins and get a hardware serial connection that way.

  But to get the usual plug-and-it-works connection over USB, the code on the ATmega32U4 needs to actually implement a USB driver which talks the CDC-ACM protocol.  Such a driver is still missing from `avr-hal` as discussed in #40.

So to move forward, I would just not include a serial example for now and revisit that point later once USB support is added.

jnnksdev:
> The difference between Uno and Leonardo/Micro is this: ...

Yes thanks I understand that.

___ 

> So to move forward, I would just not include a serial example for now and revisit that point later once USB support is added.

Sooner or later I will have to face USART though. I am building a custom keyboard and will need the "_HID mode_" at some point.
Furthermore checking the other implementations with serial to USB logging may make debugging a lot easier.
Since there is a working serial-to-USB implementation in the official C++ Arduino repo I will try to replicate what is happening there until I am frustrated :)

jnnks:
This PR is on halt for an indefinite time.
I'll get back to it when AVR USB support is easier to implement.

Usable Implementations for ravedude have been opened as separate PR https://github.com/Rahix/avr-hal/pull/185

PR #175, opened 2021-04-10
Refactored SPI support

drmorr0:
Hi, this is still a work-in-progress, but wanted to get some early feedback, if you have some time.

### DIFF
--- a/avr-hal-generic/src/spi.rs
+++ b/avr-hal-generic/src/spi.rs
@@ -179,8 +179,9 @@ where
         sclk: port::Pin<port::mode::Output, SCLKPIN>,
         mosi: port::Pin<port::mode::Output, MOSIPIN>,
         miso: port::Pin<port::mode::Input<port::mode::Floating>, MISOPIN>,
+        cs: port::Pin<port::mode::Output, CSPIN>,
         settings: Settings,
-    ) -> Self {
+    ) -> (Self, ChipSelectPin<CSPIN>) {
         let mut spi = Self {
             p,
             sclk,
@@ -191,7 +192,7 @@ where
             _h: PhantomData,
         };
         spi.p.raw_setup(&settings);
-        spi
+        (spi, ChipSelectPin(cs))
     }
 
     /// Reconfigure the SPI peripheral after initializing
--- a/avr-hal-generic/src/spi.rs
+++ b/avr-hal-generic/src/spi.rs
@@ -74,7 +74,8 @@ impl Default for Settings {
 pub trait SpiOps<H, SCLK, MOSI, MISO, CS> {
     fn raw_setup(&mut self, settings: &Settings);
     fn raw_release(&mut self);
-    fn raw_flush(&mut self) -> nb::Result<(), void::Void>;
+
+    fn raw_check_iflag(&self) -> bool;
     fn raw_read(&self) -> u8;
     fn raw_write(&mut self, byte: u8);
 }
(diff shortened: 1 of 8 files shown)


avr-hal-generic/src/spi.rs
```diff
@@ -64,220 +66,258 @@ impl Default for Settings {
...
-        }
-    };
-    
+/// Wrapper for the CS pin
+///
+/// Used to contain the chip-select pin during operation to prevent its mode from being
+/// changed from Output. This is necessary because the SPI state machine would otherwise
+/// reset itself to SPI slave mode immediately. This wrapper can be used just like an
+/// output pin, because it implements all the same traits from embedded-hal.
+pub struct ChipSelectPin<CSPIN: hal::digital::v2::OutputPin>(CSPIN);
```
Rahix, line 88: It is good practice not to have trait bounds on the struct, but only on the impls (if possible).  Otherwise this might become a semver-compat hazard.

avr-hal-generic/src/spi.rs
```diff
@@ -64,220 +66,258 @@ impl Default for Settings {
...
+    /// The pins are not actually used directly, but they are moved into the struct in
+    /// order to enforce that they are in the correct mode, and cannot be used by anyone
+    /// else while SPI is active.  CS is placed into a `ChipSelectPin` instance and given
+    /// back so that its output state can be changed as needed.
+    pub fn new(
+        p: SPI,
+        sclk: port::Pin<port::mode::Output, SCLKPIN>,
+        mosi: port::Pin<port::mode::Output, MOSIPIN>,
+        miso: port::Pin<port::mode::Input<port::mode::PullUp>, MISOPIN>,
+        cs: CSPIN,
```
Rahix, line 168: I think this should also be `port::Pin<port::mode::Output, CSPIN>` here and everywhere else?

avr-hal-generic/src/spi.rs
```diff
@@ -64,220 +66,258 @@ impl Default for Settings {
...
+macro_rules! impl_spi {
+    (
+        hal: $HAL:ty,
+        peripheral: $SPI:ty,
+        sclk: $sclkpin:ty,
+        mosi: $mosipin:ty,
+        miso: $misopin:ty,
+        cs: $cspin:ty,
+    ) => {
+        static mut IS_WRITE_IN_PROGRESS: bool = false;
```
Rahix, line 284: This has multiple issues:

1. It means you cannot use the `impl_spi!{}` macro multiple times in the same module due to how macros work.
2. It is an unsynchronized global variable so it can very easily lead to memory races.  That's no good.

I think it's the wrong approach to have this here.  This bitflag should be stored in the `Spi<>` higher level type and the lower level APIs (`SpiOps`) should be designed such that the flag is checked and set in the `Spi<>` methods.  As an example, the following could work:

```rust
impl SpiOps for ... {
    fn raw_write(&mut self, byte: u8) {
        self.spdr.write(|w| unsafe { w.bits(byte) });
    }

    fn raw_check_iflag(&mut self) -> bool {
        self.spsr.read().spif().bit_is_set()
    }
}

impl Spi<...> {
    fn write_internal(&mut self, byte: u8) {
        self.is_write_in_progress = true;
        self.p.raw_write(byte);
    }

    fn flush_internal(&mut self) -> nb::Result<(), void::Void> {
        if self.is_write_in_progress {
            if self.p.raw_check_iflag() {
                self.is_write_in_progress = false;
            } else {
                return Err(nb::Error::WouldBlock);
            }
        }
        Ok(())
    }
}

impl FullDuplex for Spi<...> {
    fn send(&mut self, byte: u8) -> nb::Result<(), Self::Error> {
        self.flush_internal()?;
        self.write_internal(byte);
        Ok(())
    }

    fn read(&mut self) -> nb::Result<u8, Self::Error> {
        self.flush_internal()?;
        Ok(self.p.raw_read())
    }
}
```

Rahix (commented):
Hey,

I took a quick look and left a few comments.  Overall it's already looking quite good, thanks for working on this!

avr-hal-generic/src/spi.rs
drmorr0: Yea, thanks, I didn't love this solution but was just trying to get something to compile.  Your solution is better, I think.

avr-hal-generic/src/spi.rs
drmorr0: I think I fixed this; it compiles, at least?

avr-hal-generic/src/spi.rs
```diff
@@ -64,220 +66,328 @@ impl Default for Settings {
...
+            sclk,
+            mosi,
+            miso: miso.forget_imode(),
+            write_in_progress: false,
+            _cs: PhantomData,
+            _h: PhantomData,
+        };
+        spi.p.raw_setup(&settings);
+        spi
+    }
```
Rahix, line 209: `with_external_pullup()` seems to be missing the chip-select pin?

drmorr0: I was copying from here: https://github.com/Rahix/avr-hal/blob/master/avr-hal-generic/src/spi.rs#L198

I'm not too sure why the function on `master` was missing this, but I assumed there was a reason?  If this is just a bug/typo I can add it back in.

Rahix: Most likely something I missed during review of the relevant PR.  It is definitely supposed to be here, otherwise things won't work correctly.

drmorr0:
I only have an uno and a nano, so I can't test this on other boards, but I just copied what was already there in `master` and the type-checking all worked out.  I think this is ready to go?  Let me know if you have other feedback.

Rahix (approved):
Looks good to me!  Don't worry too much about breaking anything, we'll fix it when the bug reports come in - that's the nice thing about experimental software ;)

PR #180, opened 2021-05-01
minor amendment to SPI code

drmorr0:
## Summary

If you want to write code that doesn't want to depend on embedded-hal for some reason, there's no way to do that without this fix since there's no "native" receive function.  This change adds that, and it makes the embedded-hal trait implementation depend on the "native" functions.

## Testing done
Works on my nano.

### DIFF
--- a/avr-hal-generic/src/spi.rs
+++ b/avr-hal-generic/src/spi.rs
@@ -231,6 +231,10 @@ where
         Ok(())
     }
 
+    fn receive(&mut self) -> u8 {
+        self.p.raw_read()
+    }
+
     fn write(&mut self, byte: u8) {
         self.write_in_progress = true;
         self.p.raw_write(byte);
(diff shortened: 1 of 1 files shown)


Rahix (approved):
I like the change because it makes the code more consistent with the implementation for sending.

But I do not understand your rationale behind it:  `receive()`, `flush()`, and `write()` are all private functions so downstream code won't be able to call them anyway... I do agree that it would be nice to provide an "inherent" interface alongside the trait implementations (also see https://github.com/rust-embedded/embedded-hal/issues/177#issuecomment-821083026) but I think we can wait with adding that until upgrading to a newer version of _embedded-hal_.

drmorr0:
> I like the change because it makes the code more consistent with the implementation for sending.
> 
> But I do not understand your rationale behind it: `receive()`, `flush()`, and `write()` are all private functions so downstream code won't be able to call them anyway... I do agree that it would be nice to provide an "inherent" interface alongside the trait implementations (also see [rust-embedded/embedded-hal#177 (comment)](https://github.com/rust-embedded/embedded-hal/issues/177#issuecomment-821083026)) but I think we can wait with adding that until upgrading to a newer version of _embedded-hal_.

Whoops, yea, you're right, I wasn't thinking.

PR #185, opened 2021-06-18
ravedude: Add support for Arduino Micro

jnnks:
This PR copies the MC definition from `leonardo` and adds it explicitly as `micro`.

### DIFF
--- a/ravedude/src/board.rs
+++ b/ravedude/src/board.rs
@@ -79,7 +79,7 @@ impl Board for ArduinoMicro {
     }
 
     fn needs_reset(&self) -> Option<&str> {
-        None
+        Some("Reset the board by pressing the reset button once.")
     }
 
     fn avrdude_options(&self) -> avrdude::AvrdudeOptions {
@@ -92,7 +92,14 @@ impl Board for ArduinoMicro {
     }
 
     fn guess_port(&self) -> Option<std::path::PathBuf> {
-        None
+        find_port_from_vid_pid_list(&[
+            (0x2341, 0x0037),
+            (0x2341, 0x8037),
+            (0x2A03, 0x0037),
+            (0x2A03, 0x8037),
+            (0x2341, 0x0237),
+            (0x2341, 0x8237),
+        ])
     }
 }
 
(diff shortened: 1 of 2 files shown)


ravedude/src/board.rs
```diff
@@ -70,6 +71,31 @@ impl Board for ArduinoUno {
...
+        avrdude::AvrdudeOptions {
+            programmer: "avr109",
+            partno: "atmega32u4",
+            baudrate: Some(115200),
+            do_chip_erase: true,
+        }
+    }
+
+    fn guess_port(&self) -> Option<std::path::PathBuf> {
+        None
```
Rahix, line 95: Actually, there are VID:PIDs to detect this board (taken from [`boards.txt`](https://github.com/arduino/ArduinoCore-avr/blob/24e6edd475c287cdafee0a4db2eb98927ce3cf58/boards.txt#L358-L369)):

```suggestion
        find_port_from_vid_pid_list(&[
            (0x2341, 0x0037),
            (0x2341, 0x8037),
            (0x2A03, 0x0037),
            (0x2A03, 0x8037),
            (0x2341, 0x0237),
            (0x2341, 0x8237),
        ])
```

ravedude/src/board.rs
```diff
@@ -70,6 +71,31 @@ impl Board for ArduinoUno {
...
 
+struct ArduinoMicro;
+
+impl Board for ArduinoMicro {
+    fn display_name(&self) -> &str {
+        "Arduino Micro"
+    }
+
+    fn needs_reset(&self) -> Option<&str> {
+        None
```
Rahix, line 82: If this board is similar to Leonardo, you'll need to reset the board manually before programming.  To instruct the user that this is necessary, a message should be added here:

```suggestion
        Some("Reset the board by pressing the reset button once.")
```

Of course you'll need to check whether this is correct - some boards require other reset procedures like e.g. a quick double-press.

Rahix (commented):
Thanks for the PR!  Please check the two comments I've left below, otherwise everything looks good (I do not have hardware to test it, though!).

jnnks:
Thanks for the feedback, I applied the suggested changes.

PR #186, opened 2021-07-02
Automatically install necessary toolchain

MarkuBu:
In this pull request I added the rust-toolchain.toml file. Now the necssary toolchain and rust-src will be installed automatically. If another toolchain than 2021-01-07 can be used you simply can change it in this place.

I also modified README.md to match the changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,11 +9,7 @@ avr-hal ![Continuous Integration](https://github.com/Rahix/avr-hal/workflows/Con
   - [Supported Boards](#supported-boards)
 
 ## Quickstart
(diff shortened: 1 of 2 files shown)


MarkuBu:
You can also specify a target if necessary with

`targets = [ "wasm32-unknown-unknown", "thumbv2-none-eabi" ]`

Targets are just an example

Rahix (approved):
Yeah, I wanted to add this as well, but didn't get around to it so far.  Thanks!

> You can also specify a target if necessary with

This doesn't apply for us because there is no prebuilt AVR libcore (yet).  We use `build-std` to compile it on the fly for the exact MCU that is used.

---

One note though:  The `master` branch is on the verge of being deprecated, any new work is going on in the `next` branch at the moment.  So maybe you can also send a PR against that branch with the same changes?

MarkuBu:
> So maybe you can also send a PR against that branch with the same changes?

I noticed too late that the master is a bit out of date and you are working on the next branch. Will send a PR soon.

PR #177, opened 2021-04-23
Add refactored support for Arduino Diecimila

explicite:
Support for Diecimila
- [x] arduino-hal 
- [x] examples
    - [x] blink
    - [x] adc
    - [x] usart  
    - [x] i2c
    - [x] spi
- [x] ravedude

PR #188, opened 2021-07-08
trinket-pro: add board based on arduino-uno

gauteh:
https://learn.adafruit.com/introducing-pro-trinket

Have not verified all pins yet.

I see you are working on a re-factor, so this might not be relevant.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ members = [
     "boards/bigavr6",
     "boards/sparkfun-pro-micro",
     "boards/trinket",
+    "boards/trinket-pro",
 ]
 
 [patch.crates-io]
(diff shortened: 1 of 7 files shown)


Rahix:
Hi, thanks for the PR!  However, as you noted, we are refactoring almost everything at the moment, so the changes you made do not really apply anymore.  If it is okay for you, please rework the changes for the `next` branch where the refactored codebase currently lives.  You can take a look at some of the previous board-support PRs, e.g. #177 to see how things work in the new architecture.  I know documentation on the `next` branch is still lacking so if you have any questions, feel free to ask!

gauteh:
:+1: #189 takes over.. the trinket-pro is very similar to the uno, so far it seems I can use the same pin spec.

PR #189, opened 2021-07-08
trinket-pro support

gauteh:
* adds support to ravedude for boards without serial console
* pins anyhow for ravedude on version that works on rust 2021-01-07
* adds blink example for trinket-pro

tested on 5V board, but think should work on 3V as well. Have not verified all pin names yet.

### DIFF
--- a/arduino-hal/Cargo.toml
+++ b/arduino-hal/Cargo.toml
@@ -15,6 +15,7 @@ arduino-leonardo = ["mcu-atmega", "atmega-hal/atmega32u4", "board-selected"]
 arduino-mega2560 = ["mcu-atmega", "atmega-hal/atmega2560", "board-selected"]
 arduino-nano = ["mcu-atmega", "atmega-hal/atmega328p", "atmega-hal/enable-extra-adc", "board-selected"]
 arduino-uno = ["mcu-atmega", "atmega-hal/atmega328p", "board-selected"]
+trinket-pro = ["arduino-uno"]
 sparkfun-promicro = ["mcu-atmega", "atmega-hal/atmega32u4", "board-selected"]
 
 [dependencies]
--- a/arduino-hal/Cargo.toml
+++ b/arduino-hal/Cargo.toml
@@ -15,7 +15,7 @@ arduino-leonardo = ["mcu-atmega", "atmega-hal/atmega32u4", "board-selected"]
 arduino-mega2560 = ["mcu-atmega", "atmega-hal/atmega2560", "board-selected"]
 arduino-nano = ["mcu-atmega", "atmega-hal/atmega328p", "atmega-hal/enable-extra-adc", "board-selected"]
 arduino-uno = ["mcu-atmega", "atmega-hal/atmega328p", "board-selected"]
-trinket-pro = ["arduino-uno"]
+trinket-pro = ["mcu-atmega", "atmega-hal/atmega328p", "board-selected"]
 sparkfun-promicro = ["mcu-atmega", "atmega-hal/atmega32u4", "board-selected"]
 
 [dependencies]
--- a/arduino-hal/src/lib.rs
+++ b/arduino-hal/src/lib.rs
@@ -12,6 +12,7 @@
 #![cfg_attr(feature = "arduino-nano", doc = "**Arduino Nano**.")]
 #![cfg_attr(feature = "arduino-uno", doc = "**Arduino Uno**.")]
 #![cfg_attr(feature = "sparkfun-promicro", doc = "**SparkFun ProMicro**.")]
+#![cfg_attr(feature = "trinket-pro", doc = "**Trinket Pro**.")]
 //! This means that only items which are available for this board are visible.  If you are using a
 //! different board, try building the documentation locally with
 //!
@@ -54,6 +55,7 @@ compile_error!(
     * arduino-nano
     * arduino-uno
     * sparkfun-promicro
+    * trinket-pro
     "
 );
 
--- a/arduino-hal/src/port/trinket_pro.rs
+++ b/arduino-hal/src/port/trinket_pro.rs
@@ -1,0 +1,119 @@
+pub use atmega_hal::port::mode;
+pub use atmega_hal::port::Pin;
+
+avr_hal_generic::renamed_pins! {
+    type Pin = Pin;
+
+    /// Pins of the **Trinket Pro**.
+    ///
+    /// This struct is best initialized via the [`arduino_hal::pins!()`][pins] macro.
+    pub struct Pins from atmega_hal::Pins {
+        /// `A0`
+        ///
+        /// * ADC0 (ADC input channel 0)
+        /// * PCINT8 (pin change interrupt 8)
+        pub a0: atmega_hal::port::PC0 = pc0,
+        /// `A1`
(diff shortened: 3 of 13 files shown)


gauteh:
The trinket-pro doesn't have a Serial-to-Usb, see slight refactor of Board in ravedude for how I solve it.

ravedude/Cargo.toml
```diff
@@ -14,7 +14,7 @@ categories = ["embedded", "hardware-support", "development-tools"]
 colored = "2.0.0"
 tempfile = "3.2.0"
 serialport = "4.0.0"
-anyhow = "1.0.38"
+anyhow = "=1.0.38"
```
Rahix, line 17: This shouldn't be neccessary.  Are you sure you're building `ravedude` with stable instead of the old nightly that is pinned for the AVR stuff?  You might need to use

```shell
cargo +stable build
```

ravedude/src/board.rs
```diff
@@ -1,10 +1,11 @@
 use crate::avrdude;
+use anyhow::Result;
```
Rahix, line 2: Please drop this import and refer to the type as `anyhow::Result<>` explicitly below.  This makes it easier for readers to understand where this result type is coming from.

ravedude/src/board.rs
```diff
@@ -31,12 +33,12 @@ fn find_port_from_vid_pid_list(list: &[(u16, u16)]) -> Option<std::path::PathBuf
...
             for (vid, pid) in list.iter() {
                 if usb_info.vid == *vid && usb_info.pid == *pid {
-                    return Some(port_name.into());
+                    return Ok(port_name.into());
                 }
             }
         }
     }
-    None
+    Err(anyhow!("Serial port not found."))
```
Rahix, line 41:
```suggestion
    Err(anyhow::anyhow!("Serial port not found."))
```

ravedude/src/board.rs
```diff
@@ -210,8 +212,8 @@ impl Board for ArduinoDiecimila {
         }
     }
 
-    fn guess_port(&self) -> Option<std::path::PathBuf> {
-        None
+    fn guess_port(&self) -> Option<Result<std::path::PathBuf>> {
+        Some(Err(anyhow!("Not able to guess port")))
```
Rahix, line 216:
```suggestion
        Some(Err(anyhow::anyhow!("Not able to guess port")))
```

ravedude/src/main.rs
```diff
@@ -2,6 +2,9 @@ use anyhow::Context as _;
 use colored::Colorize as _;
 use structopt::clap::AppSettings;
 
+#[macro_use]
+extern crate anyhow;
+
```
Rahix, line 7: Please drop this, `macro_use` and `extern crate` aren't really ideomatic for 2018 rust anymore.  You'll then need to refer to the macro as `anyhow::anyhow!()`, I've left comments above.

Rahix (changes requested):
Thank you very much for the PR.  I was actually about to do the same change to ravedude (for another board) so it's great timing that you took care of it just now :)  Looking at it, I think I'll need to do some bigger rework of the console logic later, but for now it looks like the best option.

Regarding the trinket-pro integration into `arduino-hal`, I'd like to see it as a completely separate board to avoid corner cases.  I've left a comment with details below.  Also, it would be nice if you could squash your changes into two commits:  one for the ravedude integration and one for the arduino-hal patch.

ravedude/Cargo.toml
Rahix: Hm, maybe we should add a `rust-toolchain.toml` to the `ravedude/` directory as well, to override its toolchain to `stable`?

gauteh: Not sure if adding another rust-toolchain is a good idea. People might want to use different versions of the compiler for whatever reason, that would be unnecessarily strict.

arduino-hal/Cargo.toml
```diff
@@ -15,6 +15,7 @@ arduino-leonardo = ["mcu-atmega", "atmega-hal/atmega32u4", "board-selected"]
 arduino-mega2560 = ["mcu-atmega", "atmega-hal/atmega2560", "board-selected"]
 arduino-nano = ["mcu-atmega", "atmega-hal/atmega328p", "atmega-hal/enable-extra-adc", "board-selected"]
 arduino-uno = ["mcu-atmega", "atmega-hal/atmega328p", "board-selected"]
+trinket-pro = ["arduino-uno"]
```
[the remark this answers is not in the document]
gauteh: Adding `#[cfg...]` inside the macro didn't work so ended up just making a separate ports file for the trinket.

ravedude/Cargo.toml
gauteh: Otherwise agree that pinning the dep doesn't make sense, didn't occur to me to use different compiler version for ravedude.

arduino-hal/Cargo.toml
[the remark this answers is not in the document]
Rahix: I see, probably means we should add proper `attr` handling to the pin juggling macro.  But let's not worry about that for now...

ravedude/Cargo.toml
Rahix: I'm most worried about people falling into this trap again.  If we can solve it for 80% of cases, that already means a lot less traffic in the issue tracker & other help channels.  For the rest, you can still overwrite the rust-toolchain with your own preferences anyway, `rustup` provides [plenty of mechanisms](https://rust-lang.github.io/rustup/overrides.html) to do that.

But let's keep this out of the trinket-pro PR, I can take care of that separately.

arduino-hal/src/lib.rs
```diff
@@ -12,6 +12,7 @@ compile_error!(
     * arduino-nano
     * arduino-uno
     * sparkfun-promicro
+    * trinket-pro
```
Rahix, line 15: I did some documentation changes today, please rebase on the latest version of `next` and also add `trinket-pro` to the doc-attr-mess above this:

```rust
#![cfg_attr(feature = "trinket-pro", doc = "**Trinket Pro**.")]
```

arduino-hal/src/port/trinket_pro.rs
```diff
@@ -0,0 +1,116 @@
+pub use atmega_hal::port::mode;
+pub use atmega_hal::port::Pin;
+
+avr_hal_generic::renamed_pins! {
+    type Pin = Pin;
+
```
Rahix, line 6: Also related to the documentation update...

```suggestion

    /// Pins of the **Trinket Pro**.
    ///
    /// This struct is best initialized via the [`arduino_hal::pins!()`][pins] macro.
```

Rahix:
BTW, once this PR is merged, you might also be interested in adding the `trinket-pro` to https://github.com/Rahix/avr-hal-template as well :)

gauteh:
Please take another look.

Rahix (approved):
Very nice, looks perfect.  I appreciate that you took the time to rebase your changeset into clean commits.  Thanks a lot!

gauteh:
Thanks, a pleasure to work with you.

lør. 10. jul. 2021, 10:11 skrev Rahix ***@***.***>:

PR #196, opened 2021-07-19
The  nano168 also supports serial

enaut:
I think it does make sense to enable serial for the atmega168.

### DIFF
--- a/arduino-hal/src/lib.rs
+++ b/arduino-hal/src/lib.rs
@@ -233,7 +233,7 @@ macro_rules! default_serial {
         )
     };
 }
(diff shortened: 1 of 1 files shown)


Rahix (changes requested):
I think you need to add this in line 237 instead, alongside `Arduino Nano`.  I wouldn't expect the Nano168 to suffer the baudrate issue that the other boards listed here have.

Rahix (approved):
Looks good, thanks a lot!  Will merge once CI passes.

enaut:
the serial works on the nano168 too in addition to the passing CI tests...

PR #197, opened 2021-07-19
Add RAVEDUDE_PORT definition to example

G4TGJ:
The uno-blink example does not run unless RAVEDUDE_PORT is defined.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,9 @@ Go into `./examples/arduino-uno` (or the directory for whatever board you want),
 ```bash
 cd examples/arduino-uno
 
+# Set the serial port connected to the Arduino
+export RAVEDUDE_PORT=/dev/ttyUSB1
+
 # Build and run it on a connected board
 cargo run --bin uno-blink
 ```


Rahix:
This really shouldn't be necessary - ravedude knows the Uno's VID:PID pairs and should be able to detect the device this way.  If it can't, either your device advertises incorrectly or the detection logic is broken.  Can you share more information about your board?

G4TGJ:
My Uno is a clone:

Bus 001 Device 011: ID 1a86:7523 QinHeng Electronics HL-340 USB-Serial adapter

So RAVEDUDE_PORT needs setting if you are using a clone?

Rahix:
> QinHeng Electronics HL-340 USB-Serial adapter

That seems to be a very odd clone indeed.  Usually, the Uno uses a second AVR MCU (ATmega16U2/8U2) for the USB-Serial but here an HL-340 is used.  Just as a heads-up, as this might cause problems with the USART clock generation - `avr-hal`, just like the official Arduino-Core, needs some workarounds in the baudrate calculation due to odd USART clocking on the ATmega16U2/8U2:

https://github.com/Rahix/avr-hal/blob/ac72adbe6db8900c41681b759ff62a11f2915916/avr-hal-generic/src/usart.rs#L114-L142

> So RAVEDUDE_PORT needs setting only if you are using a clone?

The VID:PID is not one associated with Arduino Uno:

https://github.com/Rahix/avr-hal/blob/ac72adbe6db8900c41681b759ff62a11f2915916/ravedude/src/board.rs#L66-L73

That's why `ravedude` can't find your board and you need to set the port manually.  As the HL-340 is a generic USB-Serial chip used in lots of other hardware as well, we cannot simply add it here either, I am afraid.

G4TGJ:
> here an HL-340 is used

That's why it's so cheap! I guess the error message is pretty clear how to set the port and as it only happens with some clones it's perhaps not worth worrying about. It hadn't even occurred to me that my Uno is not genuine which is why I thought there was a step missing from the instructions.

PR #198, opened 2021-07-19
Examples atmega168

enaut:
This depends on: #196 (the commit is also included here) - I made a separate pr because I'm not sure you want the examples ported.

I had to remove the temperature in the adc example - does the atmega168 not have the temperature functionality?

Thanks for your time I hope this helps some.

### DIFF
--- a/examples/nano168/src/bin/nano-adc.rs
+++ b/examples/nano168/src/bin/nano-adc.rs
@@ -1,0 +1,56 @@
+#![no_std]
+#![no_main]
+
+use arduino_hal::prelude::*;
+use panic_halt as _;
+
+use arduino_hal::adc;
+
+#[arduino_hal::entry]
+fn main() -> ! {
+    let dp = arduino_hal::Peripherals::take().unwrap();
+    let pins = arduino_hal::pins!(dp);
+    let mut serial = arduino_hal::default_serial!(dp, pins, 57600);
+
+    let mut adc = arduino_hal::Adc::new(dp.ADC, Default::default());
+
+    let (vbg, gnd) = (
+        adc.read_blocking(&adc::channel::Vbg),
+        adc.read_blocking(&adc::channel::Gnd),
+    );
+    ufmt::uwriteln!(&mut serial, "Vbandgap: {}", vbg).void_unwrap();
+    ufmt::uwriteln!(&mut serial, "Ground: {}", gnd).void_unwrap();
+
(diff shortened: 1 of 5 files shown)


Rahix:
> does the ATmega168 not have the temperature functionality?

It does not, according to the datasheet, and `atmega-hal` seems to know this (note how ATmga168 is missing here):

https://github.com/Rahix/avr-hal/blob/ac72adbe6db8900c41681b759ff62a11f2915916/mcu/atmega-hal/src/adc.rs#L102-L103

I would expect that the example shouldn't have even compiled for `nano168`?

examples/nano168/src/bin/nano-adc.rs
```diff
@@ -0,0 +1,56 @@
...
+            a3.analog_read(&mut adc),
+            a4.analog_read(&mut adc),
+            a5.analog_read(&mut adc),
+        ];
+
+        for (i, v) in values.iter().enumerate() {
+            ufmt::uwrite!(&mut serial, "A{}: {} ", i, v).void_unwrap();
+        }
+
+        // Arduino Nano has two more ADC pins A6 and A7.  Accessing them works a bit different from
```
Rahix, line 45:
```suggestion
        // Nano 168 has two more ADC pins A6 and A7.  Accessing them works a bit different from
```

Rahix (approved):
Thanks!  Looks good apart from one minor nit-pick.  Let's discuss the other commit in PR #196.  Once that's merged, please rebase this branch, then I'll merge it as well.

enaut:
> I would expect that the example shouldn't have even compiled for `nano168`?

It did not compile indeed - I was just wondering whether this is because of a disabled feature in the code or missing functionality in the chip. I checked the datasheets and came to the same conclusion as you but I did not trust my ability to read datasheets :) 1st timer here.

Rahix:
> disabled feature in the code or missing functionality in the chip.

Right, in this case it could have been both.  But actually the goal with avr-hal (and the wider embedded-rust ecosystem) is to build APIs where things not supported by the hardware do not compile in the first place :)  So that once it compiles, you can at least trust that hardware interaction is done correctly (save for logic bugs in your own code, of course).

---

Now that #196 is merged, please rebase this branch on `main` so I can merge it as well :)

enaut:
I'm battling git again...

now it should work - I also tested this version on the board.

Rahix (approved):
Looking great, thank you very much for your contribution! :)

PR #199, opened 2021-07-21
examples: nano168: Add more examples

enaut:
This PR contains:

  * more examples ported to the nano168 (basically copied over, renamed and tested)
  * renaming the existing nano168 examples
  * adding a light sensor example
  * Adding some more information about the individual examples available to the Readme

Feel free to request changes or pick only some of this PR.


examples/README.md
```diff
@@ -15,5 +15,52 @@ cd examples/arduino-uno
 cargo run --bin uno-blink
 ```
 
-You need to install [`ravedude`](https://crates.io/crates/ravedude) to make
+You need to install [`ravedude`](https://crates.io/crates/ravedude) with `cargo install ravedude` to make
 this work.
+
+The Examples
+------------
```
Rahix, line 22: I think this list is just prone to be outdated all the time... I would rather add a small blurb of information to the top of each example.  That said, I would suggest you just remove this text from the README for now and I'll go back later and sync all the examples.

Rahix (commented):
Hi, and sorry for letting this PR sit for so long.  Thanks for the additional examples!  Apart from the README changes, I like it very much - please drop those as explained below, then I'll merge it.

examples/README.md
enaut: I know what you mean! I assembled the list because I would have liked this kind of list for an overview when learning the basics. I'd also say that it doesn't matter if it is not up to date (that's why I did not add a list of which example is implemented for which board - that list would have been very hard to maintain).

Rahix: > would have liked this kind of list for an overview when learning the basics.

This is valuable feedback, thanks a lot.  I'll consider what we can do here...

Rahix (approved):

PR #200, opened 2021-07-22
generic: port: Implement `embedded-hal` OutputPin trait

Marcuss2:
Implements embedded_hal::digital::v2::OutputPin, I found out it was missing when I was trying to get hd44780 lcd working again after a while.


Rahix (approved):
Thanks!  I'll have to take care of proper GPIO `embedded-hal` support at some point, but if this is enough to help your case, I think it is a good starting point.

PR #203, opened 2021-07-28
generic: adc: Move mega-specific code into atmega-hal

quentinmit:
Part one of #202


Rahix (approved):
Looking great, thank you very much :)  The only thing I noticed is that the `ClockDivider` type is still part of `avr-hal-generic`, but I suppose that's okay until a device pops up with less or more supported settings.  So let's keep it as is.

PR #205, opened 2021-07-28
port: Add methods to temporarily switch a pin to a different mode

Rahix:
Cc: @quentinmit

This allows temporarily making an output pin an input and vice versa.  Code still needs to be cleaned and probably revised a bit.  To use:

```rust
let mut pin = pins.d0.into_output();

pin.with_pin_as_floating_input(|p| {
    // p is input here
    p.is_high();
});
```

Similarly:

- `with_pin_as_pull_up_input()`
- `with_pin_as_output()`
- `with_pin_as_output_high()`

### DIFF
--- a/avr-hal-generic/src/port.rs
+++ b/avr-hal-generic/src/port.rs
@@ -182,6 +182,91 @@ impl<PIN: PinOps, MODE: mode::Io> Pin<MODE, PIN> {
     }
 }
 
+
+impl<PIN: PinOps> Pin<mode::Output, PIN> {
+    /// Temporarily put this pin into input mode (with pull-up enabled) and revert back to output afterwards.
+    pub fn with_pin_as_pull_up_input<F, R>(&mut self, f: F) -> R
+    where
+        F: FnOnce(&mut Pin<mode::Input<mode::PullUp>, PIN>) -> R,
+    {
+        // Save state so we can restore it afterwards
+        let state = self.is_set_high();
+        // To "move" the pin out of current "owner" variable without actually moving
+        // SAFETY: This is okay because the pins are trivial structs.
+        let pin: Self = unsafe { core::ptr::read(self) };
+        let mut pin = pin.into_pull_up_input();
+        let res = f(&mut pin);
+        *self = if state {
+            pin.into_output_high()
+        } else {
+            pin.into_output()
+        };
+        res
+    }
+
+    /// Temporarily put this pin into input mode (floating) and revert back to output afterwards.
+    pub fn with_pin_as_floating_input<F, R>(&mut self, f: F) -> R
+    where
+        F: FnOnce(&mut Pin<mode::Input<mode::Floating>, PIN>) -> R,
+    {
+        // Save state so we can restore it afterwards
+        let state = self.is_set_high();
+        // To "move" the pin out of current "owner" variable without actually moving
+        // SAFETY: This is okay because the pins are trivial structs.
+        let pin: Self = unsafe { core::ptr::read(self) };
+        let mut pin = pin.into_floating_input();
+        let res = f(&mut pin);
+        *self = if state {
+            pin.into_output_high()
+        } else {
+            pin.into_output()
+        };
+        res
+    }
+}
+
+impl<PIN: PinOps, IMODE: mode::InputMode> Pin<mode::Input<IMODE>, PIN> {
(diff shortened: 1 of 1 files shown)


Rahix:
FWIW, other HALs have come up with a very similar API.  For example: https://docs.rs/stm32f4xx-hal/0.11.1/stm32f4xx_hal/gpio/struct.Pin.html#method.with_floating_input

mutantbob:
I wonder how this would behave if the pin were currently being used as the CS for an SPI transaction.

Rahix:
Well, the CS pin must be managed manually anyway so you can do weird stuff in any case... I don't think this changeset here would make it worse than it already is, or am I missing something?

haennes:
Hello, 
I really need this for my project.
Is there any way I can help?
Should I fork your fork and then open a new PR?

Rahix:
@haennes, giving me some feedback and testing this PR here is all I need.  I just held off merging it because I wanted to hear from users first.  Maybe you can comment on some of the open questions from the OP?  What would be best for your usecase?

haennes:
It works on an Arduino Nano (from Elegoo) (Atmega m328p).
I will test it on an Arduino Uno R3 (from Elegoo) (Atmega 328p) too.
Although the 2nd implementations seems better at a first glance, the way it currently is is quite perfect, as it removes a lot of boilerplate code in the users project.

avr-hal-generic/src/port.rs
```diff
@@ -215,6 +215,91 @@ impl<PIN: PinOps, MODE: mode::Io> Pin<MODE, PIN> {
...
+        // To "move" the pin out of current "owner" variable without actually moving
+        // SAFETY: This is okay because the pins are trivial structs.
+        let pin: Self = unsafe { core::ptr::read(self) };
+        let mut pin = pin.into_pull_up_input();
+        let res = f(&mut pin);
+        *self = if state {
+            pin.into_output_high()
+        } else {
+            pin.into_output()
+        };
```
Rahix, line 236: This implementation is still quite ugly and not 100% sound either.  I quite like the approach in `stm32f4xx-hal`: https://docs.rs/stm32f4xx-hal/0.11.1/src/stm32f4xx_hal/gpio/convert.rs.html#476-490  

This is at least more sound than my hack.  So I'll try to reimplement that pattern here.

Rahix: Source: https://github.com/stm32-rs/stm32h7xx-hal/blob/21d9c06a57169e92cf5745e582dca171c73294c9/src/gpio/convert.rs

It looks like this will unfortunately require a larger change to make it work.  But I think following that route is still a good idea and will improve the remainder of the `port` as well.  Will just need a bit more time...

Rahix:
@hiszd, not really.  `IoPin` moves the pin to change its state more permanently while the methods in this PR only change the pin state for the duration of a closure.

RenjiSann:
Hi,

I just finished implementing a driver for the DHT11 sensor on an Elegoo Arduino Uno R3.
I had to use a pull up pin with temporary output mode to make it work

jaqx0r:
With regards to the final bullet in the first comment, I think the user should not be responsible for returning a valid pin from the closure. We should assume the `with*` function takes care of maintaining state and correctness. I think that also makes it easier to return a pin's value from the closure in the case the pin was made input.

Is there a difference between changing mode with `into_output_high()` and calling `set_high()` on a low output? I think there isn't so a user can change state if they need after calling this closure with no return control.

Rahix:
> With regards to the final bullet in the first comment, I think the user should not be responsible for returning a valid pin from the closure. We should assume the with* function takes care of maintaining state and correctness. I think that also makes it easier to return a pin's value from the closure in the case the pin was made input.

If we switch to the design from stm32-hal I linked above, this can be done.  The pin is only passed to the closure by (mutable) reference so there is no way the closure can move the pin to the outside world and break invariants.

I agree that a user-controller closure return type makes for a much nicer API.

> Is there a difference between changing mode with into_output_high() and calling set_high() on a low output? I think there isn't so a user can change state if they need after calling this closure with no return control.

Yes, there is a difference.  Calling `into_output()` and then `set_high()` leads to a brief glitch on the pin where it's state is LOW for a short time.  In some applications, this can trip up external components.  `into_output_high()` switches from high-impedance input mode to HIGH output without any glitches.

PR #207, opened 2021-08-01
ravedude: Add autoreset logic for Arduino Leonardo

wt:
The arduino leonardo uses a funny method for autoreset. Whenever
you connect to the usb driven serial port at 1200 baud, it resets.
I have implemented this logic. If it can't find a device to
connect to, then it will fall back to asking you to manually
reset the device.

This protocol is not yet implemented in avr-hal, so this will
only work with loading firmware on a device that was flashed with
the arduino tools. Until avr-hal implements this protocol for
the leonardo, you will have to continue to manually reset if
during programming. However, I think that it is useful to
implement this side of the protocol as a motivation for getting
the USB CDC serial device working on leonardo.

I tested it numerous times, and it seems to be reliable when I
flash a device previously flashed with the arduino tools.

### DIFF
--- a/ravedude/src/board.rs
+++ b/ravedude/src/board.rs
@@ -138,7 +138,19 @@ impl Board for ArduinoLeonardo {
     }
 
     fn needs_reset(&self) -> Option<&str> {
-        Some("Reset the board by pressing the reset button once.")
+        let a = self.guess_port();
+        match a {
+            Some(Ok(name)) => {
+                match serialport::new(name.to_str().unwrap(), 1200).open() {
+                    Ok(_) => {
+                        std::thread::sleep(core::time::Duration::from_secs(1));
+                        None
+                    },
+                    Err(_) => Some("Reset the board by pressing the reset button once.")
+                }
+            },
+            _ => Some("Reset the board by pressing the reset button once.")
+        }
     }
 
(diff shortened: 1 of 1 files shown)


wt:
FWIW, I think that the port discovery is pretty ugly and am wondering if there is a better way to plumb that in so that the port detection is only done once. Thanks!

Rahix:
> as a motivation for getting the USB CDC serial device working on leonardo.

The problem here is mainly that the "right" way to implement USB support is currently blocked on a number of compiler problems.  Until those are resolved, we cannot implement a USB device driver without going completely off the best-practice path in `rust-embedded`.  If someone wants to write a custom app-specific one just for CDC-ACM, I'll be open to accept it into `avr-hal` but I will not work on such a solution myself for the time being.

Rahix (approved):
Thanks for the PR!  As you noted, this isn't all that useful at this point, but I don't see an issue with including it anyway.

> FWIW, I think that the port discovery is pretty ugly and am wondering if there is a better way to plumb that in so that the port detection is only done once. Thanks!

`board.rs` is supposed to go soon anyway, and get refactored into a sane solution which does not mix code and data as badly.  So I think the best way forward is to tidy this up alongside the refactoring work and not worry about it now.

wt:
Rock on. Thanks.

PR #208, opened 2021-08-01
Use the mode constant from the embedded hal.

wt:
The setting in the avr-generic-hal match the MODE_1
constant in the embedded hal spi implementation. Use
that instead of duplicating it.


Rahix (approved):
Thanks!

PR #209, opened 2021-08-02
trinket: erase chip on upload

explicite:
After few tests I spotted that chip should be erased for Trinket. Still I'm getting 
```
avrdude: verifying ...
avrdude: verification error, first mismatch at byte 0x0000
         0x00 != 0x0e
avrdude: verification error; content mismatch

avrdude: safemode: Fuses OK (E:00, H:00, L:00)

avrdude done.  Thank you.
```
Blink example is uploaded with success - tested for few version. @Rahix can you please review? Maybe after first load bootloader was corrupted.

### DIFF
--- a/ravedude/src/board.rs
+++ b/ravedude/src/board.rs
@@ -301,7 +301,7 @@ impl Board for Trinket {
             programmer: "usbtiny",
             partno: "attiny85",
             baudrate: None,
-            do_chip_erase: false,
+            do_chip_erase: true,
         }
     }
 


Rahix:
Nice, thanks a lot.  To clarify, when you say

> Still I'm getting [...]

Is this _with_ `do_chip_erase: true` or without?

explicite:
with `do_chip_erase: true`. will try to fix bootloader today and review

Rahix:
I also found an [Adafruit document about programming the trinket](https://learn.adafruit.com/introducing-trinket/programming-with-avrdude) a while ago, maybe check what they recommend as well.

Rahix (approved):
Cannot check right now but I'll trust that you tested this :)  Thanks again!

PR #215, opened 2021-08-18
Implement InputPin from embedded-hal

bootstrap-prime:
Quick implementation of the embedded-hal InputPin trait so that avr-hal can be used with bitbang-hal for a software-serial like experience.

(I'm not actually sure how to implement a timer, though, so I haven't gotten anything working. I'm also not sure where to begin.)

### DIFF
--- a/avr-hal-generic/src/port.rs
+++ b/avr-hal-generic/src/port.rs
@@ -3,7 +3,7 @@
 //! Please take a look at the documentation for [`Pin`] for a detailed explanation.
 
 use core::marker::PhantomData;
-use embedded_hal::digital::v2::OutputPin;
+use embedded_hal::digital::v2::{OutputPin, InputPin};
 
 pub trait PinMode: crate::Sealed {}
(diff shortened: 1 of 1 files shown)


Rahix (approved):
Looks good, thanks a lot!

Rahix:
> (I'm not actually sure how to implement a timer, though, so I haven't gotten anything working. I'm also not sure where to begin.)

Yeah, that's a bigger task... In the meantime, maybe the [`Delay`](https://rahix.github.io/avr-hal/arduino_hal/type.Delay.html) type is sufficient for your needs?

bootstrap-prime:
No unfortunately, bitbang_hal consumes a Timer object specifically, and it looks like Delay is a busy-loop, not timer based, which means I can't just steal it.
I have a scaffold type compiling, and I have some resources on timers that I'm struggling through, so hopefully a PR with a Timer will follow soon (but it won't be very good and may have to be implemented chip specific).

PR #217, opened 2021-08-31
Open Drain Output

knoby:
After the redesign of the crate you can't use the open drain output as discussed in #16 .

Perhaps this also solves the point " Re-implement tri-state mode in the new port API" in #130.

### DIFF
--- a/avr-hal-generic/src/port.rs
+++ b/avr-hal-generic/src/port.rs
@@ -3,7 +3,7 @@
 //! Please take a look at the documentation for [`Pin`] for a detailed explanation.
 
 use core::marker::PhantomData;
-use embedded_hal::digital::v2::{OutputPin, InputPin};
+use embedded_hal::digital::v2::{InputPin, OutputPin};
 
 pub trait PinMode: crate::Sealed {}
 /// GPIO pin modes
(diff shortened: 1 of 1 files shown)


Rahix:
Hi, thanks a lot for the PR.

I think the previous "tristate" mode was a bit misnamed, because it wasn't a full tristate implementation.  I think what we actually want is two different modes:

#### 1. `OutputOpenDrain`
Open-Drain should behave exactly as open-drain outputs on other platforms do.  That means, strictly being an output where `set_high()` puts the pin into HIGHZ and `set_low()` pulls it to GND.  This mode should just implement `OutputPin` and `StatefulOutputPin`.

#### 2. `Tristate`
The tristate mode should be a flexible mode where there are actually 3 states and thus also 3 methods for setting the state.  Those are:

- `set_high()` for actually setting the pin HIGH (i.e. pull it to VCC).
- `set_tristate()` for putting the pin into HIGHZ/tristate mode.
- `set_low()` for pulling to GND.

Pins in this mode should now implement `OutputPin` (using the real `set_high()` and `set_low()`) and `InputPin`, but _not_ `StatefulOutputPin` because the pin can be in a state which cannot be represented by that trait.

Still, outside the e-h traits, we should provide `is_set_high()`, `is_set_low()`, and `is_set_tristate()` methods for checking the current state.  Importantly, these should reflect the actual state _set_, not the electrical value read from the pin.  This means that even if an external pull-down is connected, after `set_tristate()`, `is_set_tristate()` should return true (but `is_low()` is also true, of course).

---

What do you think?

Rahix:
@nikklassen, if the author of this PR doesn't respond, feel free to send a new PR superseding this one.  Especially if you already continued work in this direction in your fork!

Rahix:
Superseeded by #263.  Thanks for your initial work on this @knoby!

Issue #211, opened 2021-08-09
Using i2c to PWM 16 servo x 12bit board, how would I send a signal to a servo?


joshuajbouw:
Hey there, this is a general question. I also understand that this is a fairly low-level library which I am completely fine with. However, I am struggling to find good information on exactly what the data packet should be to be able to control a servo on board 0x40. If you can point me in the right direction, I'll be more than grateful.

Rahix:
You need to provide more information, I cannot help you from this.  What do you mean by "Using i2c to PWM 16 servo x 12bit board"?  Do you have an I2C PWM port-expander which you're using to control a servo?  Please provide a description of the circuit (even better: schematic), and exact part numbers for the components involved.  Otherwise there is no way I can know what you need...

joshuajbouw:
Sure.

# Hardware
- Arduino Uno
- [Adafruit 16-Channel 12-bit PWM/Servo Driver](https://www.adafruit.com/product/815)
- [Tower Pro MG995](https://www.electronicoscaldas.com/datasheet/MG995_Tower-Pro.pdf)

# Connections
- Arduino Uno connected to PWM.
    - Arduino `A4` connected to PWM `SDA`
    - Arduino `A5` connected to PWM `SCL`
    - Arduino `5V` connected to PWM `VCC`
    - Arduino `GND` connected to PWM `GND`
    
 I had followed the `i2c` [example](https://github.com/Rahix/avr-hal/blob/main/examples/arduino-uno/src/bin/uno-i2cdetect.rs) for Arduino Uno and the detection is successful. However, I only see that I can send bytes to a board at an address. In my case 0x40. What does that data need to look like?

Rahix:
Thanks.  It looks like that board of yours uses a PCA9685 "16-channel, 12-bit PWM Fm+ I2C-bus LED controller" ([datasheet](https://cdn-shop.adafruit.com/datasheets/PCA9685.pdf)).  You'll need to correctly write to its registers to archieve your desired goal.  There are three routes you can take here:

1. You're in luck and there already is an embedded-rust driver for this chip: [`pwm-pca9685`](https://crates.io/crates/pwm-pca9685). You might be able to just hook that up to the i2c driver from avr-hal and then use its API to control the generated PWM signal.  Here is an example I found, which might help: [`pca9685-servos-f3.rs`](https://github.com/eldruin/driver-examples/blob/master/stm32f3-discovery/examples/pca9685-servos-f3.rs).
[...]

PR #218, opened 2021-09-07
ravedude documentation

michelepagot:
Mention in the ravedude documentation what is a suitable avrdude versioin to be used (6.3)

### DIFF
--- a/ravedude/README.md
+++ b/ravedude/README.md
@@ -1,7 +1,7 @@
 ravedude [![crates.io page](https://img.shields.io/crates/v/ravedude.svg)](https://crates.io/crates/ravedude)
 ========
 `ravedude` is a CLI utility to make Rust development for AVR microcontrollers
-super smooth.  It's a wrapper around `avrdude` and provides easy access to the
+super smooth.  It's a wrapper around `avrdude` (tested with avrdude version 6.3) and provides easy access to the
 target's serial console, similar to the Arduino IDE.
 
 `ravedude` is meant to be used as a cargo "runner".  This allows you to just use


Rahix:
Hi, thanks for the PR.

Instead of just documenting this in some README, I think it would be much better to have `ravedude` actually verify what version of `avrdude` is installed by calling `avrdude -?` and parsing the version string.  If the version is unsupported, it should bail out instead of letting people run into the issues.  This is going reach people much more likely than if we just have a small note in the README.

Would you be interested in implementing that?

luukvanderduim:
> 
> Would you be interested in implementing that?

I would!  If you and @michelepagot don't object, that is. ;)
Is there a minimum version that we know of or is the idea to declare 6.3> because it is known to work fine?
Thanks in advance!

Rahix:
@luukvanderduim, that's awesome, please go for it!

>  Is there a minimum version that we know of or is the idea to declare 6.3> because it is known to work fine?

Yeah, let's not try to deal with older versions, it's enough of a mess as it stands.  avrdude 6.3 was released 2016 so I'd say there is no excuse for using older version in 2021.

Thanks for picking this up! :)

stappersg:
On Mon, Sep 13, 2021 at 08:05:08AM -0700, Luuk van der Duim wrote:
> > 
> > Would you be interested in implementing that?
> 
> I would!  If you and @michelepagot don't object, that is. ;)

Sent patches     :-)


> Is there a minimum version that we know of
> or is the idea to declare 6.3> because it is known to work fine?

Put minimum version in a variable, constant or other data structure
that enables future maintenance. Yes, sent patches ...


Regards Stappers

P.S.
It is not important what my opinion is on ravedude checking
which avrdude version is being used.

Important is there is software with 100 freedoms
001 freedom to use
010 freedom to share
011 freedom to study
100 freedom to modify

Rahix:
Closing as this is now superseded by #211.  Thanks @michelepagot for kicking off this topic.

luukvanderduim:
> Closing as this is now superseded by #211. Thanks @michelepagot for kicking off this topic.

Is it possible you meant #221, or am I missing something?

Rahix:
Of course, sorry!

PR #221, opened 2021-09-14
Implement avrdude runtime version check.

luukvanderduim:
This is my initial idea on how to resolve the runtime version checking.
I welcome advice, suggestions and/or comments.

May resolve issue #218

- CSpell found a few spelling errors.
- Introduces a version const in main.
- Adds 'require_min_ver' method to the Avrdude impl.

### DIFF
--- a/avr-hal-generic/src/adc.rs
+++ b/avr-hal-generic/src/adc.rs
@@ -1,4 +1,4 @@
-/// Analog-to-Digital converter
+/// Analog-to-Digial converter
 use core::marker::PhantomData;
 
 /// The division factor between the system clock frequency and the input clock to the AD converter.
--- a/avr-hal-generic/src/adc.rs
+++ b/avr-hal-generic/src/adc.rs
@@ -1,4 +1,4 @@
-/// Analog-to-Digial converter
+/// Analog-to-Digital converter
 use core::marker::PhantomData;
 
 /// The division factor between the system clock frequency and the input clock to the AD converter.
--- a/ravedude/src/avrdude/mod.rs
+++ b/ravedude/src/avrdude/mod.rs
@@ -17,14 +17,14 @@ pub struct Avrdude {
 }
 
 impl Avrdude {
-    pub fn require_min_ver(ver: (u8, u8)) -> anyhow::Result<()> {
+    pub fn require_min_ver((req_major, req_minor): (u8, u8)) -> anyhow::Result<()> {
         let output = process::Command::new("avrdude")
             .arg("-?")
             .output()
             .context("Failed to start and gather avrdude output.")?;
-        let stdout: &str =
-            std::str::from_utf8(&output.stdout).context("Avrdude's stdout contains non-utf8.")?;
-        let version: &str = stdout
+        let stderr: &str =
+            std::str::from_utf8(&output.stderr).context("Avrdude's stderr contains non-utf8.")?;
+        let version: &str = stderr
             .split("version")
             .last()
             .context("Unable to obtain avrdude's version string")?
@@ -40,22 +40,22 @@ impl Avrdude {
         let minor = minor
             .parse::<u8>()
             .context("Unable to parse minor version number.")?;
-        if major < ver.0 {
+        if major < req_major {
             bail!(
-                "Avrdude does not meet minimum version requirements ({}.{}), but {}.{} was found.",
-                ver.0,
-                ver.1,
+                "Avrdude does not meet minimum version requirements. v{}.{} was found while v{}.{} or greater is required.",
                 major,
-                minor
+                minor,
+                req_major,
+                req_minor,
             );
         }
-        if major == ver.0 && minor < ver.1 {
+        if major == req_major && minor < req_minor {
             bail!(
-                "Avrdude does not meet minimum version requirements ({}.{}), but {}.{} was found.",
-                ver.0,
-                ver.1,
+                "Avrdude does not meet minimum version requirements. v{}.{} was found while v{}.{} or greater is required.",
                 major,
-                minor
+                minor,
+                req_major,
+                req_minor,
             );
         }
         Ok(())
--- a/ravedude/src/avrdude/mod.rs
+++ b/ravedude/src/avrdude/mod.rs
@@ -38,7 +38,7 @@ impl Avrdude {
         if (major, minor) < (req_major, req_minor) {
             anyhow::bail!(
                 "Avrdude does not meet minimum version requirements. v{}.{} was found while v{}.{} or greater is required.\n\
-                You may find a more suitable version here: http://download.savannah.gnu.org/releases/avrdude/",
+                You may find a more suitable version here: https://download.savannah.gnu.org/releases/avrdude/",
                 major, minor,
                 req_major, req_minor,
             );
--- a/ravedude/src/avrdude/mod.rs
+++ b/ravedude/src/avrdude/mod.rs
@@ -1,4 +1,4 @@
-use anyhow::Context as _;
+use anyhow::{bail, Context as _};
 use std::path;
 use std::process;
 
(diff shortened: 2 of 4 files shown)


luukvanderduim:
It its morning.. and I see I have some things to fix. Sorry, please hold. ;)

luukvanderduim:
It appears avrdude sends help messages to stderr, which is unusual:
`avrdude -? > /dev/null`
Fixed expectations.

ravedude/src/avrdude/mod.rs
```diff
@@ -17,6 +17,50 @@ pub struct Avrdude {
 }
 
 impl Avrdude {
+    pub fn require_min_ver((req_major, req_minor): (u8, u8)) -> anyhow::Result<()> {
```
Rahix, line 20: I think you can make this function more manageable by splitting it into two parts:  The parsing part and the reporting part.  Notice how I dropped all the `.context()` calls in favor of a single one when invoking the `get_avrdude_version()` function.  A user won't care about the exact failure, we'll just need to tell them that _something_ didn't work.  They can piece together the rest from the error trace.

```rust
    fn get_avrdude_version() -> anyhow::Result<(u8, u8)> {
        let output = process::Command::new("avrdude").arg("-?").output()?;
        let stderr: &str = std::str::from_utf8(&output.stderr)?;
        let version: &str = stderr.split("version").last()?.trim();
        let (version, _) = version.split_once('-')?.to_owned();
        let (major, minor) = version.split_once('.').unwrap();
        let major = major.parse::<u8>()?;
        let minor = minor.parse::<u8>()?;
        (major, minor)
    }

    pub fn require_min_ver((req_major, req_minor): (u8, u8)) -> anyhow::Result<()> {
        let (major, minor) = Self::get_avrdude_version()
            .context("Failed reading avrdude version information")?;
        if major < req_major {
            // ...
        }
        Ok(())
    }
```

ravedude/src/avrdude/mod.rs
```diff
@@ -17,6 +17,50 @@ pub struct Avrdude {
...
+            .context("Unable to obtain avrdude's version string")?
+            .to_owned();
+        let (major, minor) = version.split_once('.').unwrap();
+        let major = major
+            .parse::<u8>()
+            .context("Unable to parse major version number.")?;
+        let minor = minor
+            .parse::<u8>()
+            .context("Unable to parse minor version number.")?;
+        if major < req_major {
```
Rahix, line 43: Actually, you can combine the two `if`s because tuples in Rust implement comparison operations.  See this demonstration: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=836cd2133053e4b2e547af94a8fe8bda

ravedude/src/avrdude/mod.rs
```diff
@@ -17,6 +17,50 @@ pub struct Avrdude {
...
+            bail!(
+                "Avrdude does not meet minimum version requirements. v{}.{} was found while v{}.{} or greater is required.",
```
Rahix, line 45: Just as an idea: Can we point to the official download page for avrdude in this error message?  Just so people are immediately guided in the right direction?  Especially for the Windows users, not having them download avrdude from the first link off Google might be good.

Rahix (changes requested):
Thank you very much for taking the time to implement this!  I've left some comments below :)

avr-hal-generic/src/adc.rs
```diff
@@ -1,4 +1,4 @@
-/// Analog-to-Digial converter
+/// Analog-to-Digital converter
```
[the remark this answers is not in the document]
luukvanderduim: I am perfectly fine with opening a new PR for these spell-check fixes. I'll pull master and see which are left on a later moment.

ravedude/src/avrdude/mod.rs
luukvanderduim: Yes, much easier on the eye. And you are right, while the context messages were helpful during implementing and testing, in the end, the user won't care that much.

ravedude/src/avrdude/mod.rs
luukvanderduim: Really? Oh wow. Rust keeps on giving! ;)

ravedude/src/avrdude/mod.rs
luukvanderduim: Sure, no problem!

ravedude/src/main.rs
```diff
@@ -71,6 +74,7 @@ fn main() {
 }
 
 fn ravedude() -> anyhow::Result<()> {
+    avrdude::Avrdude::require_min_ver(MIN_VERSION_AVRDUDE)?;
```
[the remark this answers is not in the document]
luukvanderduim: Yes of course. Thanks!

ravedude/src/avrdude/mod.rs
luukvanderduim: Thank you. I have found your suggestions and guidance very helpful and appreciate you took the time.
Just pushed an new rendition of the PR hopefully addressing most concerns. 
I am not sure when to resolve a conversation. Intuitively I'd think 'when the outcome is satisfactory'?

luukvanderduim: Sorry, it seems I botched. I tried to revert 'digital' spelling fix, but it seems I changed two instances. I cannot find the simple way of reverting a single hunk in a patch, so I did it manually. Sigh.

luukvanderduim: I did and I fixed it.

Rahix (commented):
> Thank you. I have found your suggestions and guidance very helpful and appreciate you took the time.

Happy to hear that :)  That's what code-review is meant to be.

> Just pushed an new rendition of the PR hopefully addressing most concerns.

The implementation looks great, however during testing on my machine, I couldn't get it to work.  I think your avrdude version might have slightly differing output?  Here's what it looks like for me:

```text
❯ avrdude -?
Usage: avrdude [options]
Options:
  -p <partno>                Required. Specify AVR device.
  [...]
  -?                         Display this usage.

avrdude version 6.3, URL: <http://savannah.nongnu.org/projects/avrdude/>
```

Your code splits on a `-` in line 25 which is not present in my output as you can see... Can you share your `-?` message so we can compare?

> I am not sure when to resolve a conversation. Intuitively I'd think 'when the outcome is satisfactory'?

Honestly, I am unsure as well.  GitHub allows both the contributor and the reviewer to resolve conversations, so there isn't a clear responsibility.  I'd just say conversations should be resolved when a comment is "addressed" and it is up to the reviewer to recheck whether it was addressed appropriately and to open a new conversation if not... But that might be just me...

ravedude/src/avrdude/mod.rs
luukvanderduim: In `get_avrdude_version()` I had to introduce `ok_or_else(err)` because `last()` yields an `Option` while we need a `Result`.

ravedude/src/main.rs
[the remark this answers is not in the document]
luukvanderduim: Tested it with a mock avrdude with 'too old version'. Calling `ravedude --help` seems to work fine now.

luukvanderduim:
> > Just pushed an new rendition of the PR hopefully addressing most concerns.
> 
> The implementation looks great, however during testing on my machine, I couldn't get it to work. I think your avrdude version might have slightly differing output? Here's what it looks like for me:
> 
> ```
> ❯ avrdude -?
> Usage: avrdude [options]
> Options:
>   -p <partno>                Required. Specify AVR device.
>   [...]
>   -?                         Display this usage.
> 
> avrdude version 6.3, URL: <http://savannah.nongnu.org/projects/avrdude/>
> ```
> 
> Your code splits on a `-` in line 25 which is not present in my output as you can see... Can you share your `-?` message so we can compare?

I was afraid we would find these kind of differences. But good to find them now. I have an idea to make it more robust.

Mine looks like this: 
``` Bash
$ avrdude -?
Usage: avrdude [options]
Options:
  -p <partno>                Required. Specify AVR device.
 [...]

avrdude version 6.3-20171130, URL: <http://savannah.nongnu.org/projects/avrdude/>

```
I suspect this is an Ubuntuism.

Instead of looking for '-', I can look for the first non-numeric after '.'.

> 
> > I am not sure when to resolve a conversation. Intuitively I'd think 'when the outcome is satisfactory'?
> 
> Honestly, I am unsure as well. GitHub allows both the contributor and the reviewer to resolve conversations, so there isn't a clear responsibility. I'd just say conversations should be resolved when a comment is "addressed" and it is up to the reviewer to recheck whether it was addressed appropriately and to open a new conversation if not... But that might be just me...

Thanks!

luukvanderduim:
> > Your code splits on a `-` in line 25 which is not present in my output as you can see... Can you share your `-?` message so we can compare?

 What I _really_ want to do is something like this:
``` Rust
if let Some(major, minor): (u8, u8) = stderr[.., "version ", major@0_u8..=255_u8,".", minor@0_u8..=255_u8, ..].match_magic() {
    (major, minor) } 
else { 
    anyhow!("could not match") 
}
```
Looking for pattern magic..

luukvanderduim:
> Looking for pattern magic..
Magic _that fancy_ does not seem to exist.
``` Rust
        let version: String = version
            .chars()
            .take_while(|c| c.is_ascii_digit() || *c == '.')
            .collect();
```
Looks reasonably robust for now.

ravedude/src/avrdude/mod.rs
```diff
@@ -17,6 +17,35 @@ pub struct Avrdude {
...
+        Ok((major, minor))
+    }
+
+    pub fn require_min_ver((req_major, req_minor): (u8, u8)) -> anyhow::Result<()> {
+        let (major, minor) =
+            Self::get_avrdude_version().context("Failed reading avrdude version information.")?;
+        if (major, minor) < (req_major, req_minor) {
+            anyhow::bail!(
+                "Avrdude does not meet minimum version requirements. v{}.{} was found while v{}.{} or greater is required.\n\
+                You may find a more suitable version here: http://download.savannah.gnu.org/releases/avrdude/",
```
Rahix, line 41: One minor nit, please make this an `https` link so people don't go through an unsecured connection if their browser isn't strict about encryption.

```suggestion
                You may find a more suitable version here: https://download.savannah.gnu.org/releases/avrdude/",
```

Rahix (approved):
Alright, retested and now it works flawlessly :)  I have one more minor comment, please apply my suggestion and then I will merge your PR.  Thank you very much for working on this!

luukvanderduim:
> Alright, retested and now it works flawlessly :) I have one more minor comment, please apply my suggestion and then I will merge your PR. Thank you very much for working on this!

My pleasure. If there is anything else you think I could do next, please nudge me! ;)

PR #233, opened 2022-01-05
README: Add a note about necessary dependencies

futurechimp:
I ran into trouble running the examples as I didn't realize from the README that it was necessary to install a few things from `apt` before running the Rust code. In case it's useful I've documented them here.


Rahix (approved):
Good point, thanks!

PR #234, opened 2022-01-10
Added one pair of vid_pid for Arduino Nano

Fatalon:
Hey, thank you for your great Arduino libraries.
I wanted to try out rust for embedded on one of my Arduino Nano's and noticed, that the `guess_port` method failed for my device due to missing vid,pid pairs for the arduino nano via:
``` 
Error: no matching serial port found, use -P or set RAVEDUDE_PORT in your environment

Caused by: Not able to guess port
```
I added at least my pair and added a sentence about this failure inside `Readme`.

```
[34817.652918] usb 1-4: New USB device found, idVendor=1a86, idProduct=7523, bcdDevice= 2.64
[34817.652927] usb 1-4: New USB device strings: Mfr=0, Product=2, SerialNumber=0
[34817.652932] usb 1-4: Product: USB Serial
[34817.654680] ch341 1-4:1.0: ch341-uart converter detected
[34817.655335] usb 1-4: ch341-uart converter now attached to ttyUSB0

Bus 001 Device 009: ID 1a86:7523 QinHeng Electronics HL-340 USB-Serial adapter
```

### DIFF
--- a/ravedude/README.md
+++ b/ravedude/README.md
@@ -7,6 +7,9 @@ target's serial console, similar to the Arduino IDE.
 `ravedude` is meant to be used as a cargo "runner".  This allows you to just use
 `cargo run` for building, deploying, and running your AVR code!
 
+if you get an `Error: no matching serial port found, use -P or set RAVEDUDE_PORT in your environment` , 
+run `cargo run` with set environment variable or adjust `runner = "ravedude {X} -cb {X} -P /dev/ttyUSB{X}"` inside `.cargo/config.toml` (replace {X} with your respective values)
+
 ## Installation
 On Linux systems, you'll need pkg-config and libudev development files
 installed:
@@ -46,7 +49,7 @@ Reading | ################################################## | 100% 0.00s
 
 avrdude: Device signature = 0x1e950f (probably m328p)
 avrdude: erasing chip
-avrdude: reading input file &quot;avr-hal/target/avr-atmega328p/debug/uno-i2cdetect.elf&quot;
+avrdude: reading input file &quot;avr-hal/target/avr-atmega328p/debug/uno-i2cdetect.elf&quot; 
 avrdude: writing flash (1654 bytes):
 
 Writing | ################################################## | 100% 0.27s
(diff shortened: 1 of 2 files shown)


Rahix:
Hi, thanks a lot for the PR.

Unfortunately, the story isn't quite as simple... CH341 is a generic USB-to-serial converter chip which is built into many devices.  The VID:PID pair here does not uniquely identify an Arduino Nano (clone) - it could be any other device using this chip as well.  Genuine Nano's have the same problem, they use a generic chip from FTDI whose VID:PID also does not uniquely identify the device.

In contrast, for example the Arduino Uno has a VID:PID (2341:0043) which is unique to this board - thus we can match on it.  For Arduino Nano, we don't have any such option available... Thus I would rather not include any port guessing for this board and prompt the user to select the correct port explicitly.  While it might be working for you right now, this could go wrong for people who have other CH341 based devices connected at the same time.

That said, the README change is not a bad idea!

Fatalon:
Hi Rahix, sorry for my late response.
Oh okay, no problem. I reverted my changes to the port_guessing such that you can include only the Readme change. 
I have no problem, with manually choosing the port, but I noticed, that it looks like the Platform IO project still decided to guess a port and let the user manually adjust the port, if a wrong port was chosen.
I just tried to flash a simple LED/C program and I didn't needed to select my device.
I added a link to a post from their forum for a use-case similar to mine, I think:
https://community.platformio.org/t/auto-detect-port-explained/16062/2
Just out of curiosity, is there a deeper motivation behind your decision?

Rahix:
> Just out of curiosity, is there a deeper motivation behind your decision?

Well, I am a huge opponent of "smart" software:  If there is an obvious path to choose, do it, but in face of ambiguity, I think software should never guess.  At least software targetting tech-savvy people, that is.

As I wrote above, these CH341 chips show up in all kinds of devices, so assuming that it is the right one feels wrong to me.

That said, I certainly agree that usability of `ravedude` in this regard is quite bad and some form of a solution is needed... My "plan" is to implement a prompt when devices were found but could not be identified.  The user can then select one of the devices and optionally have it written to a config file for next time.

Issue #237, opened 2022-01-22
External Interrupts


tsemczyszyn:
I'm new to embedded programming and rust. I have a rotary encoder and it seems I should be able to use external interrupt pins to tell when the encoder is being turned. I tried to refer to the timer interrupt example and adapt it to work in my use case but nothing I'm doing seems to work.

I thought I could do something like
```
#[avr_device::interrupt(atmega328p)]
fn INT0() {
       CLK_PIN_INT.store(true, atomic::Ordering::SeqCst);
}
```
to toggle a global flag and then read that in my main function but it doesn't seem to be working as I thought. I did enable the interrupts with `unsafe { avr_device::interrupt::enable() };`. I'm a bit stuck. Are there any examples out there for doing this type of thing?

stappersg:
On Fri, Jan 21, 2022 at 04:15:30PM -0800, Travis Semczyszyn wrote:
> Are there any examples out there for doing this type of thing?

Please tell how usefull #235 is.

tsemczyszyn:
Yes this has worked for me. There's a lot I don't understand but I have been able to get it working. Thanks.

stappersg:
On Sat, Jan 22, 2022 at 09:58:02AM -0800, Travis Semczyszyn wrote:
> There's a lot I don't understand

That is known as knowledge.  And knowledge is an island.
When it grows, the shore line of the unknown increases.  8^)

Rahix:
@tsemczyszyn, if you don't mind, I'd love a pull request with a small example so we have something for others who have the same problem in the future.

tsemczyszyn:
Sure I'll try and put something together.

PR #238, opened 2022-01-25
Add an example for pin change interrupts on uno

tsemczyszyn:
As requested in #237

### DIFF
--- a/examples/arduino-uno/src/bin/uno-pin-change-interrupt.rs
+++ b/examples/arduino-uno/src/bin/uno-pin-change-interrupt.rs
@@ -23,6 +23,17 @@ fn PCINT2() {
     PIN_CHANGED.store(true, Ordering::SeqCst);
 }
 
+fn rotate(flag: &AtomicBool) -> bool {
+    avr_device::interrupt::free(|_cs| {
+        if flag.load(Ordering::SeqCst) {
+            flag.store(false, Ordering::SeqCst);
+            true
+        } else {
+            false
+        }
+    })
+}
+
 #[arduino_hal::entry]
 fn main() -> ! {
     let dp = arduino_hal::Peripherals::take().unwrap();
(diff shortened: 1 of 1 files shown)


examples/arduino-uno/src/bin/uno-pin-change-interrupt.rs
```diff
@@ -0,0 +1,72 @@
...
+
+static PIN_CHANGED: AtomicBool = AtomicBool::new(false);
+
+//This function is called on change of pin 2
+#[avr_device::interrupt(atmega328p)]
+#[allow(non_snake_case)]
+fn PCINT2() {
+	avr_device::interrupt::free(|_cs| {
+        PIN_CHANGED.store(true, Ordering::SeqCst);
+    });
```
Rahix, line 25: There is no need to have this critical section here.  For one, interrupts are already disabled when entering an ISR.  And secondly, as this variable is "atomic", no external synchronization is necessary to ensure safe mutation.

```suggestion
    PIN_CHANGED.store(true, Ordering::SeqCst);
```

(atomic really means "emulated atomic" here as AVR does not really have any atomic operations)

examples/arduino-uno/src/bin/uno-pin-change-interrupt.rs
```diff
@@ -0,0 +1,72 @@
...
+    
+    //Rotary encoder attached on these pins
+    let rotary_pins = [
+        pins.d2.into_floating_input().downgrade(), //CLK
+        pins.d3.into_floating_input().downgrade() //DT
+    ];
+
+    //Needed to enable pin change interrupts
+    dp.EXINT.pcicr.write(|w| unsafe { w.bits(0b100) });
+    dp.EXINT.pcmsk2.write(|w| unsafe { w.bits(0b100) });
```
Rahix, line 45: I think it would be good to have a comment here explaining those magic values.  E.g.

```rust
// Enable the PCINT2 pin change interrupt
dp.EXINT.pcicr.write(|w| unsafe { w.bits(0b100) });

// Enable pin change interrupts on PCINT18 which is pin PD2 (= d2)
dp.EXINT.pcmsk2.write(|w| unsafe { w.bits(0b100) });
```

Rahix (commented):
Thanks for the contribution!  I've left two comments below, please fix these so I can merge it :)

examples/arduino-uno/src/bin/uno-pin-change-interrupt.rs
```diff
@@ -0,0 +1,72 @@
...
+
+    //From this point on an interrupt can happen
+    unsafe { avr_device::interrupt::enable() };
+
+    loop {
+        avr_device::interrupt::free(|_cs| {
```
jeroenvlek, line 51: First of all, thanks for the inspiration! This critical section seems a bit long, especially since you're already using an AtomicBool?

tsemczyszyn: If I'm using the AtomicBool I wonder do I even need to have a critical section here? I'm coming to Rust from Python so there's a lot of new concepts and I'm a bit unsure about some of it.

jeroenvlek: A critical section is part of a bigger synchronization paradigm, like Barriers Mutexes etc. It's not just restricted to Rust. The upside of having a section where you know that your thread/process is the only one accessing a resource is that you know the ordering of your operations and their outcome. The downside is that all other threads/processes that need something from that resource have to wait. So for the overall performance of your program you want to make your time in the critical section as short as possible; take what you need and get out asap.

The AtomicBool is thread safe, but you are doing two subsequent calls to that struct on line 52 and 53. Between those two calls anything can happen, even though you expect the value to remain the same, so those need to be enclosed. However, the remainder of that if arm can be done outside of the CS, because at that point you already know you're going to execute that.

tsemczyszyn: Thanks for the explanation. I think this explains why I've seen another example where a separate function is used to read the AtomicBool, update it, and return the result. That would avoid blocking the interrupts unnecessarily.

Rahix:
Thanks!

PR #239, opened 2022-01-25
Run `cargo install cargo-generate`

stappersg:
Added
  cargo install cargo-generate
to the advice on project start.

In case `cargo-generate` is already present, will the user get
      Updating crates.io index
       Ignored package `cargo-generate v0.12.0` is already installed
and an
  use --force to override
hint.


Rahix (approved):
Seems like a good idea, thanks!

PR #244, opened 2022-02-02
Add KontroLIR board support

jkristell:
Add support for the AnalysIR KontroLIR remote control.

https://www.analysir.com/blog/product/kontrolir-arduino-compatible-ir-remote-control/


ravedude/src/avrdude/avrdude.conf
```diff
@@ -8683,6 +8683,14 @@ part parent "m328"
     ocdrev              = 1;
 ;
```
stappersg, line 8685: Here should be something like
```text
# https://www.analysir.com/blog/product/kontrolir-arduino-compatible-ir-remote-control/
```
to express it is an  ATmega328PB chip on an InfraRed-remote-controller.

Rahix: NAK, the avrdude config does not care what board an MCU is mounted on.  Take a look at the ATmega328P entry, for example, it also does not mention anything but the chip itself.

PR #259, opened 2022-04-22
make serial port example portable across targets

mutantbob:
extract the serial port example into a portable function that demonstrates how to declare the serial port using generics such that it will compile on multiple targets.

I think folks can benefit from seeing concrete examples of how to write cross-platform fuctions like

```
pub fn report<H, USART: UsartOps<H, RX, TX>, RX, TX, CLOCK>(
    serial: &mut Usart<H, USART, RX, TX, CLOCK>,
) {
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,8 @@ members = [
 exclude = [
     # The RAVEDUDE! Yeah!
     "ravedude",
+    # requires the user of this crate to specify the features for embedded-hal
+    "examples/arduino-portable",
 ]
 
 [patch.crates-io]
--- a/examples/arduino-diecimila/Cargo.toml
+++ b/examples/arduino-diecimila/Cargo.toml
@@ -14,3 +14,6 @@ embedded-hal = "0.2.3"
 [dependencies.arduino-hal]
 path = "../../arduino-hal/"
 features = ["arduino-diecimila"]
+
+[dependencies.arduino-portable]
+path = "../arduino-portable"
--- a/examples/arduino-diecimila/src/bin/diecimila-usart.rs
+++ b/examples/arduino-diecimila/src/bin/diecimila-usart.rs
@@ -4,7 +4,7 @@
 use arduino_hal::prelude::*;
 use panic_halt as _;
 
-use embedded_hal::serial::Read;
+// use embedded_hal::serial::Read;
 
 #[arduino_hal::entry]
 fn main() -> ! {
@@ -15,10 +15,6 @@ fn main() -> ! {
     ufmt::uwriteln!(&mut serial, "Hello from Arduino!\r").void_unwrap();
 
     loop {
-        // Read a byte from the serial connection default
-        let b = nb::block!(serial.read()).void_unwrap();
-
-        // Answer
-        ufmt::uwriteln!(&mut serial, "Got {}!\r", b).void_unwrap();
+        arduino_portable::report(&mut serial);
     }
 }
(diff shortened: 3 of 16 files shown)


stappersg:
```text
arduino_portable::report(&mut serial);
```

Uhm, I like the idea, but ...
having `arduino` in the name is something that I dislike.   (note that I have no approval privilege, that I'm just a passenger, surely not a crew member)

The word _arduino_ has too many meanings. It **was** once a PCB with an `avr328` and it was/is an _"IDE"_.

I think ( I suggest) that _portable across targets_ is also portable outside the _arduino island_.  Think   ARM (STM32)  and  RISC-V.

mutantbob:
Sweet platypus on a unicycle : I forgot to add the new directory to the repo.  While I'm fixing that, I'll rename it to avr-portable.

Rahix:
Hey, thanks a lot for this PR!

I like the idea of including examples which show how to write "generic" code ontop of `avr-hal`.  That said, there are two dimensions to this:

- One one hand, there are "real" generics.  This is what your code is making use of.  Generics are best used when trying to either write a library (which then works with _any_ peripheral from _any_ supported MCU) or when writing application code that works with any of multiple similar peripherals (e.g. any of the USART blocks).  In code:

  ```rust
  pub fn read_command<H, USART: UsartOps<H, RX, TX>, RX, TX, CLOCK>(
      serial: &mut UsartReader<H, USART, RX, TX, CLOCK>,
      buffer: &mut [u8],
  ) -> Result<(), Error> {
      // reads a command from the given UART
  }
  ```
- On the other hand, the crates of `avr-hal` are designed to be (where possible) drop-in replacements for each other.  This means you could use cargo-features to switch between different boards.  The application code would, in this case, not require any generics at all.  Instead, it would use type-aliases and cfg-directives.  To make it more clear, consider this example code:

  ```toml
  # Cargo.toml
  [dependencies]
  cfg-if = "0.1.10"

  [dependencies.arduino-hal]
  git = "https://github.com/rahix/avr-hal"
  rev = "todotodo"

  [features]
  # Build for Arduino-Uno
  uno = ["arduino-hal/arduino-uno"]
  # Build for Arduino-Leonardo
  leonardo = ["arduino-hal/arduino-leonardo"]
  ```

  ```rust
  // src/main.rs
  #![no_std]
  #![no_main]

  use cfg_if::cfg_if;
  use panic_halt as _;
  use arduino_hal::hal::port;

  // Let's say on the Uno we want to use pin d0 and on Leonardo pin d4 for something
  cfg_if! {
      if #[cfg(feature = "uno")] {
          // D0
          type FooPin = port::Pin<port::mode::Output, port::PD0>;
      } else if #[cfg(feature = "leonardo")] {
          // D4
          type FooPin = port::Pin<port::mode::Output, port::PD4>;
      } else {
          compile_error!("Select uno or leonardo!");
      }
  };

  fn do_foo(pin: &mut FooPin) {
      pin.toggle();
  }

  #[arduino_hal::entry]
  fn main() -> ! {

      // D13 is an LED on both uno & leonardo so this just works (tm)
      led.set_high();

      let mut foo_pin: FooPin = cfg_if! {
          if #[cfg(feature = "uno")] {
              pins.d0.into_output()
          } else if #[cfg(feature = "leonardo")] {
              pins.d4.into_output()
          }
      };

      loop {
          led.toggle();
          do_foo(&mut foo_pin);
          arduino_hal::delay_ms(100);
      }
  }
  ```

  You can compile this program for both uno and leonardo so it is also "portable" in that sense.  This kind of "portability" is, what I imagine most users will need.  The big advantage is that it does not require any kind of generic juggling.

That said, there of course still is a use-case for the former as well.  So I would like to include your examples to demonstrate it.  However, I don't think we should modify the existing examples to make use of it.  Instead I would just add your `avr-portable` crate and add some binaries to call into the library to it.  You can then make it work for a number of boards using the cargo-feature mechanism shown above.

stappersg:
Does the existence of #264 allow to close this merge request?

I'm asking because dangling merge requests do obstruct the flow of evolution.

PR #260, opened 2022-04-23
Fix to support avrdude version numbers with multiple decimal points

ForgottenMaster:
I was getting a hello world app up and running with avr-hal on Windows and took this version of avrdude: https://github.com/mariusgreuel/avrdude/releases/tag/v6.3.1.1-windows

The problem is that the version number reported in this version was **6.3.1.1** which causes an error in the code as it assumes only one decimal point (thus parsing "3.1.1" as the minor version number)

I worked around it by grabbing the latest version of avrdude which is 6.99 and that now works fine


stappersg:
Looks good to me

Rahix (approved):
Thanks for the fix!

Issue #254, opened 2022-03-31
cannot find macro `llvm_asm` in this scope


piegamesde:
Apparently, `llvm_asm` was deprecated and removed over time, so it does not build anymore on current nightly.

Rahix:
Yes, this is known, there is already a MR to update this: #229  As everyone is however still using the older nightly until upstream fixes land, we'll have to wait a bit longer before merging this...

piegamesde:
Which nightly is the latest one that still works?

stappersg:
> Which nightly is the latest one that still works?

You are all welcome at https://github.com/avr-rust/blink/issues/38

PR #229, opened 2021-12-02
generic: delay: update to asm!

couchand:
Not mergeable until:
- [ ] LLVM [D114611](https://reviews.llvm.org/D114611) lands
- [ ] and is cherry-picked into Rust, and
- [x]  rust-lang/rust#91224 lands,
- [ ] and a nightly is released with both.

Will need to:

- [ ] update `rust-toolchain` file to use `nightly`
- [ ] remove `#![feature(asm)]` if https://github.com/rust-lang/rust/issues/72016 merges first (likely)

Tested on a local build of rustc bringing in both of those changes.

PR #261, opened 2022-04-24
Add support for `asm!()`

Patryk27:
Closes https://github.com/Rahix/avr-hal/issues/254.
Closes https://github.com/Rahix/avr-hal/pull/229.

Abstract & rest the same as for https://github.com/Rahix/avr-device/pull/97 🙂 

As for the LLVM IR, there's a small change in `delay()`:
```
older rustc:
  %7 = tail call addrspace(0) i16 asm sideeffect "1: sbiw $0,1\0A\09brne 1b", "=w,0"(i16 %6) #10, !srcloc !6

newer rustc:
  tail call addrspace(0) void asm sideeffect alignstack "1:\0Asbiw ${0}, 1\0Abrne 1b", "w,~{sreg},~{memory}"(i16 %5) #14, !srcloc !7
```
... but I think it looks alright (& the code works fine on an Atmega328p).

### DIFF
--- a/avr-hal-generic/Cargo.toml
+++ b/avr-hal-generic/Cargo.toml
@@ -18,3 +18,6 @@ features = ["unproven"]
 [dependencies.void]
 version = "1.0.2"
 default-features = false
+
+[build-dependencies]
+rustversion = "1.0"
--- a/avr-hal-generic/build.rs
+++ b/avr-hal-generic/build.rs
@@ -1,0 +1,16 @@
+fn main() {
+    println!("cargo:rerun-if-changed=build.rs");
+
+    maybe_enable_asm();
+}
+
+#[rustversion::before(1.59.0)]
+fn maybe_enable_asm() {
+    //
+}
+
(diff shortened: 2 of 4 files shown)


Rahix (approved):
Here as well: thanks a lot for this change!

piegamesde:
Thanks for the work. I still have the problem that `embedded-hal v0.2.7` does not support the new asm macro though, and `avr-hal-generic` seems to be depending on it. I tried one of the `1.0.0-alpha.X` versions of `embedded-hal` and they seem to work, however I can't locally override the dependency because of semver.

Rahix:
Hm, that's odd, where does `embedded-hal` need inline assembly?

piegamesde:
My bad, I read the compiler error wrong. It is `avr-device` which needs to be bumped

Rahix:
Ah, yeah, that is true indeed.  As soon as I have released a new version of `avr-device`, I will bump the dependencies here.

PR #263, opened 2022-05-02
Implement open drain output

agausmann:
Based on #217 as it has been inactive. I've rebased it on the latest `main` and addressed the comments in the original pull request.

### DIFF
--- a/avr-hal-generic/src/port.rs
+++ b/avr-hal-generic/src/port.rs
@@ -157,7 +157,6 @@ impl<PIN: PinOps, MODE: mode::Io> Pin<MODE, PIN> {
     /// Convert this pin into an open-drain output pin, setting the state to high.
     /// See [Digital Output Open Drain](#digital-output-open-drain)
     pub fn into_opendrain_high(mut self) -> Pin<mode::OpenDrain, PIN> {
-        unsafe { self.pin.out_clear() };
         unsafe { self.pin.make_input(false) };
         Pin {
             pin: self.pin,
--- a/avr-hal-generic/src/port.rs
+++ b/avr-hal-generic/src/port.rs
@@ -3,7 +3,7 @@
 //! Please take a look at the documentation for [`Pin`] for a detailed explanation.
 
 use core::marker::PhantomData;
-use embedded_hal::digital::v2::{OutputPin, InputPin};
+use embedded_hal::digital::v2::{InputPin, OutputPin};
 
 pub trait PinMode: crate::Sealed {}
 /// GPIO pin modes
@@ -17,6 +17,12 @@ pub mod mode {
     impl super::PinMode for Output {}
     impl Io for Output {}
     impl crate::Sealed for Output {}
+
+    // Pin is configures as a digital output with open drain behaviour
+    pub struct OpenDrain;
+    impl super::PinMode for OpenDrain {}
+    impl Io for OpenDrain {}
+    impl crate::Sealed for OpenDrain {}
 
     pub trait InputMode: crate::Sealed {}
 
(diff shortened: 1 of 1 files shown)


avr-hal-generic/src/port.rs
```diff
@@ -3,7 +3,7 @@
 //! Please take a look at the documentation for [`Pin`] for a detailed explanation.
 
 use core::marker::PhantomData;
-use embedded_hal::digital::v2::{OutputPin, InputPin};
+use embedded_hal::digital::v2::{InputPin, OutputPin};
```
agausmann, line 6: Oops I left rustfmt on 😅

avr-hal-generic/src/port.rs
```diff
@@ -137,6 +143,28 @@ impl<PIN: PinOps, MODE: mode::Io> Pin<MODE, PIN> {
...
+            pin: self.pin,
+            _mode: PhantomData,
+        }
+    }
+
+    /// Convert this pin into an open-drain output pin, setting the state to high.
+    /// See [Digital Output Open Drain](#digital-output-open-drain)
+    pub fn into_opendrain_high(mut self) -> Pin<mode::OpenDrain, PIN> {
+        unsafe { self.pin.out_clear() };
+        unsafe { self.pin.make_input(false) };
```
Rahix, line 161: The `out_clear()` here is not needed, `make_input()` already takes care of it:

https://github.com/Rahix/avr-hal/blob/5d757696e338104622a0f89c51020b718bf48b62/avr-hal-generic/src/port.rs#L563-L572

Doing it here is actually harmful because it will lead to a glitch on the electrical pin:  If the pin was previously a HIGH output, it will briefly go low before entering the floating state. `make_input()` has the order correct such that this glitch cannot occur.

avr-hal-generic/src/port.rs
```diff
@@ -18,6 +18,12 @@ pub mod mode {
     impl Io for Output {}
     impl crate::Sealed for Output {}
 
+    // Pin is configures as a digital output with open drain behaviour
```
Rahix, line 21:
```suggestion
    /// Pin is configured as a digital output with open drain behaviour
```

Rahix (changes requested):
Hey, thanks for picking this back up!  I've got two comments, apart from that it looks good to go.

Rahix (approved):
Looks good, thanks!  Just to make sure: Did you test that it behaves correctly on real hardware?  If not, I can take care of it in the coming days.

agausmann:
I tested it with this modified uno-blink example, and it seems to work:
(Switched to d12 pin so there's no other circuitry going on, and also because it's the reverse current direction of the built-in LED)

Measured the resistance to ground and the voltage (with a pullup) and it seems to work as expected.

```rust
#![no_std]
#![no_main]

use panic_halt as _;

#[arduino_hal::entry]
fn main() -> ! {

    // Digital pin 13 is also connected to an onboard LED marked "L"
    let mut led = pins.d12.into_opendrain_high();

    loop {
        led.set_high();
        arduino_hal::delay_ms(500);
        led.set_low();
        arduino_hal::delay_ms(500);
        led.set_high();
        arduino_hal::delay_ms(500);
        led.set_low();
        arduino_hal::delay_ms(1000);
    }
}
```

One thing I noticed is that there isn't a `toggle` method or an implementation of `ToggleableOutputPin` for `Pin<OpenDrain>` (in fact, there is no `StatefulOutputPin` or `ToggleableOutputPin` for `Pin<Output>` either :thinking: ). Is that something that we'd want?

agausmann:
One thing I noticed is that there isn't a `toggle` method or an implementation of `ToggleableOutputPin` for `Pin<OpenDrain>` (in fact, there is no `StatefulOutputPin` or `ToggleableOutputPin` for `Pin<Output>` either :thinking: ). Is that something that we'd want?

stappersg:
> One thing I noticed is that there isn't a `toggle` method or an implementation of `ToggleableOutputPin` for `Pin<OpenDrain>` (in fact, there is no `StatefulOutputPin` or `ToggleableOutputPin` for `Pin<Output>` either thinking ). Is that something that we'd want?

We should want that the hardware provides.  If the _open drain output_ can **not** be read back, **don't** do it, do **not fake toggle** for something that **can not be toggled**.

(My definition of _toggle_ is an _atomic read level and write invert level back_, something that will work for a push-pull-output-pin. (But we are dealing with _open drain outpt_.))

agausmann:
@stappersg Glancing at the ATmega328P datasheet, the bits of DDRx registers are labeled R/W, so I'm pretty sure we can read the state of the pin, and I'd assume that's also a feature on other models. So that's some justification for a `StatefulOutputPin` impl.

Rahix:
> One thing I noticed is that there isn't a toggle method or an implementation of ToggleableOutputPin for Pin<OpenDrain> (in fact, there is no StatefulOutputPin or ToggleableOutputPin for Pin<Output> either thinking ). Is that something that we'd want?

Sure, this would also be nice.  But I suggest adding it in a separate MR...

> didn't even need to check that, we already have `is_high()` and `is_low()` methods for open-drain pins

Careful!  `is_high()`/`is_low()` have a different meaning than `is_set_high()`/`is_set_low()`.  The former is meant to read the electrical state of the pin.  Thus their return value can be influenced by outside circuitry.  The latter _must not_ read the electrical state but return what state the microcontroller last set the pin to.  This means their return value _must not_ be influenced by outside circuitry.

This is especially important for open-drain outputs.  After a `set_high()`, the open-drain pin might still return `true` for `is_low()` when pulled to GND externally.  But `is_set_low()` must not be `true`.

PR #266, opened 2022-05-10
Fix delay inline assembly for new rust versions

Rahix:
The register needs to be marked `inout` to have the same behavior as the old code.

@Patryk27, just fyi.  Without this change, the code still "works" but delays are much longer.  I didn't dig deep enough to truly understand why this is happening but I figured changing the behavior back to what `llvm_asm!()` was doing is good enough for now.


Patryk27:
Thanks for pinging me, interesting stuff - gotta look closer!

PR #267, opened 2022-05-10
add an example of how to use timer interrupts

mutantbob:
I took the code from https://github.com/Rahix/avr-hal/issues/75 and modified it until it compiled and worked on the Arduino Uno.  This is the new `examples/arduino-uno/src/bin/uno-timer.rs` example which shows how to service timer-based interrupts.

### DIFF
--- a/examples/arduino-uno/src/bin/uno-timer.rs
+++ b/examples/arduino-uno/src/bin/uno-timer.rs
@@ -9,7 +9,6 @@ and then modernized to account for API drift since 2020
 
 */
 
-use arduino_hal::hal::port::Dynamic;
 use arduino_hal::port::mode::Output;
 use arduino_hal::port::Pin;
 use arduino_hal::prelude::*;
@@ -20,7 +19,7 @@ use panic_halt as _;
 use ufmt::{uWrite, uwriteln};
 
 struct InterruptState {
-    blinker: Pin<Output, Dynamic>,
+    blinker: Pin<Output>,
 }
 
 static mut INTERRUPT_STATE: mem::MaybeUninit<InterruptState> = mem::MaybeUninit::uninit();
--- a/examples/arduino-uno/src/bin/uno-timer.rs
+++ b/examples/arduino-uno/src/bin/uno-timer.rs
@@ -1,0 +1,136 @@
+#![no_std]
+#![no_main]
+#![feature(abi_avr_interrupt)]
+
+/*
+ This is derived from Rahix' comment to
+https://github.com/Rahix/avr-hal/issues/75
+and then modernized to account for API drift since 2020
+
+*/
+
+use arduino_hal::hal::port::PB5;
+use arduino_hal::port::mode::Output;
+use arduino_hal::port::Pin;
+use arduino_hal::prelude::*;
+use avr_device::atmega328p::tc1::tccr1b::CS1_A;
+use avr_device::atmega328p::TC1;
+use core::mem;
+use panic_halt as _;
+use ufmt::{uWrite, uwriteln};
+
+struct InterruptState<LED> {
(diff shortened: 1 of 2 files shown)


examples/arduino-uno/src/bin/uno-timer.rs
```diff
@@ -0,0 +1,136 @@
...
+use arduino_hal::prelude::*;
+use avr_device::atmega328p::tc1::tccr1b::CS1_A;
+use avr_device::atmega328p::TC1;
+use core::mem;
+use panic_halt as _;
+use ufmt::{uWrite, uwriteln};
+
+struct InterruptState<LED> {
+    blinker: Pin<Output, LED>,
+}
```
Rahix, line 24: I think it would be nicer to get rid of the `<LED>` type parameter and instead use a "dynamic Pin" as described in ["Downgrading"](https://rahix.github.io/avr-hal/avr_hal_generic/port/struct.Pin.html#downgrading).  The type would then be

```rust
struct InterruptState {
    blinker: Pin<Output>,
}
```

examples/arduino-uno/src/bin/uno-timer.rs
```diff
@@ -0,0 +1,136 @@
...
+    clock_hz / target_hz / prescale - 1
+}
+
+pub fn rig_timer<W: uWrite<Error = void::Void>>(tmr1: &TC1, serial: &mut W) {
+    /*
+     https://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-7810-Automotive-Microcontrollers-ATmega328P_Datasheet.pdf
+     section 15.11
+    */
+
+    const ARDUINO_UNO_CLOCK_FREQUENCY_HZ: u32 = 16_000_000;
```
Rahix, line 92: Please reference the clock via the clock-type from `arduino-hal`: [`arduino_hal::DefaultClock`](https://rahix.github.io/avr-hal/arduino_hal/type.DefaultClock.html)

You can get the frequency using `arduino_hal::DefaultClock::FREQ`.

examples/arduino-uno/src/bin/uno-timer.rs
```diff
@@ -0,0 +1,136 @@
...
+        .void_unwrap();
+    }
+    loop {
+        if true {
+            avr_device::asm::sleep()
+        } else {
+            arduino_hal::delay_ms(100);
+
+            ufmt::uwriteln!(&mut serial, "counter = {}", tmr1.tcnt1.read().bits()).void_unwrap();
+        }
```
Rahix, line 74: Please drop the `if true {}` blocks and instead always do the print.  Let's keep the example code as simple as possible.

Rahix (commented):
Hey, thanks for the contribution - more examples are always better :)

examples/arduino-uno/src/bin/uno-timer.rs
mutantbob: I am having difficulty finding the right `use` clause to get the `Pin<Output,Dynamic>` to compile.  I'll keep banging on it.  So far I just have
```
struct InterruptState {
    blinker: Pin<Output, <PB5 as PinOps>::Dynamic>,
}
```

Rahix: Not sure what you're doing wrong...

```rust
struct InterruptState {
    blinker: arduino_hal::port::Pin<Output>,
}
```

should just work.  Maybe the problem is elsewhere?  Are you actually downgrading the pin before storing it in the static?

mutantbob: Ah, the second parameter to the generic is optional...

Rahix (approved):
Thanks!

PR #272, opened 2022-05-15
Reimplement simple pwm

### LINKED ISSUE #251: Analog output for Arduino Uno
The Arduino supports PWM on a few ports, yet looking through ports.rs, the HAL only seems to allow digital outputs. Is there an example of doing analog writes? Does this require including a lower level implementation?

dalpil:
This is a reimplementation of the "simple" fastpwm functionality from the old-branch.

I've only been able to test this on a Arduino Uno, a Leonardo and a Trinket, but I've tried to add implementations for some of the other MCUs (atmega48p, atmega168, atmega328pb, atmega2650) as well, based on the code in the old-branch.

I also added implementations for atmega32u8, attiny85, attiny88 and atmega1280 as well, which wasn't available in the old branch IIRC.

Any feedback would be greatly appreciated.

Closes #251.

### DIFF
--- a/arduino-hal/src/lib.rs
+++ b/arduino-hal/src/lib.rs
@@ -113,7 +113,7 @@ pub use delay::{delay_ms, delay_us, Delay};
 #[cfg(feature = "board-selected")]
 pub mod port;
 
-#[cfg(feature = "mcu-atmega")]
+#[cfg(feature = "board-selected")]
 pub mod simple_pwm;
 
 #[doc(no_inline)]
--- a/arduino-hal/src/lib.rs
+++ b/arduino-hal/src/lib.rs
@@ -113,7 +113,7 @@ pub use delay::{delay_ms, delay_us, Delay};
 #[cfg(feature = "board-selected")]
 pub mod port;
 
-#[cfg(feature = "board-selected")]
+#[cfg(feature = "mcu-atmega")]
 pub mod simple_pwm;
 
 #[doc(no_inline)]
--- a/arduino-hal/src/lib.rs
+++ b/arduino-hal/src/lib.rs
@@ -112,6 +112,10 @@ pub use delay::{delay_ms, delay_us, Delay};
 
 #[cfg(feature = "board-selected")]
 pub mod port;
+
+#[cfg(feature = "board-selected")]
+pub mod simple_pwm;
+
 #[doc(no_inline)]
 #[cfg(feature = "board-selected")]
 pub use port::Pins;
--- a/arduino-hal/src/simple_pwm.rs
+++ b/arduino-hal/src/simple_pwm.rs
@@ -1,0 +1,10 @@
+//! Simple PWM output for supported Pins.
+//!
+//! This module implements simple (FastPWM) PWM output for supported Pins.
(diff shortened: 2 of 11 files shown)


dalpil:
Fixed my typo which tried to use simple_pwm on all MCU types instead of just the atmega ones.

dalpil:
This PR should probably have been a Draft from the beginning, sorry about the noise..

I'll add simple_pwm implementations for the remaining MCU's and fix the documentation errors that I've made this coming weekend.

dalpil:
I didn't set out to implement this for all the (currently) supported MCUs from the beginning, but it was rather fun, so I just kept going.

This should now be fairly "complete" as far as the current MCU support goes.

The tests break because they depend on a new release of avr-device (due to the missing timer-patches for atmega1280), so I guess the "waiting-for-avr-device" label would be appropriate here.

**Pros**
- I can now fade my LEDs.
- The end-user API stays the same as on the old-branch, which makes it easier to port my projects.
- I learned a ton about the different timers on AVR-chips just by skimming the datasheets.

**Cons**
- I'm not sure if this thin abstraction will fit the 4809. Those timers are configured in a totally different way, and since I lack the necessary hardware, I haven't been able to take a stab at it at all.
- This is just plain old FastPWM, which now feels a bit "meh", considering how much cool stuff you can do with the rest of the timers and/or modes on AVRs. So I guess this is more of a stopgap between no PWM API at all, and a fully fleshed out PWM API with support for phase-correct wave generation modes, frequency control, PLLs, dead-time configuration and what have you. This is being discussed in other issues; #44 for example.
- Adding this as a "weak" API right now might cause headaches/churn further down the road, if a more powerful API come along to replace it.

As said; any feedback is greatly appreciated.

rogday:
Thank you SO MUCH for this PR! I know nothing about microcontrollers and that means I'm hesitant to use workaround from other issues that uses unfamiliar for me internal names. Using this branch I was able to play with my LED too :D

Rahix:
Hey @dalpil, with #283 the new version of `avr-device` is now used.  Please rebase your PR ontop of the latest `main` to pull in this change and hopefully get CI to pass this way.

Sorry for the slow progress from my side here...

dalpil:
No worries, I'm guessing that you maintain these libraries as a hobby, and it wasn't my intention to cause any stress by opening a fairly big unsolicited pull request. I'm just happy to tinker with this. :)

Rahix (approved):
Hey, thank you very much again for this changeset!  Sorry for all the delays from my side...

I took the time to review it now, everything looks good enough that I will merge it as is.  I think it is better to get this into the project ASAP and iterate on anything else in the future.

Thanks btw for the cleanly rebased commit series.  Not many people take the time to do this so I appreciate it a lot.

I tested the Uno example. It works fine but for some reason the built-in LED on pin 13 seems to behave weirdly while it runs... Configuring the pin as output to drive it explicitly fixes this.  But I do not think we should worry about it here.

Finally one minor nit for the future:  You should configure your editor to always save files with a trailing newline.  In some files here, it is missing, but I will fix this for you after merging.

PR #275, opened 2022-05-28
Ravedude: Allow skipping the reset question for automated resets

TerminatorNL:
This commit allows us to bypass the "reset" instructions by providing a additional command-line arguments to ravedude.

This adds one additional option:
<pre>
    -d, --reset-delay <reset-delay>    
            This assumes the board is already resetting. Instead of giving the reset instructions and waiting for user
            confirmation, we wait the amount in milliseconds before proceeding. Set this value to 0 to skip the board
            reset question instantly
</pre>

Why this change: I needed this myself as I had already implemented a self-reset function through the USB interface which I can call before flashing.

Thank you so much for all your hard work for Rust on AVR. I am loving it!

### DIFF
--- a/ravedude/src/main.rs
+++ b/ravedude/src/main.rs
@@ -2,6 +2,9 @@ use anyhow::Context as _;
 use colored::Colorize as _;
 use structopt::clap::AppSettings;
 
+use std::time::Duration;
+use std::thread;
+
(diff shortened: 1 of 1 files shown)


Rahix (approved):
This makes a lot of sense, thanks!

PR #285, opened 2022-06-27
Add Attiny84 to attiny-hal

blueberry-fan:
This adds feature _attiny84_ to attiny-hal. The `pins` macro and necessary types are implemented like the ones for attiny85/88.


Rahix (approved):
Looks good, thank you very much!

PR #288, opened 2022-07-06
Add macos dependency instructions

NickLarsenNZ:
Add install instructions for running on Macos, this might save some users a few minutes of searching and guessing.


Rahix (approved):
This is a good idea, thanks!

PR #290, opened 2022-07-09
Fix DataRegisterEmpty and add TxComplete on Usart

Idix:
Wrong bit was set when listening to `DataRegisterEmpty` Event.
Added `TxComplete` Event


Rahix (approved):
Thanks!

PR #296, opened 2022-08-07
Update README.md

PixelcrafterEXE:
added libudev and cc dependencies to linux dependencies

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Install dependencies:
 
 - Ubuntu
   ```bash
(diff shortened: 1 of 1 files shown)


Rahix:
Hi, thanks for the PR.

udev and build-essential make sense to me, but `cmake`?  Do you remember what step failed without cmake?

Rahix:
I've manually merged the part which I do not have questions about.  If you can tell why `cmake` was required, we can add that too.  I'll close this PR in the meantime as part of it has been applied.  Feel free to open a new one for `cmake` if it is indeed required.

PR #306, opened 2022-08-31
atmega-hal: Add support for ATmega1284P

icanwalkonwater:
Added support for the ATmega1284P by following [the datasheet](https://ww1.microchip.com/downloads/en/DeviceDoc/doc8059.pdf).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,6 +55,10 @@ jobs:
             spec: atmega48p
             crate: atmega-hal
           - type: mcu
+            name: atmega1284p
+            spec: atmega1284p
+            crate: atmega-hal
+          - type: mcu
             name: attiny85
             spec: attiny85
             crate: attiny-hal
--- a/avr-specs/avr-atmega1284p.json
+++ b/avr-specs/avr-atmega1284p.json
@@ -1,0 +1,25 @@
+{
+  "arch": "avr",
+  "atomic-cas": false,
+  "cpu": "atmega1284p",
+  "data-layout": "e-P1-p:16:8-i8:8-i16:8-i32:8-i64:8-f32:8-f64:8-n8-a:8",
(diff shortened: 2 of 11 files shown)


Rahix:
Hi, thanks a lot for the PR!

Can you take a look at the CI failures?  It seems some other MCUs no longer build with your changes.

icanwalkonwater:
It seems I forgot to add a `#[cfg]` somewhere, it should compile fine now.

Rahix (approved):
Changes look good, thanks!

I don't have the hardware so I'll have to rely on your testing.  How much did you verify this code on a real device?  Or is it just a blind implementation for now?  I think I'll merge the changes either way, but of course it would be nice to know that things have been proven to work ;)

icanwalkonwater:
Thank you =)

I have the board, it is salvaged from an Anet A8 printer.
I only really needed the USART implementation, but it felt wrong to not do the other ones.

I tested the USART0 and LED and it does work, as for the other peripherals I might need the ADC at some point but since it is not a developement board, the pins arent't exposed making it hard to test with my non-existent equipement.

I was thinking of also adding it to the `arduino-hal` crate with the pins renamed [like this](https://github.com/ralf-e/ANET-3D-Board-V1.0/blob/master/ANET3D_Board_Schematic.png). It will make my life easier but that may not be the purpose of this crate (since its called arduino you know).

Rahix:
Alright, good enough for me.  Let's not worry about the other pieces until someone tries using them.

> I was thinking of also adding it to the arduino-hal crate named something like anet3d with the pins renamed like this. It will make my life easier but that may not be the purpose of this crate (since its called arduino you know).

My suggestion would be that you create yourself a custom "board support crate" that essentially has the same structure as `arduino-hal` but only for your board.  This is the usual approach when working with custom hardware.

PR #335, opened 2022-09-08
Pacify clippy

mutantbob:
`cargo clippy` had some complaints.  I think I pacified them.  There are other complaints, but my instincts tell me that "solving" those would break APIs.

### DIFF
--- a/examples/arduino-uno/src/bin/uno-ext-interrupt.rs
+++ b/examples/arduino-uno/src/bin/uno-ext-interrupt.rs
@@ -10,7 +10,6 @@
 #![feature(abi_avr_interrupt)]
 
 use panic_halt as _;
-use core::ops::Range;
 use core::sync::atomic::{AtomicBool, Ordering};
 use arduino_hal::port::{mode, Pin};
 use either::*;
@@ -18,7 +17,7 @@ use either::*;
 static REVERSED: AtomicBool = AtomicBool::new(false);
 
 fn is_reversed() -> bool {
-    return REVERSED.load(Ordering::SeqCst);
+    REVERSED.load(Ordering::SeqCst)
 }
 
 #[avr_device::interrupt(atmega328p)]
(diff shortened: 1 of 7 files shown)


examples/arduino-uno/src/bin/uno-ext-interrupt.rs
```diff
@@ -63,6 +63,6 @@ fn main() -> ! {
 
     loop {
         blink_for_range(0..10, &mut leds);
-        blink_for_range(10..0, &mut leds);
+        blink_for_range((0..10).rev(), &mut leds);
```
Rahix, line 66: This is a change of behavior, see this [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=270f7fdddc13442c407a6b46dadc8380).

Arguably the original code doesn't make much sense here anyway but I think it is better to fix it properly instead.  I see two options:

1. Just dropping the "count down" line and leaving only the "count up".  This should be functionally equivalent to the previous version.
2. Or simplifying this `blink_for_range()` function by not passing the range.  I think this would make the whole example a lot more readable...

Rahix (commented):
Hi, thanks for the PR.  I think it's a good idea to clean up a bit.  I wouldn't be afraid of breaking changes if they provide a tangible improvement.  If you feel they do, feel free to include more fixes here.

Rahix:
I see in the commits for the PR that you originally had a new example included here: 53f594d  This example doesn't show up in the final diff, however?  If your intention was to also submit that example, maybe send a new PR with just that?

Rahix:
Ah, right, that example is already upstream as of #267 so ignore my previous comment.  I just got confused by seeing the commit here.

examples/arduino-uno/src/bin/uno-ext-interrupt.rs
Rahix: @mutantbob, can you take care of this when you get the chance?  Otherwise the changes look good and I'd like to get them merged :)  If you can't get to it right now, I can offer to fix it up and pull in your changes manually.  If that's okay with you.

simonsso: It looks like the original writer intended the code to work as in this patch, and yes it is cluttering up the example.

Rahix:
Rebased ontop of latest `main`.

PR #344, opened 2022-09-30
avr-specs: ATtiny2313 -mmcu "at90s2313" -> "attiny2313"

### LINKED ISSUE #343: attiny2313 targe uses the wrong ISA
Not sure if by design or mistake, but the attiny2313 spec sets `--mmcu=at90s2313`:
```
$ grep tiny2313 -A 2 avr-specs/sync-from-upstream.py
    "attiny2313": {
        "cpu": "at90s2313",
    },
```

According to the avr-gcc manual, while code for the at90s2313 can probably run in the attiny2313, they actually have a slightly different ISA:
```
GCC supports the following AVR devices and ISAs:

avr2

    “Classic” devices with up to 8 KiB of program memory.
    mcu = (...) at90s2313, (...) .

avr25

    “Classic” devices with up to 8 KiB of program memory and with the MOVW instruction.
    mcu = (...) attiny2313, (...).
```

urlordjames:
resolves #343

### DIFF
--- a/avr-specs/avr-attiny2313.json
+++ b/avr-specs/avr-attiny2313.json
@@ -1,7 +1,7 @@
 {
   "arch": "avr",
   "atomic-cas": false,
-  "cpu": "at90s2313",
+  "cpu": "attiny2313",
   "data-layout": "e-P1-p:16:8-i8:8-i16:8-i32:8-i64:8-f32:8-f64:8-n8-a:8",
   "eh-frame-header": false,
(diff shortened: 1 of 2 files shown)


Rahix (approved):
Alright, tried on a few distros - it seems to work fine there.  Thanks a lot for your lightning fast response!  Again, this wasn't really a critical problem so you don't have anything to worry about.

PR #346, opened 2022-10-07
README: Add link to windows install docs

X-Wera:
Added a link for avr-gcc setup because this was a nuisance to find.


Rahix (approved):
Fair enough, thanks!

PR #349, opened 2022-10-08
README: Genuine Windows Quickstart with Scoop

X-Wera:
Scoop seems to be the quickest way to install our dependencies and making them work on Windows(10).

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,9 +7,6 @@ You need a nightly Rust compiler for compiling Rust code for AVR.  The correct v
 
 Install dependencies:
 
-- Windows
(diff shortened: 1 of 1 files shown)


Rahix (approved):
Sounds good.  Does this also install `avr-libc` as well?  I suppose if it was working for you, it probably did...

X-Wera:
Yes. Its part of [avr-gcc for windows](https://blog.zakkemble.net/avr-gcc-builds/).

X-Wera:
I should note also, avrdude is included in avr-gcc but ravedude cannot locate it (I think is the issue.). Simply installing avrdude by itself. Makes it work immediately. So windows users will have 2 installs of avrdude but I feel its a minor issue.

PR #351, opened 2022-10-09
ravedude: Add flake.nix

Patryk27:
(not sure if you'd be interested in a pull request like that, but I figured it won't hurt to try 🙂)

This PR adds a `flake.nix` that Nix / NixOS users can use to install ravedude on their machines (considering that compiling it requires udev and pkg-config, so it's not as straightforward as "regular" `cargo install`).

(btw, for NixOS flake users: that requires pulling the dependency as `inputs.ravedude.url = "github:Rahix/avr-hal?dir=ravedude";`)

### DIFF
--- a/ravedude/README.md
+++ b/ravedude/README.md
@@ -24,6 +24,10 @@ Next, install the latest version from crates.io with the following command:
 cargo install ravedude
 ```
 
+(alternatively, if you're using NixOS + Flakes, you can install `ravedude` by
+adding `inputs.ravedude.url = "github:Rahix/avr-hal?dir=ravedude";` and then
+`environment.systemPackages = [ ravedude.defaultPackage."${system}" ];`.)
+
 Now you need to add *ravedude* to your project.  For example in a project for
 Arduino Uno, place the following into your `.cargo/config.toml` (**not in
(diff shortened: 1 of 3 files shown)


Rahix (approved):
Fine by me, I just can't guarantee I won't accidentally break these files ;)  Is there anything I should know? For example, will they need an update when I release a new ravedude version?

Rahix:
> (btw, for NixOS flake users: that requires pulling the dependency as inputs.ravedude.url = "github:Rahix/avr-hal?dir=ravedude";)

Something we should mention in the README?  Sorry, I don't have any clue about NixOS and best practices there...

Patryk27:
> I just can't guarantee I won't accidentally break these files ;)

Yeah, that's understandable 😇 

> Is there anything I should know? For example, will they need an update when I release a new ravedude version?

The only things that come to my mind are major toolchain bumps and requiring new external system-dependencies (like, I dunno, some Xorg libs) -- if that ever happens, feel free to ping me!

> Something we should mention in the README?

Right, right - added 🙂

Rahix:
> if that ever happens, feel free to ping me!

Will do, thanks!

PR #361, opened 2022-11-05
Implement `hal::blocking::i2c::Transactional` and `hal::blocking::i2c::TransactionalIter` for `I2c`

### LINKED ISSUE #338: Add impl of `hal::blocking::i2c::Transactional` to `avr_hal_generic::i2c::I2c`
I'd love to be able to use the `Transactional` IO model of `embedded_hal` for `I2c` (https://docs.rs/embedded-hal/latest/embedded_hal/blocking/i2c/trait.Transactional.html), would it be possible to add this in? I'm happy to do the work if there are no objections or concerns.

The only question I have is related to multiple read operations -- in the linked docs for `Transactional`, it states: `If the last operation is a Read the master does not send an acknowledge for the last byte.` Would this mean that read operations that are not the last one in the batch send an ACK versus a NACK? Or would each read operation send a NACK?

linclelinkpart5:
Closes #338.

Would love some eyes on the calls to `raw_start`, I believe I'm following the contracts as listed in [https://docs.rs/embedded-hal/latest/embedded_hal/blocking/i2c/trait.Transactional.html] and [https://docs.rs/embedded-hal/latest/embedded_hal/blocking/i2c/trait.TransactionalIter.html].

### DIFF
--- a/avr-hal-generic/src/i2c.rs
+++ b/avr-hal-generic/src/i2c.rs
@@ -388,6 +388,67 @@ impl<H, I2C: I2cOps<H, SDA, SCL>, SDA, SCL, CLOCK> hal::blocking::i2c::WriteRead
     }
 }
 
+impl<H, I2C: I2cOps<H, SDA, SCL>, SDA, SCL, CLOCK> hal::blocking::i2c::Transactional
+    for I2c<H, I2C, SDA, SCL, CLOCK>
+{
+    type Error = Error;
+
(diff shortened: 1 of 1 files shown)


Rahix (commented):
Hi, thanks a lot for the PR!  Your implementation looks good to me from taking a look at it but I didn't run any tests so far.  I assume you verified that it's working on real hardware?

PR #362, opened 2022-11-09
add support for arduino mega 1280

ReinoutHeeck:
These changes add support for the original 'Arduino Mega' board that had an atmega1280 mcu and a generic FTDI usb-serial chip.

- adds board 'mega1280' to ravedude,
- renames mega2560.rs in arduino-hal to mega.rs since it is now shared between two boards
- adds examples that have been verified to compile and run

(no version numbers have been changed)


ravedude/src/board.rs
```diff
@@ -199,6 +200,35 @@ impl Board for ArduinoLeonardo {
...
+            partno: "atmega1280",
+            baudrate: Some(57600),
+            do_chip_erase: false,
+        }
+    }
+
+    fn guess_port(&self) -> Option<anyhow::Result<std::path::PathBuf>> {
+        Some(find_port_from_vid_pid_list(&[
+            (0x0403,0x6001), // FTDI FT232R USB UART
+        ]))
```
Rahix, line 227: Please change this as follows:

```suggestion
        Some(Err(anyhow::anyhow!("Not able to guess port")))
```

As this board only shows up as a generic FTDI chip, there is too much ambiguity here for performing automatic detection :/  You'll have to pass the port using the `RAVEDUDE_PORT` environment variable for now... In the future I'd like to add a more user-friendly mechanism for this, but until then, I don't think we should auto-detect the generic chips as certain boards.

Rahix (changes requested):
Hi, thanks for the PR!  Almost everthing looks good, I only have one comment which you can find below.  Let's get this merged very soon so I don't have to revert my mistake of merging the template support already ;)

ravedude/src/board.rs
ReinoutHeeck: Done, 

I also added a README to the mega1280 examples with instructions telling hoe to pass the port to ravedude

Rahix (approved):
Thanks a lot!

PR #367, opened 2022-11-17
Add Digistump Digispark board

ReinoutHeeck:
Adds a Digispark board. 

In the past I used the Trinket board configuration on my Digisparks, but that was buggy.

Now the Digispark has the correct clock frequency and the missing pin added.
Also its documentation now mentions which pins USB and Reset are connected to.


simonsso (approved):
Looks good to me
(it is exactly the same as what I independently wrote in PR #386)

PR #368, opened 2022-11-21
Draft: Add support for Atmega8

### LINKED ISSUE #363: Support of Atmega8
Hi, I would like to use the Atmega8 chip, which do not seem to be supported by now.
I don't know enough of the avh-hal system to implement it myself (but I could try with some help)

xfmoulet:
based on PR #306

closes #363


xfmoulet:
I have an issue (so closing this PR for the moment) : adps().prescaler_2 does not exist only bit definitions ... 
I try to find a way to temporarily disable the ADC device for the moment (I guess the other possibility would be to enable is automated extraction in avr-device ?)

Rahix:
Let's keep the PR open while you're still working on it.  Closing a PR usually means abandoning the work which probably isn't what you intend to do ;)  You can mark the PR as a draft if you feel more comfortable with that, but it's fine for me either way.

Rahix:
@xfmoulet, for your ADC problem, maybe we just need to apply the proper patches to the register definitions in `avr-device`?  Maybe you can cross-check the registers in the datasheet of the ATmega8 with the ATmega328P.  If they are the same, we just need to apply the same patches as for ATmega328P for the ATmega8.  Those would be <https://github.com/Rahix/avr-device/blob/main/patch/common/adc.yaml>.

Rahix:
Closing per https://github.com/Rahix/avr-hal/issues/363#issuecomment-1372488475.

Issue #388, opened 2023-01-03
Linker errors when compiling for ATmega2560 with recent rust compiler (>2022-10-22)


arnemileswinter:
Hello!

I'm pretty new to cross-compiling.

I created a nix environment in which i can compile the .ELF file for some targets.

For example, the target avr-atmega328p.json works:
```
bash-5.1$ cargo build --target avr-specs/avr-atmega328p.json 
   Compiling compiler_builtins v0.1.85
   Compiling core v0.0.0 (/nix/store/r9ns07w4z808rpysvg4z84bx3q525yv7-rust-mixed/lib/rustlib/src/rust/library/core)
   Compiling avr-hal-generic v0.1.0 (https://github.com/rahix/avr-hal?rev=4170a773d4d76cc93433d2455ed8b14e573ebe70#4170a7
   [..... i am skipping some lines ....]
   Compiling helloworld v0.1.0 (/home/arne/Projects/hobby/rust-arduino-led-blink)
    Finished dev [optimized + debuginfo] target(s) in 20.01s
```
And i get `target/avr-atmega328p/debug/helloworld.elf`.

But the arduino that i have at my disposal is the atmega2560, i get the following output:
```
bash-5.1$ cargo build --target avr-specs/avr-atmega2560.json --verbose
       Fresh core v0.0.0 (/nix/store/r9ns07w4z808rpysvg4z84bx3q525yv7-rust-mixed/lib/rustlib/src/rust/library/core)
       Fresh rustc-std-workspace-core v1.99.0 (/nix/store/r9ns07w4z808rpysvg4z84bx3q525yv7-rust-mixed/lib/rustlib/src/rust/library/rustc-std-workspace-core)
       Fresh unicode-ident v1.0.6
       Fresh compiler_builtins v0.1.85
       Fresh proc-macro2 v1.0.49
       Fresh quote v1.0.23
       Fresh nb v1.0.0
       Fresh syn v1.0.107
       Fresh proc-macro-hack v0.5.20+deprecated
       Fresh rustversion v1.0.11
       Fresh ufmt-macros v0.2.0
       Fresh avr-device-macros v0.4.0
       Fresh nb v0.1.3
       Fresh ufmt-write v0.1.0
       Fresh bare-metal v1.0.0
       Fresh cfg-if v0.1.10
       Fresh void v1.0.2
       Fresh vcell v0.1.3
       Fresh ufmt v0.1.2
       Fresh paste v1.0.11
[...]

PR #395, opened 2023-01-24
Upgrade rust-toolchain.toml to fix CVE-2022-46176

jgerrish:
This upgrades the used Rust toolchain to fix CVE-2022-46176, a security bug with SSH host key checking in cargo.

I haven't tested it fully, so I understand if you don't want to incorporate it yet.  Because it's such a jump in versions, it may create other issues.  You have a larger community of great users, maybe some can test it before pushing it to main.

I'd love for more selective cherry-picking of patches in Rust, but this is the world we live in.

Have a great day, and thank you for this wonderful project!

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "nightly-2022-07-10"
+channel = "nightly-2023-01-10"
 components = [ "rust-src" ]
 profile = "minimal"


Rahix:
In general, I think there is nothing speaking against keeping our compiler version as close to upstream as possible.

Unfortunately, it seems CI isn't passing for some MCUs.  Without looking deeper into it, maybe this is related to #388?  Paging @agausmann.

Rahix:
Okay, I checked, the failure really is #388.  @agausmann kindly started working on an upstream fix, see <https://github.com/rust-lang/rust/pull/106619>.

edupsousa:
Writing just to let you known that the older toolchain is also causing some trouble with the most recent versions of the rust-analyzer, something about the lack of a proc-macro server in toolchains before 2022-07-29. Tested and the problem goes away with the toolchain introduced in this PR.

Rahix:
Toolchain was finally upgraded in #427.

PR #403, opened 2023-03-30
Add support for programming AVR EEPROM memory if .eeprom section foun…

cr1901:
…d in ELF file.

Tested locally and working. This is meant to complement a set of [crates](https://github.com/cr1901/postcard-infomem) I'm working on...

### DIFF
--- a/ravedude/Cargo.lock
+++ b/ravedude/Cargo.lock
@@ -123,6 +123,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "elf"
+version = "0.7.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e2b183d6ce6ca4cf30e3db37abf5b52568b5f9015c97d9fbdd7026aa5dcdd758"
+
+[[package]]
 name = "getrandom"
 version = "0.2.2"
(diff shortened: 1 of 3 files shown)


Rahix (commented):
Hi, thanks, this is a very good addition.  The only thing I am thinking to change is whether we should provide a commandline parameter to explicitly disable EEPROM writing.  I am thinking about use-cases where settings are stored in EEPROM and you want the option to not overwrite them on most uploads (yet still provide the `.eeprom` section for defaults). What's your take on this?

cr1901:
>What's your take on this?

Sure, I can add a flag to disable EEPROM flashing.

I think most Rust cases using the `.eeprom` section would want the EEPROM to be reflashed if e.g. `cargo` triggered a rebuild of the crate(s) providing an `.eeprom` section. But I can envision use cases where a Rust firmware is provided an `.eeprom` section as part of an C or asm object file/library, and you don't necessarily want to reflash the EEPROM every time.

cr1901:
Please do not merge yet, I believe this patch is incorrect/imprecise (it's not `.eeprom` section name that's special, it's "does this section's address and size match an AVR's EEPROM?").

I'll work on fixing it in the coming days.

PR #404, opened 2023-03-30
Add support for Arduino Duemilanove (328p version).

cr1901:
AIUI, the 328p version of the Duemilanove is essentially an Arduino Nano with a barrel jack. I want to add it because for better or worse, it is the only board I have :P (aside from a Sparkfun kit which only has ISP headers... btw, could `ravedude` in principle support "Arduino as ISP" or other programmers?).


ovo-Tim:
There still isn't an option of `328p` in [avr-hal-template](https://github.com/Rahix/avr-hal-template/tree/main), should I fix that and create a PR? :hugs:

PR #407, opened 2023-04-07
ravedude: Fix support for nix flakes on darwin

adrianchong518:
Previously, the `ravedude` would not install via nix flakes on Darwin / MacOS as the build dependencies unconditionally included `udev`, which is not available and not needed on Darwin. Other than this, nothing prevents it from being built and installed on Darwin.

But this is easily fixed with a check for the platform. So with this PR, `ravedude` can be installed via nix flakes on both Linux and Darwin.

### DIFF
--- a/ravedude/README.md
+++ b/ravedude/README.md
@@ -24,9 +24,7 @@ Next, install the latest version from crates.io with the following command:
 cargo install ravedude
 ```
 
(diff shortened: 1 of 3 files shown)


Rahix (approved):
Thanks!  Not a nix user myself so I can't check this, but I'll just go ahead and trust you that it works :)

PR #412, opened 2023-05-02
fix pullup on floating dyn pin

### LINKED ISSUE #411: Pin<Input<Floating>, Dynamic> has pullup enabled
After calling into_floating_input() on a dynamic pin the pullup is enabled

Example:
``` rust
fn main() -> ! {

    let mut pin3 = pins.d3.downgrade();
    pin3 = pin3.into_floating_input();
    loop {}
}
```

kabel42:
this fixes #411 on atmega168 for me


Rahix (approved):
Thanks so much for the fix!

PR #420, opened 2023-05-28
Added support for nano every uploading

bibble235:
This is to enable the Nano every to use the uploading provided in the arduino distribution.


raphaelahrens:
What is missing for this PR to be merged?

stappersg:
> What is missing for this PR to be merged?

Much better question: _What should be done to get this merged?_

Rahix:
The Arduino Nano Every uses the ATmega4809 microcontroller which is not yet supported by `avr-hal` unfortunately (see #97).  This PR seems to attempt specifying ATmega328P which unfortunately is not going to work...

The `ravedude` changes are interesting for the future when the support in `avr-hal` lands but unfortunately that has to happen first...

PR #204, opened 2021-07-28
Add ATtiny167 support and ADC support for all ATtiny chips

quentinmit:
Part two of #202 (includes the commit from #203)

PR #423, opened 2023-06-05
attiny85: ADC support

jleightcap:
I have this branch around from a long time ago, if it's of any use. To be honest, I had forgotten about this, so I may not have great rationale for changes here. 😅

### DIFF
--- a/mcu/attiny-hal/src/adc.rs
+++ b/mcu/attiny-hal/src/adc.rs
@@ -1,0 +1,105 @@
+//! Analog-to-Digital Converter
+
+use crate::port;
+pub use avr_hal_generic::adc::{AdcChannel, AdcOps, ClockDivider};
+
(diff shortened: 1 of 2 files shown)


kpcyrd:
It seems adc has been added in the meantime: https://github.com/Rahix/avr-hal/blob/main/mcu/attiny-hal/src/adc.rs

Rahix:
Hi,

sorry, yeah the ADC support in `attiny-hal` got included as a "side-effect" of #204 ...

I've checked the diff between this implementation and current main and it seems that `main` is pretty much a superset of this changeset by now... I guess the only good thing now is that I found a small typo due to this: 7fc791145f37 ("attiny-hal: Fix a typo")

Sorry for the mess :( @jleightcap, thanks a lot for your contribution anyway...

PR #424, opened 2023-06-09
Add panic example for arduino-nano

92kns:
Since the examples available in the `nano` are much less than the `uno`, I started messing around with some of the examples in the `uno` to try and get it to work.

Copying over the `uno-panic.rs` into `nano-panic.rs` and updating the `cargo.toml` file it runs as expected with
`cargo run --bin nano-panic`


Rahix (approved):
Thanks for the contribution!

PR #427, opened 2023-07-03
Upgrade to rust nightly 2023-08-08

Rahix:
With https://github.com/rust-lang/rust/pull/106619 merged, we can finally start moving forward with the toolchain again.  Upgrade to the latest version and see how things go.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "nightly-2022-07-10"
+channel = "nightly-2023-08-08"
 components = [ "rust-src" ]
 profile = "minimal"


Rahix:
Seems we're hitting https://github.com/rust-lang/compiler-builtins/issues/523 now.

Rahix:
Looking very good with `nightly-2023-08-08`.  Need to do some smoke tests still to make sure we're not missing any obvious regressions, but I guess the time has finally come where we can move to a new compiler!

@Patryk27 are you aware of anything speaking against moving to the new version at this point?  And in any case, thank you so much for your work on the compiler fixes time and time again!

Rahix:
Smoke tests all went well, let's be optimistic and merge this!

Patryk27:
> @Patryk27 are you aware of anything speaking against moving to the new version at this point? 

Nope, we should be good! (at least until something new comes up, as is the custom with llvm + avr :smile:)

PR #430, opened 2023-07-18
attiny-hal: add watchdog support

lopsided98:
This has been tested on the ATtiny85, but compiles for all the other parts and should work for them as well.


Rahix (approved):
Thanks!

Issue #425, opened 2023-06-20
Support Arduino Pro Mini (5V 16Mhz/3.3V 8Mhz)


tippfehlr:
This should be fairly straight forward because the Pro Mini 5V is exactly the same as the Arduino Nano, but without the onboard USB.

I use an Arduino Pro Mini 3.3V, and while flashing it with the Arduino Nano target (feature) works, I have to double all timers (eg. baud rate, millis) because the mcu runs at half clock speed.

norepimorphism:
(Not sure if GitHub already sends notifications for mentions, but I have created a PR at #435 for Pro Mini 5V support.)

tippfehlr:
Thanks @norepimorphism!

PR #435, opened 2023-09-04
Support the SparkFun ProMini 5V

norepimorphism:
This PR adds a new feature to *arduino-hal*, `sparkfun-promini-5v`, and a new board to ravedude, `promini-5v`. It is a copy, in nearly all respects, of the Arduino Nano implementation. I have used [this datasheet](https://cdn.sparkfun.com/assets/d/5/2/f/0/ProMini16MHzv2.pdf) as reference.

In the ravedude board config `SparkFunProMini5V` I  have implemented `guess_port` with a VID/PID pair I derived experimentally from my Pro Mini*, but I'm not sure if that is the correct approach.

Related issue: #425

*It is listed in <http://www.linux-usb.org/usb.ids> as "FT232 Serial (UART) IC" by FTDI.

### DIFF
--- a/ravedude/src/board.rs
+++ b/ravedude/src/board.rs
@@ -18,6 +18,7 @@ pub fn get_board(board: &str) -> Option<Box<dyn Board>> {
         "mega1280" => Box::new(ArduinoMega1280),
         "diecimila" => Box::new(ArduinoDiecimila),
         "promicro" => Box::new(SparkFunProMicro),
+        "promini-5v" => Box::new(SparkFunProMini5V),
         "trinket-pro" => Box::new(TrinketPro),
         "trinket" => Box::new(Trinket),
         "nano168" => Box::new(Nano168),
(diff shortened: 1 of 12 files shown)


norepimorphism:
An alternative to my implementation of the `sparkfun-promini-5v` feature in *arduino-hal* is to have it simply imply the `arduino-nano` feature (and I prefer this for the sake of maintenance). My belief is that the Pro Mini 5V is functionally an Arduino Nano sans an onboard FT232R.

ravedude/src/board.rs
```diff
@@ -316,6 +317,31 @@ impl Board for SparkFunProMicro {
...
+        avrdude::AvrdudeOptions {
+            programmer: "arduino",
+            partno: "atmega328p",
+            baudrate: Some(57600),
+            do_chip_erase: true,
+        }
+    }
+
+    fn guess_port(&self) -> Option<anyhow::Result<std::path::PathBuf>> {
+        Some(find_port_from_vid_pid_list(&[(0x0403, 0x6001)]))
```
Rahix, line 341: We can't assume that any old FTDI chip is a sparkfun pro mini 5v.  The `guess_port` should only be populated when there is a unique VID/PID that identifies this board without any doubt.

```suggestion
        Some(Err(anyhow::anyhow!("Not able to guess port")))
```

Rahix (changes requested):
Hi, so sorry for the delay here...

> An alternative to my implementation of the sparkfun-promini-5v feature in arduino-hal is to have it simply imply the arduino-nano feature

I think I actually prefer the way you've done things right now. This means that both boards are really and truly handled separately and a change introduced for one of them doesn't unintentionally affect the other.

I have one comment about the FTDI VID/PID, see below. Apart from that, everything else LGTM!

norepimorphism:
No worries, and sorry I didn't see your reply until now. I'll commit that change when I get home.

Rahix (approved):
Thanks!

PR #451, opened 2023-11-13
Use `paste!` to simplify port macros in avr-hal-generic

### LINKED ISSUE #449: Refactor most or all macros using `paste` to replace identifier-passing
Hello. I am currently trying to develop a Timer/Counter abstraction for a university project, and plan to contribute it back once done. While taking care of that, I noticed both existing macro code as well as my own (which I am modeling after the former) could be heavily simplified with [paste](https://docs.rs/paste/latest/paste). It is already a dependency of `avr-hal`, so would you accept a PR to refactor into using it for more than just Usart?

As an example, take `impl_port_traditional`. It's currently used like:
```rust
#[cfg(any(feature = "atmega48p", feature = "atmega168", feature = "atmega328p"))]
avr_hal_generic::impl_port_traditional! {
    enum Ports {
        PORTB: (crate::pac::PORTB, portb, pinb, ddrb),
        PORTC: (crate::pac::PORTC, portc, pinc, ddrc),
        PORTD: (crate::pac::PORTD, portd, pind, ddrd),
    }

    pub struct Pins {
        pb0: PB0 = (crate::pac::PORTB, PORTB, 0, portb, pinb, ddrb),
        pb1: PB1 = (crate::pac::PORTB, PORTB, 1, portb, pinb, ddrb),
        pb2: PB2 = (crate::pac::PORTB, PORTB, 2, portb, pinb, ddrb),
        pb3: PB3 = (crate::pac::PORTB, PORTB, 3, portb, pinb, ddrb),
        pb4: PB4 = (crate::pac::PORTB, PORTB, 4, portb, pinb, ddrb),
        pb5: PB5 = (crate::pac::PORTB, PORTB, 5, portb, pinb, ddrb),
        pb6: PB6 = (crate::pac::PORTB, PORTB, 6, portb, pinb, ddrb),
        pb7: PB7 = (crate::pac::PORTB, PORTB, 7, portb, pinb, ddrb),
        pc0: PC0 = (crate::pac::PORTC, PORTC, 0, portc, pinc, ddrc),
        pc1: PC1 = (crate::pac::PORTC, PORTC, 1, portc, pinc, ddrc),
        pc2: PC2 = (crate::pac::PORTC, PORTC, 2, portc, pinc, ddrc),
        pc3: PC3 = (crate::pac::PORTC, PORTC, 3, portc, pinc, ddrc),
        pc4: PC4 = (crate::pac::PORTC, PORTC, 4, portc, pinc, ddrc),
        pc5: PC5 = (crate::pac::PORTC, PORTC, 5, portc, pinc, ddrc),
        pc6: PC6 = (crate::pac::PORTC, PORTC, 6, portc, pinc, ddrc),
        pd0: PD0 = (crate::pac::PORTD, PORTD, 0, portd, pind, ddrd),
        pd1: PD1 = (crate::pac::PORTD, PORTD, 1, portd, pind, ddrd),
        pd2: PD2 = (crate::pac::PORTD, PORTD, 2, portd, pind, ddrd),
        pd3: PD3 = (crate::pac::PORTD, PORTD, 3, portd, pind, ddrd),
        pd4: PD4 = (crate::pac::PORTD, PORTD, 4, portd, pind, ddrd),
        pd5: PD5 = (crate::pac::PORTD, PORTD, 5, portd, pind, ddrd),
        pd6: PD6 = (crate::pac::PORTD, PORTD, 6, portd, pind, ddrd),
        pd7: PD7 = (crate::pac::PORTD, PORTD, 7, portd, pind, ddrd),
    }
}
```

With `paste`, I've been able to shorten to:
```rust
#[cfg(any(feature = "atmega48p", feature = "atmega168", feature = "atmega328p"))]
avr_hal_generic::impl_port_traditional! {
    enum Ports {
        B: crate::pac::PORTB = [0, 1, 2, 3, 4, 5, 6, 7],
        C: crate::pac::PORTC = [0, 1, 2, 3, 4, 5, 6],
        D: crate::pac::PORTD = [0, 1, 2, 3, 4, 5, 6, 7],
    }
}
```
So far, this generates equivalent (I'm tempted to say identical) code. It retains control over:
- Port letter/identifier;
- Register type in `pac`;
- Which pins are actually available for use (vs. reserved).

...but abstracts away the identifiers for fields and other types, which are constructed from concatenation via `paste`. This could be seen as a disadvantageous lack of granularity, but I don't think there should be much variation in the names used by the Peripheral Access Crates.

I'll try to push the preview implementation later today, to see if this could be useful.

LuigiPiucco:
Following #449, a draft of the reimplementation. Currently done is the port module and the simple_pwm module. I will be working on the others as well, and feel free to suggest changes/improvements. The intent is to deduplicate the implementations for each CPU, hoping to allow easier refactorings and additions of features. 

Closes #449.


Rahix:
Hi, thanks a lot for this work.  You have to be careful with simplifying the interface too much because in some places the chips themselves are not consistent enough to abstract them properly.  I believe this might be the cause for the CI failures here.

In general, my suggestion would be to do this work in one module per PR.  So please open one PR for the `port` module changes and a second one for the `simple_pwm` changes so each can be looked at and worked on in isolation.

LuigiPiucco:
> You have to be careful with simplifying the interface too much because in some places the chips themselves are not consistent enough to abstract them properly. I believe this might be the cause for the CI failures here.

Yes, I'm noticing the inconsistencies. For the port module they don't seem that bad, but the PWM one requires extra care, and probably the ones I haven't looked at yet as well. On that note, I should mention I couldn't fit Timer 1 on the Attiny85 into the module. It has a dual waveform and outputs to 2 pins simultaneously, so we probably can't reliably represent that with the current framework. I left it unimplemented with a TODO comment for later. It had an implementation prior, but it considered only one pin, which could lead to issues if the user expects to be able to use its dual normally.

> In general, my suggestion would be to do this work in one module per PR. So please open one PR for the port module changes and a second one for the simple_pwm changes so each can be looked at and worked on in isolation.

Okay, that's a good idea. I'll edit this one to have only the port module as scope, and open others as I proceed to other modules.

LuigiPiucco:
I'll mark this as ready for review, since we limited the scope. I rebased and fixed some issues in `arduino-hal` (I forgot to update some board definitions with the new macro). Also built lib for all CPUs and examples for all boards, it should be good now, but let's see if CI catches anything else.

As a side note, and since you thanked me for this, I should say this crate is very high quality, it sure beats the regular Arduino stack by a wide margin. So thank you as well.

avr-hal-generic/src/port.rs
```diff
@@ -523,169 +521,176 @@ macro_rules! impl_port_traditional {
...
-        pub struct Pins from $McuPins:ty {
-            $($(#[$pin_attr:meta])* pub $pin:ident: $Pin:ty = $pin_orig:ident,)+
+        pub struct Pins {
+            $($(#[$pin_attr:meta])* pub $pin_name:ident: $pin_type:ty = $pin_orig:ident,)+
+        }
+
+        impl Pins {
+            type Pin = $pin_wrapper:ident;
+            type McuPins = $mcu_pins:ty;
         }
```
Rahix, line 679: Can you comment on the motivation behind this change?  I don't have strong opinions either direction, but would like to hear why you prefer this macro syntax.

arduino-hal/src/port/mega.rs
```diff
@@ -1,340 +1,343 @@
...
         /// * `PCINT21`: External Interrupt (Pin Change)
         pub a13: atmega_hal::port::PK5 = pk5,
-        /// `A14`
+        /// `PUB A14`
         ///
         /// * `ADC14`: A/D converter input 14
         /// * `PCINT22`: External Interrupt (Pin Change)
         pub a14: atmega_hal::port::PK6 = pk6,
-        /// `A15`
+        /// `PUB A15`
```
Rahix, line 332: Looks like a lot of accidental changes in this file?

Rahix (changes requested):

avr-hal-generic/src/port.rs
LuigiPiucco: This is indeed a more stylistic change. Now that I think about it, I could probably have left this bit untouched, just changing the transcriber. But in general, when writing macros I prefer to keep syntax as close to the base language as possible, and the `Pins from $McuPins` part both introduces a "new" keyword (`from`) and looks more like a syntax for inheritance (in the OOP sense). The type that the macro generates doesn't really inherit the other, in the sense of "class A is a B and therefore can do what B can and be where B would", it's more like they're alternative interfaces. I thought making it seem as an associated type would be more fitting, because 1) it still shows the relationship between the types; 2) it doesn't add any syntax (though inherent associated types is an unstable grammar, I believe, so maybe we could improve); 3) it looks more native to Rust, since it doesn't have proper inheritance.

All that being said, it's mostly superfluous, so if you'd rather make the diff a bit smaller I can change it back.

arduino-hal/src/port/mega.rs
LuigiPiucco: Yes, this was the result of carefree regex substitution... 😆
I'm fixing it right away.

avr-hal-generic/src/port.rs
Rahix: Fair, thanks a lot for the detailed elaboration :)  Sounds good to me!

> All that being said, it's mostly superfluous, so if you'd rather make the diff a bit smaller I can change it back.

I think for the future, it would be a good idea to at least put each of the changes into its own commit as they don't logically relate to one another.  But let's not bother with such perfectionism here, I'll take your changes as they stand right now.

Rahix (approved):
Looks great now, thanks a lot again!

avr-hal-generic/src/port.rs
LuigiPiucco: > I think for the future, it would be a good idea to at least put each of the changes into its own commit as they don't logically relate to one another.

I can apply this suggestion, just have a question of preference before: Do you value the ability to compile each commit separately, or having it working by the end of the pull request is enough? There are some instances where using a more micro approach will require introducing changes that would not need to be there otherwise, to maintain self-containment. I'm fine with either, just would rather ask to make sure.

PR #459, opened 2023-12-06
attiny84a: add eeprom support

MichaelDarr:
## Overview

These changes add the ATtiny84a chip with EEPROM support.

## Status

These changes are blocked on [attiny84a support](https://github.com/Rahix/avr-device/pull/143) in [avr-device](https://github.com/Rahix/avr-device). Once that PR is merged, a new version is released, and the [avr-hal](https://github.com/Rahix/avr-hal) packages rely on that new version, these changes will be ready for proper review.

## Risks

I'm not very familiar with this library yet, and my confidence in the correctness of these changes is low. I've only tested these changes with an ATtiny84a - there may be unintentional side effects for other chips.

In one case, I had to include the config attribute `#[cfg(not(feature = "attiny84a"))]`. I assume there's a better option available, perhaps by implementing the `prescaler_n` functions for the Attiny84a. I am unsure how to do this - is there some documentation I'm missing?

## Testing

Locally, I've modified the relevant `Cargo.toml` files to target [the relevant `avr-hal` version](https://github.com/Rahix/avr-device/pull/143). Then, I was able to successfully read from and write to the EEPROM with the `avr_hal_generic` and `attiny_hal` libraries. I've included a simplified example below, and the full code can be found [here](https://github.com/MichaelDarr/liclock/tree/main/firmware) (this code in this repo is very messy - it's a proof of concept which will be fully rewritten).

```rust
use attiny_hal;
use avr_device::attiny84a;
use avr_hal_generic;

type Eeprom = avr_hal_generic::eeprom::Eeprom<attiny_hal::Attiny, attiny84a::EEPROM>;

#[avr_device::entry]
fn main() -> ! {
    let dp = attiny84a::Peripherals::take().unwrap();
    let eeprom = Eeprom::new(dp.EEPROM);

    let mut read_data = [0_u8; 16];
    if eeprom.read(0, &mut read_data).is_err() {
        panic!("eeprom read failed")
    }

    let mut write_data = [0_u8; 16];
    if eeprom.read(0, &mut write_data).is_err() {
        panic!("eeprom write failed")
    }
}
```

### DIFF
--- a/mcu/attiny-hal/src/adc.rs
+++ b/mcu/attiny-hal/src/adc.rs
@@ -68,6 +68,7 @@ pub mod channel {
     pub struct Temperature;
 }
 
+#[cfg(not(feature = "attiny84a"))]
 fn apply_clock(peripheral: &crate::pac::ADC, settings: AdcSettings) {
     peripheral.adcsra.write(|w| {
         w.aden().set_bit();
--- a/mcu/attiny-hal/Cargo.toml
+++ b/mcu/attiny-hal/Cargo.toml
@@ -8,6 +8,7 @@ edition = "2018"
 rt = ["avr-device/rt"]
 device-selected = []
 attiny84 = ["avr-device/attiny84", "device-selected"]
+attiny84a = ["avr-device/attiny84a", "device-selected"]
 attiny85 = ["avr-device/attiny85", "device-selected"]
 attiny88 = ["avr-device/attiny88", "device-selected"]
 attiny167 = ["avr-device/attiny167", "device-selected"]
--- a/mcu/attiny-hal/src/eeprom.rs
+++ b/mcu/attiny-hal/src/eeprom.rs
@@ -24,6 +24,17 @@ avr_hal_generic::impl_eeprom_attiny! {
     },
 }
 
+#[cfg(feature = "attiny84a")]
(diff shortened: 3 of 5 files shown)


mcu/attiny-hal/src/adc.rs
```diff
@@ -68,6 +68,7 @@ pub mod channel {
     pub struct Temperature;
 }
 
+#[cfg(not(feature = "attiny84a"))]
```
Rahix, line 71: So, this is a multi-level issue:

First of all, to not build the ADC module for this chip, it would be better to disable it at crate level.  Here:

https://github.com/Rahix/avr-hal/blob/21342dcace7184f01fdc4e9703b01197bd4b4b4f/mcu/attiny-hal/src/lib.rs#L65-L69

However, the chip does have an ADC so this code shouldn't fail to compile.  The reason it does is probably that registers in `avr-device` are not correctly defined.  I've already investigated this and opened https://github.com/Rahix/avr-device/pull/151 to fix it.

So I think the best course of action is to wait for the change to land in `avr-device` and then things should just work without the `#[cfg(not())]` line.

Rahix (commented):
Thanks a lot for the contribution and sorry for only getting to it now.  The code is already looking very good, for a few minor comments, check below.

PR #465, opened 2023-12-13
add note about `winget` to README (for windows installations)

rursprung:
[winget] is the official command-line package manager for Windows starting with Windows 10, thus there's no need to use 3rd party package managers anymore.
both `avr-gcc` and `avrdude` are packaged in winget.

installing them this way works fine and no further manual modifications of the environment are needed (tested in Git Bash and Powershell).

the note about scoop has been left in the README as i wasn't sure whether there were still any `var-hal` users left on older Windows releases, though at some point it can/should be removed for being obsolete.

[winget]: https://learn.microsoft.com/en-us/windows/package-manager/winget/

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,11 @@ Install dependencies:
   ```
 - Windows
 
-  Install [Scoop](https://scoop.sh/) using Powershell
+  Use [`winget`](https://learn.microsoft.com/en-us/windows/package-manager/winget/) on Windows 10 & Windows 11:
+  ```
+  winget install AVRDudes.AVRDUDE ZakKemble.avr-gcc
+  ```
+  On older systems you can use [Scoop](https://scoop.sh/) which you first have to install using Powershell:
   ```PowerShell
   Set-ExecutionPolicy RemoteSigned -Scope CurrentUser # Needed to run a remote script the first time
   irm get.scoop.sh | iex


rursprung:
i've also updated [the wiki page](https://github.com/Rahix/avr-hal/wiki/Setting-up-environment) accordingly.

stappersg:
On Wed, Dec 13, 2023 at 01:05:33AM -0800, Ralph Ursprung wrote:
> [winget] is the official command-line package manager for Windows
> starting with Windows 10, thus there's no need to use 3rd party package
> managers anymore.

That statement contradicts the provided link.


Regards
Geert Stappers

[winget]: https://learn.microsoft.com/en-us/windows/package-manager/winget/
-- 
Silence is hard to parse

rursprung:
> That statement contradicts the provided link.

huh, in what way?

stappersg:
On Thu, Dec 14, 2023 at 01:43:42AM -0800, Ralph Ursprung wrote:
> > That statement contradicts the provided link.
> 
> huh, in what way?

The statement: All fine starting with Windows 10

The provided link: "part of the App Installer. You can get App Installer from the Microsoft Store"
and "only supported on Windows 10 1709 (build 16299) or later"


Other problem I have with current proposed merge request
is that it claims position above existing solution and
(dis)qualifing that solution, `scoop`, to older systems.


Suggestion for next version of "Winget is the best!"

* Below   scoop.sh
* Windows 10 1709 (build 16299) or later
* "get App Installer from the Microsoft Store"
* Now you should have "winget"
* `winget install AVRDudes.AVRDUDE ZakKemble.avr-gcc`

 
Groeten
Geert Stappers
-- 
Silence is hard to parse

Rahix (approved):
I have no stakes in the windows package management world but I think this addition is entirely reasonable.  Thanks a lot for the contribution!

PR #473, opened 2023-12-23
enable GitHub Dependabot

rursprung:
this ensures that the dependencies are kept up to date. see [the docs] for further information.

[the docs]: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates


Rahix (approved):
Yeah, it's a good idea!  Don't know why I didn't add it previously because I did enable dependabot for all my other projects...

Thanks!

PR #479, opened 2023-12-28
update to Rust Edition 2021

rursprung:
this does not involve any code changes, it just sets the new edition, allowing to use newer features in the future.

as the crate is anyway only compiling with a recent nightly compiler there's no problem with requiring edition 2021 (it was first introduced in Rust 1.56).

note that running `cargo fix --edition` (as suggested by the [migration guide]) isn't possible as workspace-level cargo commands fail in this repository due to the feature flags :(

the resolver has been specified in the workspace due to the following build warning being issued if it's not present:

> warning: some crates are on edition 2021 which defaults to `resolver = "2"`, but virtual workspaces default to `resolver = "1"`
> note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
> note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest

[migration guide]: https://doc.rust-lang.org/edition-guide/editions/transitioning-an-existing-project-to-a-new-edition.html

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,3 +39,4 @@ exclude = [
     # The RAVEDUDE! Yeah!
     "ravedude",
 ]
+resolver = "2"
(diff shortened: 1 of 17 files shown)


Rahix (approved):
Thanks! :)

PR #480, opened 2023-12-28
update toolchain to `nightly-2023-12-28`

rursprung:
today Rust 1.75.0 has been released. this is the basis for the upcoming `embedded-hal` v1.0 release (at least for the `embedded-hal-async` traits). by updating to the latest available nightly build we ensure that it'll be possible to implement these traits in the future.

PS: that doesn't mean that i plan on implementing these traits 😉. i'm just laying the groundwork for whoever wants to do this in the future.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "nightly-2023-08-08"
+channel = "nightly-2023-12-28"
 components = [ "rust-src" ]
 profile = "minimal"


Rahix (approved):
Thanks!

Rahix:
Hm, after merging, CI magically stops passing - see https://github.com/Rahix/avr-hal/actions/runs/7493773898/job/20400220985#step:5:58 ...

Investigating...

PR #481, opened 2023-12-28
migrate `extern crate` to `pub use`

rursprung:
`extern crate` is from the old rust 2015 edition. since the 2018 edition `extern crate` is no longer needed.
here it was used to re-export certain crates (through `pub extern crate`) which are being referenced mainly in macros (which is why e.g. `paste` is re-exported).
this can also be achieved using `pub use`.

`avr-hal-generic` internally now has to use `crate::hal` in some cases where it could previously just use `hal` as the renaming is now bound to the usage in the crate and not to the imported crate. in most cases i've just replaced `hal::` usage with `embedded_hal_v0::` usage since it makes it more obvious which e-h version it is about (and in the future `hal` should either be removed or be replaced with e-h v1.0).

note: i've explicitly dropped the re-export of `ufmt` as it's not used in this repository and i don't see `ufmt` as an integral part of these crates. consumers can (and should) just add it as their own dependency. this is a breaking change for consumers which currently do something like `use avr_hal_generic::ufmt`.

for external consumers nothing changes besides the removed `ufmt` re-export.

### DIFF
--- a/avr-hal-generic/src/delay.rs
+++ b/avr-hal-generic/src/delay.rs
@@ -1,7 +1,7 @@
 //! Delay implementations
 
 use core::marker;
-use hal::blocking::delay as delay_v0;
+use embedded_hal_v0::blocking::delay as delay_v0;
 use embedded_hal::delay::DelayNs;
 
 #[cfg(all(target_arch = "avr", avr_hal_asm_macro))]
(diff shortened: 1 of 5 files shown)


rursprung:
it's just something i noticed. not sure whether this PR is really worth it... feel free to close it if you prefer to keep `extern crate` for the time being.
i can also re-add `ufmt` if you prefer.

Rahix:
:+1: on switching to `use`, let's get rid of all that `extern crate` legacy cruft.  Will come back to this PR after #482 is merged.

Rahix (approved):

PR #482, opened 2023-12-28
remove `void` dep, use `core::convert::Infallible` instead

rursprung:
there's no need for the `void` dependency anymore. accordingly it can be removed. this will also help with implementing `embedded-hal` v1 traits as some of them rely on (or at least have better support for) `core::convert::Infallible`.

this is a breaking change for consumers as they have to e.g. replace `void_unwrap()` calls with `unwrap()`.

the prelude is now no longer needed in most cases (potentally the whole `mod prelude` could be removed in the future as it now adds little benefit?).

note that this PR will conflict with #481; when one gets merged i'll have to rebase the other on top of it (though it will be a fairly trivial rebase). this is due to the fact that both modify the same lines in a few places. merging this one here first might be easier as the other one is smaller.

### DIFF
--- a/arduino-hal/Cargo.toml
+++ b/arduino-hal/Cargo.toml
@@ -34,10 +34,6 @@ ufmt = "0.2.0"
 version = "0.2.3"
 package = "embedded-hal"
 
-[dependencies.void]
-version = "1.0.2"
-default-features = false
-
 [dependencies.avr-hal-generic]
 path = "../avr-hal-generic/"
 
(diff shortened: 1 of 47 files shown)


Rahix (commented):
Yeah, it's time to drop `void`.  However in my eyes it isn't quite as simple as this.  I am really not a fan of switching to regular `.unwrap()` in all the example code.  Please check https://github.com/rust-embedded/embedded-hal/issues/329 for a detailed discussion of the reasons why.  This linked issue also has the solution I'd prefer to use in our examples: The [`unwrap-infallible`](https://crates.io/crates/unwrap-infallible) crate.  I think we should put it into the `avr-hal` prelude as the direct replacement for `ResultVoidExt`.

Rahix:
I took the liberty to add the unwrap_infallible here so it's part of the same PR :)

Issue #478, opened 2023-12-28
uno USART examples fail due to too high baudrate


rursprung:
the USART examples for the arduino uno fail when i try to run them. `ravedude` fails with the following error (note that the first two characters printed come from USART, though already the 2nd one is garbled):

> ```
> hYthread '<unnamed>' panicked at [...]\ravedude-0.1.6\src\console.rs:34:32:
> called `Result::unwrap()` on an `Err` value: Error { kind: InvalidData, message: "Windows stdio in console mode does not support writing non-UTF-8 byte sequences" }
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```

PuTTY doesn't crash but shows a lot of garbled text.

i narrowed it down to the baud rate of `57600`. if i switch this to `9600` (obviously both in the code instantiating the serial connection as well as the call to `ravedude` or in PuTTY) then the problem goes away.

do others have the same issue? i remember that i had this issue before with arduino uno, i think `75600 baud` is at the upper limit of what the hardware can handle and using a slower baudrate makes it work more consistently (though slower; but for something as simple as these examples that's irrelevant).

maybe it'd make sense to reduce the default baud rate for the examples?

Rahix:
Hm, I was even able to get 115200 baud working okay in the past IIRC so I'd expect this to not be too bad.  What board and MCU are you using here?

I think there is a second issue here: `ravedude` shouldn't crash in such situations.  We either need to simply remove the `unwrap()` here: https://github.com/Rahix/avr-hal/blob/f6b79dccbb36808b06f9c5d3e0050dc72ccc25cf/ravedude/src/console.rs#L34C20-L34C20 or implement some sort of input sanitization to ensure we only output correct UTF-8...

Rahix:
(And off-topic: Are you at 37c3 by any chance? :D)

rursprung:
> What board and MCU are you using here?

i'm using some obscure arduino uno clone called [gscheiduino](https://www.ehajo.de/dokuwiki/bausaetze:gscheiduino:anleitung) (the manufacturer is also discontinuing it, so i wouldn't recommend it to anyone 😉) which i got from university (i think due to supply chain issues they couldn't get the real arduino unos at the time) running with an atmega328p.

so yeah, maybe this is just an issue of this board?

> (And off-topic: Are you at 37c3 by any chance? :D)

no, but i wish you a lot of fun there! 😄

Rahix:
[...]

Issue #468, opened 2023-12-22
`embedded-hal@1.0.0` roadmap? 


onkoe:
Hey there! I recently went to port one of my driver crates to [the new `embedded-hal@1.0.0-rc.3`](https://docs.rs/embedded-hal/1.0.0-rc.3/embedded_hal/index.html), with its (honestly much nicer) APIs. However, I also wanted to check if my port worked on real hardware, so I grabbed an Arduino and searched for a matching alpha/rc/etc for this set of crates. 

To my surprise, there wasn't anything new that I could see, and checking for updates in the various `embedded` channels didn't get me very far. 🥹

Is there any plan to begin porting `avr-hal` to `embedded-hal@1.0.0`? The API is pretty stable now, and it's a good idea to get started before the release on [December 28th](https://github.com/rust-embedded/embedded-hal/issues/177#issuecomment-1821527247). 

If so, could you please link it to me? Otherwise, is this issue a good spot to start sticking down checkboxes? I'd at least like to progress on SPI, I2C, and other higher-level parts of the interface - though some other parts of `avr-hal-generic` are a bit beyond me. (I'm looking at you, `delay.rs`!) 😄

Rahix:
Initial trait implementations were done by @rursprung in #470!

If anyone is interested in picking up the SPI and I2C drivers, please feel free to do so :)

PR #484, opened 2023-12-28
implement `embedded-io` v0.6 `Read` & `Write`

rursprung:
with `embedded-hal` v1 the USART traits have been removed in favour of the new `embedded-io` crate.

this adds a (very basic) implementation for `Read` and `Write`. other traits (such as the `*Ready` or `BufRead` traits) have not (yet) been implemented and some (like `Seek`) probably can't be implemented for this HAL.

a better implementation might use a buffer in the background to receive more than one byte at once.

see also #249 for a related PR.

this is part of #468

note that this PR is based on top of #482 which has to be merged first. thus that commit is also shown here. for this PR you should only review the 2nd commit (and merge the other one first).
i've tested this with an Arduino Uno and the newly provided example (though running at 9600 baud, as outlined in #478).


rursprung:
this seems to have the same issue (at least on my device) as the existing implementation: for larger inputs it drops bytes when reading.

output from `uno-usart` (unchanged):

> ```
> Hello from Arduino!
> test
> Got 116!
> Got 101!
> Got 115!
> Got 10!
> foobar
> Got 102!
> Got 111!
> Got 111!
> Got 10!
> x
> Got 120!
> Got 13!
> Got 10!
> ```

output from `uno-usart-embedded-io`:

> ```
> Hello from Arduino!
> x
> Got [120] (which is 1 long)
> Got [13, 10] (which is 2 long)
> test
> Got [116] (which is 1 long)
> Got [101, 115, 10] (which is 3 long)
> q
> Got [113] (which is 1 long)
> Got [13, 10] (which is 2 long)
> ```

avr-hal-generic/src/usart.rs
```diff
@@ -360,16 +363,78 @@ impl<H, USART: UsartOps<H, RX, TX>, RX, TX, CLOCK> embedded_hal_v0::serial::Writ
...
 
     fn read(&mut self) -> nb::Result<u8, Self::Error> {
         self.p.raw_read()
     }
 }
 
+impl<H, USART: UsartOps<H, RX, TX>, RX, TX, CLOCK> embedded_io::Read for Usart<H, USART, RX, TX, CLOCK> {
+    fn read(&mut self, buf: &mut [u8]) -> Result<usize, Self::Error> {
+        // block for first byte
+        buf[0] = self.read_byte();
```
Rahix, line 413: Missing the treatment of `buf.len() == 0`.  From the trait docs:

> If `buf.len() == 0`, `read` returns without blocking, with either `Ok(0)` or an error. The `Ok(0)` doesn’t indicate EOF, unlike when called with a non-empty buffer.

avr-hal-generic/src/usart.rs
```diff
@@ -413,6 +478,14 @@ impl<H, USART: UsartOps<H, RX, TX>, RX, TX, CLOCK> UsartWriter<H, USART, RX, TX,
...
         }
     }
+
+    /// Transmit a byte.
+    ///
+    /// This method will block until the byte has been enqueued for transmission but **not** until
+    /// it was entirely sent.
+    fn write_byte(&mut self, byte: u8) {
+        nb::block!(self.p.raw_write(byte)).unwrap()
+    }
```
Rahix, line 488: Unsure whether it wouldn't be more readable to simply call ` nb::block!(self.p.raw_write(byte))` in the `write()` implementation directly to remove the additonal layer.

examples/arduino-uno/src/bin/uno-usart-embedded-io.rs
```diff
@@ -0,0 +1,29 @@
...
+use embedded_io::{Read, Write};
+
+fn usart_handler(serial: &mut (impl Read + Write)) -> ! {
+    serial.write_all("Hello from Arduino!\r\n".as_bytes()).unwrap();
+
+    loop {
+        let mut rx_buf: [u8; 16] = [0; 16];
+        let len = serial.read(&mut rx_buf).unwrap();
+
+        writeln!(serial, "Got {:?} (which is {} bytes long)", &rx_buf[..len], len).unwrap();
```
Rahix, line 18: The problem with this code is that `read()` won't every return more than a few bytes.  Immediately after receiving them, the `writeln!()` will block for a long time to spell out the response.  During this time, we probably receive more data but can't handle it as we're still busy and the hardware doesn't have a (large enough) buffer.

I'm trying to think what a better example could look like.  Maybe waiting for a newline?  Or waiting for a large enough time of silence?  I'd choose whatever variant is easier to implement, so probably the newline...

Rahix (changes requested):
> is there anything specific from my side which you need for the review?

No, just my horrible response latency at work again :(

In any case, thanks a lot for the contribution!  Please check my comments below.

PR #488, opened 2024-01-09
update to `embedded-hal` v1.0.0 (no more RC!)

rursprung:
it finally happened: the stable release of `embedded-hal` has just been released.


Rahix (approved):
:tada: 

Thanks!

PR #445, opened 2023-11-06
feat: Support the SparkFun ProMini 3V

emilbonnek:
This PR adds a new feature to arduino-hal, `sparkfun-promini-3v`, and a new board to ravedude, `promini-3v`. It is a copy, in nearly all respects, of the 5v implementation. The only difference is the clock speed of 8 MhZ.

EDIT: I just learned that the 3v board is actually 3.3v so maybe it should be named `promini-3v3` instead for clarification?

PR #491, opened 2024-01-14
feat: add support for Sparkfun ProMini 3.3V 8Mhz (sparkfun-promini-3v3)

tippfehlr:
This PR adds support for the ProMini 3.3V (8Mhz) (`sparkfun-promini-3v3`).
Apart from `arduino-hal/src/clock.rs` it is entirely copied from #435.

~~Because `sparkfun-promini-3.3v` is not a valid name, I left out the `.3`.~~
~~Another option would be to rename both the 5V and the 3.3V version to 16Mhz and 8Mhz respectively, eg.~~
~~- `sparkfun-promini-16mhz`~~
~~- `sparkfun-promini-8mhz`~~

~~Both variants are fine in my opinion, however renaming `5v` to `16mhz` would be a breaking change.~~

Original issue: #425


stappersg:
[This comment provides a link to a similar merge request](https://github.com/Rahix/avr-hal/pull/491)

tippfehlr:
@stappersg You linked to this issue, not #445, which I assume you meant.

True, I didn’t check for existing *PRs*. However, because this is mergeable as is, I’ll leave it open.
I also changed the name to `sparkfun-promini-3v3` because that it’s more appropriate.

https://github.com/Rahix/avr-hal/pull/445#issuecomment-1815774078: As I understand this the ProMini 3.3V ***is*** an Arduino (328P) clocked down, meaning it is only appropriate to handle it as such.

Rahix (approved):
(Rebased on `main`)

Thanks for the contribution!

Issue #490, opened 2024-01-12
No way to switch an analog pin back to digital mode


QuinnFreedman:
If I set an pin to a digital input/output, I can later change it to a different input/output config:

```rust
let dp = arduino_hal::Peripherals::take().unwrap();
let pins = arduino_hal::pins!(dp);
let a0_input = pins.a0.into_pull_up_input();
a0_input.is_low();                      // use pin
let a0_output = a0_input.into_output();
a0_output.set_high();                   // use pin as output
```

But, if I change it to an analog input, I cant figure out how to use it as an output afterwards:

```rust
let dp = arduino_hal::Peripherals::take().unwrap();
let pins = arduino_hal::pins!(dp);
let mut adc = arduino_hal::Adc::new(dp.ADC, Default::default());
let a0 = pins.a0.into_analog_input(&mut adc);

// Error: the method `into_output` exists for struct `Pin<Analog, PC0>`, but its trait bounds were not satisfied
let a0_output = a0.into_output();
```

Am I missing something? Is there a different way to read from a pin and get it back?

My only current workaround is to start an `unsafe` block, get a second copy of `arduino_hal::Peripherals` with `steal()`, and use a second copy of the pin. But that is not a great solution for something that should be safe to do.

stappersg:
On Fri, Jan 12, 2024 at 03:39:56PM -0800, Quinn Freedman wrote:
> 
> Am I missing something?
> 

I am missing real life schematic of PCB where the analog input
also can be used as digital output.


> My only current workaround is to start an `unsafe` block, get a second
> copy of `arduino_hal::Peripherals` with `steal()`, and use a second
> copy of the pin. But that is not a great solution for something that
> should be safe to do.

Please tell about the electronics that is on the embedded appliance.


Groeten
Geert Stappers
-- 
Silence is hard to parse

Rahix:
@stappersg this is entirely besides the point.  There are certainly usecases where a pin is used with the ADC and digital I/O at the same time.

Rahix:
> Am I missing something? Is there a different way to read from a pin and get it back?

[...]

PR #493, opened 2024-01-16
Add an into_digital method to analog input pins

QuinnFreedman:
Adds a way to get a pin back into digital mode after using it as an analog input re: our conversation in #490.

```rust
let mut adc = arduino_hal::Adc::new(dp.ADC, Default::default());
let a5 = pins.a5.into_analog_input(&mut adc);

let floating_voltage = a5.analog_read(&mut adc);

let mut my_button = a5.into_digital(&mut adc).into_pull_up_input();

if my_button.is_low() {
    // ...
}
```

Confirmed working on ATmega328p Arduino Nano.

It's a little awkward that you have to pass in `&adc` to set the pin back to digital mode. For that mater, it's a little awkward that you have to do `pin.into_analog_input(&mut adc).analog_read(&mut adc);` to read from it. But I didn't see another way to do it without affecting runtime performance and/or getting into borrow checker hell. Let me know if you have another idea.

### DIFF
--- a/avr-hal-generic/src/port.rs
+++ b/avr-hal-generic/src/port.rs
@@ -207,11 +207,12 @@ impl<PIN: PinOps, MODE: mode::Io> Pin<MODE, PIN> {
         ADC: crate::adc::AdcOps<H>,
         CLOCK: crate::clock::Clock,
     {
-        let new = Pin {
+        let mut new = Pin {
             pin: self.pin,
             _mode: PhantomData,
         };
         adc.enable_pin(&new);
+        unsafe { new.pin.make_input(false) };
         new
     }
 }
@@ -536,7 +537,7 @@ impl<PIN: PinOps> Pin<mode::Analog, PIN> {
     /// input. You can get to other digital modes by calling one of the usual `into_...` methods
     /// on the return value of this function.
     pub fn into_digital<H, ADC, CLOCK>(
-        mut self,
+        self,
         adc: &mut crate::adc::Adc<H, ADC, CLOCK>,
     ) -> Pin<mode::Input<mode::Floating>, PIN>
     where
@@ -544,7 +545,6 @@ impl<PIN: PinOps> Pin<mode::Analog, PIN> {
         ADC: crate::adc::AdcOps<H>,
         CLOCK: crate::clock::Clock,
     {
-        unsafe { self.pin.make_input(false) };
         adc.disable_pin(&self);
         Pin {
             pin: self.pin,
--- a/avr-hal-generic/src/port.rs
+++ b/avr-hal-generic/src/port.rs
@@ -529,6 +529,28 @@ impl<PIN: PinOps> Pin<mode::Analog, PIN> {
     {
         crate::adc::Channel::new(self)
     }
+
+    /// Convert this pin to a floating digital input pin.
+    ///
+    /// The pin is re-enabled in the digital input buffer and is no longer usable as an analog
+    /// input. You can get to other digital modes by calling one of the usual `into_...` methods
+    /// on the return value of this function.
+    pub fn into_digital<H, ADC, CLOCK>(
+        mut self,
+        adc: &mut crate::adc::Adc<H, ADC, CLOCK>,
+    ) -> Pin<mode::Input<mode::Floating>, PIN>
+    where
+        Pin<mode::Analog, PIN>: crate::adc::AdcChannel<H, ADC>,
+        ADC: crate::adc::AdcOps<H>,
+        CLOCK: crate::clock::Clock,
+    {
+        unsafe { self.pin.make_input(false) };
+        adc.disable_pin(&self);
+        Pin {
+            pin: self.pin,
+            _mode: PhantomData,
+        }
+    }
 }
 
 #[macro_export]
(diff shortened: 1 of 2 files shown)


QuinnFreedman:
@Rahix good question. As things currently stand, it does seem to be necessary. Without it, the pin will revert to whatever state it was in before it was used as analog.

In testing this, I came across another potential issue. If you set a pin to output and write it HIGH, and then use `into_analog_input()`, it stays outputting HIGH, even with an external pulldown resistor. `analog_read()` will always give ~5v. `into_analog_input()` sets the `DIDR` (digital _input_ disable) register, which fixes the `PINC` (digital input register) to logical 0, but it does not disable _output_; it does not clear the `DDRC` (data direction) and `PORTC` (output/pullup-enable/disable) registers.

I could maybe imagine a scenario where you would want to analog read while outputting logic 1 (maybe for calibration?) but IMO this behavior is unclear and likely a bug. If we want to support analog reading while digital writing maybe that should be a function with a different name, or just force people to use direct port manipulation in that case.

If we set the pin to floating input in `into_analog_input()`, then it would not be necessary to do so in `into_digital()`.


For reference, here is my test program:

```rust
fn report_portc(
    serial: &mut arduino_hal::Usart<
        arduino_hal::pac::USART0,
        arduino_hal::port::Pin<arduino_hal::port::mode::Input, arduino_hal::hal::port::PD0>,
        arduino_hal::port::Pin<arduino_hal::port::mode::Output, arduino_hal::hal::port::PD1>,
    >,
    label: &str,
) {
    uwriteln!(serial, "\n{}:", label).unwrap_infallible();
    let dp = unsafe { arduino_hal::Peripherals::steal() };
    let ddrc = dp.PORTC.ddrc.read().bits();
    uwriteln!(serial, "DDRC: {}", ddrc).unwrap_infallible();
    let pinc = dp.PORTC.pinc.read().bits();
    uwriteln!(serial, "PINC: {}", pinc).unwrap_infallible();
    let portc = dp.PORTC.portc.read().bits();
    uwriteln!(serial, "PORTC: {}", portc).unwrap_infallible();
}

#[arduino_hal::entry]
fn main() -> ! {
    let mut adc = arduino_hal::Adc::new(dp.ADC, Default::default());
    let mut led = pins.d13.into_output_high();

    report_portc(&mut serial, "Initial state");

    let a5_output = pins.a5.into_output_high();

    report_portc(&mut serial, "After set digital output");

    arduino_hal::delay_ms(3000);

    led.set_low();
    let a5_analog = a5_output.into_analog_input(&mut adc);

    report_portc(&mut serial, "After set analog input");

    arduino_hal::delay_ms(3000);

    uwriteln!(
        &mut serial,
        "Analog read: {}",
        a5_analog.analog_read(&mut adc)
    )
    .unwrap_infallible();

    let a5_digital = a5_analog.into_digital(&mut adc);
    report_portc(&mut serial, "After set digital input");

    uwriteln!(&mut serial, "Digital read: {}", a5_digital.is_high()).unwrap_infallible();

    loop {
        led.set_high();
        arduino_hal::delay_ms(300);
        led.set_low();
        arduino_hal::delay_ms(300);
    }
}
```

Output (current version)
```
Initial state:
DDRC: 0
PINC: 0
PORTC: 0

After set digital output:
DDRC: 32
PINC: 32
PORTC: 32

After set analog input:
DDRC: 32
PINC: 0
PORTC: 32

Analog read: 989                // Always reads near max, even with external pulldown resistor

After set digital input:
DDRC: 0
PINC: 0
PORTC: 0

Digital read: false             // Digital read works "correctly" (depends on external connection)
```


Output (without `make_input(false)`)
```
Initial state:
DDRC: 0
PINC: 0
PORTC: 0

After set digital output:
DDRC: 32
PINC: 32
PORTC: 32

After set analog input:
DDRC: 32
PINC: 0
PORTC: 32

Analog read: 989                // Always reads near max, even with external pulldown resistor

After set digital input:
DDRC: 32
PINC: 32
PORTC: 32

Digital read: true              // Always reads HIGH, even with external pulldown resistor
```

QuinnFreedman:
I pushed a change that moves that reset into the `into_analog_input` method. That makes the most sense to me but if you don't want that behavior we can just revert it.

The new output is:
```
Initial state:
DDRC: 0
PINC: 0
PORTC: 0

After set digital output:
DDRC: 32
PINC: 32
PORTC: 32

After set analog input:
DDRC: 0
PINC: 0
PORTC: 0
Analog read: 0

After set digital input:
DDRC: 0
PINC: 0
PORTC: 0
Digital read: false
```

avr-hal-generic/src/port.rs
```diff
@@ -207,11 +207,12 @@ impl<PIN: PinOps, MODE: mode::Io> Pin<MODE, PIN> {
         ADC: crate::adc::AdcOps<H>,
         CLOCK: crate::clock::Clock,
     {
-        let new = Pin {
+        let mut new = Pin {
             pin: self.pin,
             _mode: PhantomData,
         };
         adc.enable_pin(&new);
+        unsafe { new.pin.make_input(false) };
```
Rahix, line 215: Thanks so much for digging into this.  I agree that this is the sensible thing to do and arguably, it should have always been this way.

PR #494, opened 2024-01-19
Fix eeprom write and erase

Eggtosch:
Hey, i authored the Atmega164PA support at work with my colleague (see https://github.com/Rahix/avr-hal/pull/452).
I took some spare time to look into the EEPROM issue once again (https://github.com/Rahix/avr-hal/issues/406):

The various datasheets for the atmega* chips essentially state the following procedure :
1. Wait for the EEPE bit to become zero.
2. Write a logical one to the EEMPE bit while writing a zero to EEPE in EECR.
3. Within four clock cycles after setting EEMPE, write a logical one to EEPE.

While we do step 1 and 2 correctly, for step 3, we use `self.eecr.write(|w| w.eepe().set_bit());` which sets the EEPE bit, but also sets the rest of the register to zero.
This seems to cause some slightly different behavior on different boards, e.g. on my Arduino Uno (Atmega328p) on which i tested this, only a 0xff write fails, while on the atmega164pa an erase also fails.

Using `self.eecr.modify(|_, w| w.eepe().set_bit());` fixes the issue, because it preserves the rest of the register.

Please let me know what you think :).

### DIFF
--- a/avr-hal-generic/src/eeprom.rs
+++ b/avr-hal-generic/src/eeprom.rs
@@ -190,14 +190,14 @@ macro_rules! impl_eeprom_common {
                                 let $periph_ewmode_var = &self;
                                 $set_erasewrite_mode
                             }
-                            self.eecr.write(|w| w.eepe().set_bit()); // Start Erase+Write operation.
+                            self.eecr.modify(|_, w| w.eepe().set_bit()); // Start Erase+Write operation.
                         } else {
                             // Now we know that all bits should be erased.
                             {
                                 let $periph_emode_var = &self;
                                 $set_erase_mode
                             }
-                            self.eecr.write(|w| w.eepe().set_bit()); // Start Erase-only operation.
+                            self.eecr.modify(|_, w| w.eepe().set_bit()); // Start Erase-only operation.
                         }
                     }
                     //Now we know that _no_ bits need to be erased to '1'.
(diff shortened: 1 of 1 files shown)


Eggtosch:
Little update: I now have a better in detail explanation :).

Using `self.eecr.write(|w| w.eepe().set_bit());` sets all bits except the `eepe` bit to zero. This does not lead to the eeprom operation failing, in fact it sets the eeprom operation to erase+write. This leads to different behaviors depending on the case that is entered in the `raw_write_byte` function.

1. Some bits need to be programmed to 0 and some to 1: The erase+write mode is used and the data register is set, everything works as expected.
2. All changed bits need to be set to 0: The write mode is used and the data register is set, everything works as expected.
3. All changed bits need to be set to 1: The erase mode is used, but the data register is not set (of course, we normally wouldn't need the data register). But the bug then sets the erase+write mode instead and the data from the data register is used, in this case the value that was read at the beginning of the function, so the same value that is stored at the address is written again.

For the `raw_erase_byte` function, the data register also isn't set and whatever was written there before is used. The function worked for the `atmega328p` in my tests only because of my test setup, i was reading a `0xff` from a different address before using the `erase_byte` function.

Rahix (approved):
Very nice work, thanks a lot for digging into the issue!  This certainly is much better than the workaround ;)

PR #495, opened 2024-01-19
Add atmega168 feature to target ATmega168 chip running with its internal 8MHz oscillator

ghismary:
I'm working on a project based on an ATmega168 chip without an external oscillator and therefore using its own internal 8MHz oscillator.
These changes introduce a new feature to be able to target this chip.

However, the changes made in the PR for adding an internal-clock feature (https://github.com/Rahix/avr-hal/pull/443) may be a better way to achieve what I want. With this I could use both the 'nano168' and 'internal-clock' features to get the same result.


Rahix:
You are using the ATmega168 directly, instead of a prebuilt Arduino/SparkFun/etc. dev-board, right?  For such situations, you shouldn't use `arduino-hal` and instead depend on `atmega-hal` directly.  That way, you are not bound to a specific clock selection or similar.

Fortunately for you, someone is currently contributing examples of how to use `atmega-hal` directly.  See PR #524.

PR #496, opened 2024-01-19
Implement the I2c<SevenBitAddress> trait from embedded-hal 1.0.

ghismary:
Here is the implementation of the I2c<SevenBitAddress> trait from embedded-hal 1.0 allowing to pass an I2C device to a driver based on embedded-hal 1.0.

### DIFF
--- a/avr-hal-generic/src/i2c.rs
+++ b/avr-hal-generic/src/i2c.rs
@@ -2,6 +2,8 @@
 //!
 //! Check the documentation of [`I2c`] for details.
 
+use embedded_hal::i2c::SevenBitAddress;
+
 use crate::port;
 use core::marker::PhantomData;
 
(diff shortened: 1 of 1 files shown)


avr-hal-generic/src/i2c.rs
```diff
@@ -388,6 +412,42 @@ impl<H, I2C: I2cOps<H, SDA, SCL>, SDA, SCL, CLOCK> embedded_hal_v0::blocking::i2
...
+                embedded_hal::i2c::Operation::Write(bytes) => {
+                    if idx == 0 || previous_direction != Direction::Write {
+                        self.p.raw_start(address, Direction::Write)?;
+                    }
+                    self.p.raw_write(bytes)?;
+                    previous_direction = Direction::Write;
+                }
+            }
+        }
+        if nb_operations > 0 {
```
Rahix, line 443: I think it's cleaner to drop the `nb_operations` variable and directly do

```rust
if operations.len() > 0 {
```

here.

Rahix (commented):
Hi, thanks a lot for the contribution.  The implementation looks very good, I only have one minor nit (see below).

How much did you test this on real hardware?  I'd be especially interested in knowing how well transactions with multiple operations work.  I don't have hardware on me to test it myself right now, so I'll rely on what results you can provide :)

avr-hal-generic/src/i2c.rs
ghismary: Hi,
I made this little change.

ghismary:
> Hi, thanks a lot for the contribution. The implementation looks very good, I only have one minor nit (see below).
> 
> How much did you test this on real hardware? I'd be especially interested in knowing how well transactions with multiple operations work. I don't have hardware on me to test it myself right now, so I'll rely on what results you can provide :)

I am using it in a small project using several device drivers I wrote and that are based on it, eg. [embedded-sht3x](https://crates.io/crates/embedded-sht3x), [embedded-sgp30](https://crates.io/crates/embedded-sgp30) and [embedded-qmp6988](https://crates.io/crates/embedded-qmp6988). Actually, I made some changes that I have not pushed yet in these drivers to limit the use of multiple operations, because they work perfectly well with avr-hal, but I'm also using them on an other hardware via linux-embedded-hal where the use of multiple operations is causing problems. From my research to understand why this is no working well there, the issue comes directly from the Linux kernel, at least of the version of the kernel running on my hardware. So I chose to modify my drivers for them to work correctly on all the hardwares I am targetting.

Rahix:
> Actually, I made some changes that I have not pushed yet in these drivers to limit the use of multiple operations, because they work perfectly well with avr-hal, but I'm also using them on an other hardware via linux-embedded-hal where the use of multiple operations is causing problems.

Haha, ouch!  Is this an issue known to the `linux-embedded-hal` project?  If not, you really should tell them...

Rahix (approved):

PR #497, opened 2024-01-19
Use toggle from embedded-hal 1.0 in the uno blink example.

ghismary:
As embedded-hal 1.0 has now been released, make the changes that were suggested in a comment in the uno blink example.


Rahix (commented):
Hi, thanks for the contribution!  Please check the CI error, I think you missed changing the import...

ghismary:
Hi, you're right, I fixed that.

Rahix:
Thanks!

PR #501, opened 2024-01-28
generic: port: Fix toggle() for dynamic pins

### LINKED ISSUE #499: "uno-timer" example no longer works

Rahix:
During commit 063f845a0940 ("generic: simplify port macro via `paste!`"), an accidental change was introduced where out_toggle() for dynamic pins wrote to the `PORT` register instead of the `PIN` register. This meant that toggle() for dynamic pins no longer worked.

Fix this regression by making out_toggle() write to `PIN` again.

Fixes #499.

### DIFF
--- a/avr-hal-generic/src/port.rs
+++ b/avr-hal-generic/src/port.rs
@@ -644,7 +644,7 @@ macro_rules! impl_port_traditional {
                 #[inline]
                 unsafe fn out_toggle(&mut self) {
                     match self.port {
-                        $(DynamicPort::[<PORT $name>] => (*<$port>::ptr()).[<port $name:lower>].write(|w| {
+                        $(DynamicPort::[<PORT $name>] => (*<$port>::ptr()).[<pin $name:lower>].write(|w| {
                             w.bits(self.mask)
                         }),)+
                     }


Rahix (commented):
Verified on ATmega328P.

andber1:
@Rahix While debugging a toggle problem on an AtMega16 (see #642) I came across this PR. I did not understand how the `w.bits(self.mask)` toggles a bit. Shouldn't it be `w.bits(r.bits() ^ self.mask)`? And why is the PIN register the right one? Doesn't the PORT register control the output state? I need to change the code in my PR exactly the opposite way as you did and I am not sure why these two places are so different.

PR #503, opened 2024-01-30
Implement the SpiBus trait from embedded-hal 1.0.

ghismary:
Here is the implementation of the embedded-hal 1.0 traits for SPI.
I need to do further testing, but from what I have seen up to now, it seems to work quite well.

### DIFF
--- a/avr-hal-generic/src/spi.rs
+++ b/avr-hal-generic/src/spi.rs
@@ -1,7 +1,8 @@
 //! SPI Implementation
 use crate::port;
+use core::cmp::Ordering;
 use core::marker::PhantomData;
-use embedded_hal_v0::spi;
+use embedded_hal::spi;
 
 /// Oscillator Clock Frequency division options.
 ///
@@ -75,6 +76,10 @@ pub trait SpiOps<H, SCLK, MOSI, MISO, CS> {
     fn raw_check_iflag(&self) -> bool;
     fn raw_read(&self) -> u8;
     fn raw_write(&mut self, byte: u8);
+
+    fn raw_read_buf(&self, buf: &mut [u8]);
+    fn raw_write_buf(&mut self, buf: &[u8]);
+    fn raw_read_write_buf(&mut self, buffer: &mut [u8], bytes: &[u8]);
 }
 
 /// Wrapper for the CS pin
@@ -106,11 +111,39 @@ impl<CSPIN: port::PinOps> embedded_hal_v0::digital::v2::StatefulOutputPin for Ch
     }
 }
 
-impl<CSPIN: port::PinOps> embedded_hal_v0::digital::v2::ToggleableOutputPin for ChipSelectPin<CSPIN> {
+impl<CSPIN: port::PinOps> embedded_hal_v0::digital::v2::ToggleableOutputPin
+    for ChipSelectPin<CSPIN>
+{
     type Error = core::convert::Infallible;
     fn toggle(&mut self) -> Result<(), Self::Error> {
         self.0.toggle();
         Ok(())
+    }
+}
+
+impl<CSPIN: port::PinOps> embedded_hal::digital::ErrorType for ChipSelectPin<CSPIN> {
+    type Error = core::convert::Infallible;
+}
+
+impl<CSPIN: port::PinOps> embedded_hal::digital::OutputPin for ChipSelectPin<CSPIN> {
+    fn set_high(&mut self) -> Result<(), Self::Error> {
+        self.0.set_high();
+        Ok(())
+    }
+
+    fn set_low(&mut self) -> Result<(), Self::Error> {
+        self.0.set_low();
+        Ok(())
+    }
+}
+
+impl<CSPIN: port::PinOps> embedded_hal::digital::StatefulOutputPin for ChipSelectPin<CSPIN> {
+    fn is_set_high(&mut self) -> Result<bool, Self::Error> {
+        Ok(self.0.is_set_high())
+    }
+
+    fn is_set_low(&mut self) -> Result<bool, Self::Error> {
+        Ok(self.0.is_set_low())
     }
 }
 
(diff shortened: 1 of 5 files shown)


avr-hal-generic/src/spi.rs
```diff
@@ -68,13 +72,17 @@ impl Default for Settings {
 /// This trait defines the common interface for all SPI peripheral variants.  It is used as an
 /// intermediate abstraction ontop of which the [`Spi`] API is built.  **Prefer using the
 /// [`Spi`] API instead of this trait.**
-pub trait SpiOps<H, SCLK, MOSI, MISO, CS> {
+pub trait SpiOps<H, SCLK, MOSI, MISO> {
```
Rahix, line 75: You can't drop the CS pin here, as it has meaning to the hardware.  You can read up on the details in https://github.com/Rahix/avr-hal/issues/27.  We had to add this rather inflexible CS abstraction to ensure you can't cause unintentional side effects by misusing the CS pin.

avr-hal-generic/src/spi.rs
```diff
@@ -282,12 +318,301 @@ where
...
+pub struct SpiDevice<H, SPI, SCLKPIN, MOSIPIN, MISOPIN, CSPIN, CLOCK> {
+    p: SPI,
+    sclk: port::Pin<port::mode::Output, SCLKPIN>,
+    mosi: port::Pin<port::mode::Output, MOSIPIN>,
+    miso: port::Pin<port::mode::Input, MISOPIN>,
+    write_in_progress: bool,
+    cs: ChipSelectPin<CSPIN>,
+    delay: delay::Delay<CLOCK>,
+    _h: PhantomData<H>,
+}
```
Rahix, line 475: `SpiDevice` should only be implemented by a HAL when you can make use of hardware CS handling (See https://docs.rs/embedded-hal/latest/embedded_hal/spi/index.html#for-hal-authors).  As you are just taking an arbitrary pin here and calling `set_high()`, `set_low()`, there is no advantage of providing this `SpiDevice` implementation over the `ExclusiveDevice` from [`embedded-hal-bus`](https://docs.rs/embedded-hal-bus/latest/embedded_hal_bus/spi/index.html).

Rahix (commented):
Hi, thanks a lot for picking up this work!  I have two preliminary comments, see below.

ghismary:
Hi, thanks for your comments.
I had not seen this stuff about hardware CS management, and had not taken a look to embedded-hal-bus yet.
It makes sense, I'll rework on that.

avr-hal-generic/src/spi.rs
```diff
@@ -290,6 +326,70 @@ where
...
+        let write = unsafe {
+            let p = buffer.as_ptr();
+            core::slice::from_raw_parts(p, buffer.len())
+        };
+        self.p.raw_read_write_buf(buffer, write);
+        Ok(())
+    }
+
+    fn flush(&mut self) -> Result<(), Self::Error> {
+        if self.write_in_progress {
```
Rahix, line 385: Just to make sure I understand this correctly:

You are only making use of the `write_in_progress` flag here to ensure correct behavior when both `e-h0.2::FullDuplex` and `e-h1.0::SpiBus` are used at the same time, right?

All the other methods of `e-h1.0::SpiBus` synchronize with the end of the transfer internally and thus they don't need to set `write_in_progress`.  Is that correct?

And related:  What if `write_in_progress` is set when calling e.g. `SpiBus::write()`?  Shouldn't `SpiBus::write()` call `<Self as SpiBus>::flush()` before operating on the bus to ensure any `FullDuplex` transfer from before has actually finished?

avr-hal-generic/src/spi.rs
```diff
@@ -370,6 +470,39 @@ macro_rules! impl_spi {
...
+                assert_eq!(buffer.len(), bytes.len());
+
+                let mut buffer_it = buffer.iter_mut();
+                let mut bytes_it = bytes.iter();
+                while let Some(byte) = bytes_it.next() {
+                    self.raw_write(*byte);
+                    while self.spsr.read().spif().bit_is_clear() {}
+                    let mut data = buffer_it.next().unwrap();
+                    *data = self.raw_read();
+                }
```
Rahix, line 504: Another option would be to zip the two iterators:

```rust
for (byte, data) in bytes.iter().zip(buffer.iter_mut()) {
    self.raw_write(*byte);
    while self.spsr.read().spif().bit_is_clear() {}
    *data = self.raw_read();
}

File: mcu/atmega-hal/src/spi.rs [hunk 5]
```diff
@@ -95,13 +115,3 @@ avr_hal_generic::impl_spi! {
...
-
-#[cfg(any(feature = "atmega8"))]
-avr_hal_generic::impl_spi! {
-    hal: crate::Atmega,
-    peripheral: crate::pac::SPI,
-    sclk: port::PB5,
-    mosi: port::PB3,
-    miso: port::PB4,
-    cs: port::PB2,
-}
```
Rahix, line 107: Nice catch!  I took the liberty to merge this change already in commit b2be3f950c5a ("atmega-hal: Drop redundant atmega8 macro invocation").  Please rebase your PR on latest `main` when you send the next iteration :)

avr-hal-generic/src/spi.rs
```diff
@@ -106,19 +111,50 @@ impl<CSPIN: port::PinOps> embedded_hal_v0::digital::v2::StatefulOutputPin for Ch
...
+}
+
 /// Behavior for a SPI interface.
 ///
 /// Stores the SPI peripheral for register access.  In addition, it takes
 /// ownership of the MOSI and MISO pins to ensure they are in the correct mode.
 /// Instantiate with the `new` method.
+///
+/// This can be used both with the embedded-hal 0.2 spi::FullDuplex trait, and
+/// with the embedded-hal 1.0 spi::SpiBus trait.
```
Rahix, line 157: You can link the traits like this:

```suggestion
/// This can be used both with the embedded-hal 0.2 [`spi::FullDuplex`] trait, and
/// with the embedded-hal 1.0 [`spi::SpiBus`] trait.
///
/// [`spi::FullDuplex`]: `embedded_hal_v0::spi::FullDuplex`
/// [`spi::SpiBus`]: `embedded_hal::spi::SpiBus`
```

Rahix (changes requested):
Took a closer look at the changes now.  Please check my comments below :)

PR #504, opened 2024-02-07
attiny85: pwm: Add PWM support for PB3

sjoerdsimons:
PB3 on is the "Complementary Timer/Counter1 Compare Match B Output", which can be used a PWM, but has the reverse polarity.

### DIFF
--- a/mcu/attiny-hal/src/simple_pwm.rs
+++ b/mcu/attiny-hal/src/simple_pwm.rs
@@ -129,7 +129,7 @@ avr_hal_generic::impl_simple_pwm! {
 
 #[cfg(feature = "attiny85")]
 avr_hal_generic::impl_simple_pwm! {
-    /// Use `TC1` for PWM (pins `PB4`)
+    /// Use `TC1` for PWM (pins `~PB3`, `PB4`)
     ///
     /// # Example
     /// ```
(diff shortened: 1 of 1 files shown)


sjoerdsimons:
Fwiw the main awkwardness is the reverse polarity; I'm not sure if that's ok or if the `impl_simple_pwm` should be adopted to allow masking that difference for the user (by setting the duty to `MAX - duty`  essentially)

Rahix:
> or if the impl_simple_pwm should be adopted to allow masking that difference for the user

If that's possible by simply reversing the duty cycle, I'd be in favor of doing that.  The abstraction should behave the same for all pins, even if the underlying hardware works differently.  If a difference can't be abstracted away, it needs to be visible to the user of the abstraction.  At least that's how I see it.

sjoerdsimons:
My bad; After spending some quality time with the atmel documentation it turns out this didn't actually work as I was wrongly assuming :)

The "Complementary Timer/Counter1 Compare Match B Output" aka ~OCR1B on PB3 for the attiny can only be enable together with OCR1B on PB4; Or in other words when using PB3 as a PWM signal, PB4 will also be a PWM output iff PB4 is set to an output.  For my application i've ended up using PB4 as input and PB3 as PWM as i'm hacking up a existing pcb; Which works fine ofcourse.. 

But it's not something that can be expressed in simple_pwm at the moment.. I guess to support the complementary pwm outputs there should pwm implementation that takes *both* pins but allows you to get something like  e.g. `Pin<Input<_>, PB4>` but such that it's restricted in being set to an output.

Unfortunately that's a bit too much atm for me for a quick project :) So i'll close this

Rahix:
> After spending some quality time with the atmel documentation it turns out this didn't actually work as I was wrongly assuming :)
> 
> The "Complementary Timer/Counter1 Compare Match B Output" aka ~OCR1B on PB3 for the attiny can only be enable together with OCR1B on PB4; Or in other words when using PB3 as a PWM signal, PB4 will also be a PWM output iff PB4 is set to an output.

Ah, I just love these impossible to abstract hardware implementations...

I guess it doesn't make much sense trying to shoehorn this design into our existing abstraction.  If someone needs to use this specific feature, they can always do so using bare register writes...

PR #508, opened 2024-02-16
ravedude: Use a different avrdude config for avrdude 7.X

### LINKED ISSUE #498: avr-hal using an old avrdude.config file
I think this file: 

https://github.com/avrdudes/avrdude/discussions/1636

Rahix:
The config we've been shipping so far was only for avrdude version 6.X. This config does not work when being used with avrdude 7.X versions.

As we do not have control over the installed avrdude version at this time, we need to provide different configurations depending on the avrdude version we find.

As a stop-gap solution, import the released default config for avrdude 7.3 and attempt using it for all avrdude 7.X versions.

In the future, we'll have to rethink this.  It is probably a much better solution to simply vendor avrdude entirely.

Fixes #498.


Rahix:
Please give this a try, to see whether it fixes the problem.  You can install the patched ravedude like this:

```bash
cargo install --git https://github.com/rahix/avr-hal --branch avrdude-7 ravedude
```

Paging @weiying-chen, @Johnwf01john, and @vs-mark-graham-smith.

Johnwf01john:
checked this and working OK now. Thanks

vs-mark-graham-smith:
@Rahix 

Checked this now and also working.

Cheers

SushantChandla:
Thanks 🙏

Rahix:
(Released ravedude 0.1.7 with this change)

PR #517, opened 2024-03-08
Upgrading spi to embedded_hal 1.0

rubend056:
This is my follow up to #503, with your suggestions, since I needed the new Trait for a library I'm using. Perhaps we can continue to refine the implementation.

This might be good progress towards #468.

### DIFF
--- a/avr-hal-generic/src/spi.rs
+++ b/avr-hal-generic/src/spi.rs
@@ -370,7 +370,7 @@ where
         SpiBus::flush(self)?;
 
         let longest = read.len().max(write.len());
-        for i in 0..=longest {
+        for i in 0..longest {
             let r = self.p.raw_transaction(*write.get(i).unwrap_or(&0x00));
             if i < read.len() {
                 read[i] = r;
--- a/avr-hal-generic/src/spi.rs
+++ b/avr-hal-generic/src/spi.rs
@@ -78,7 +78,9 @@ pub trait SpiOps<H, SCLK, MOSI, MISO, CS> {
 
     fn raw_read_buf(&self, buf: &mut [u8]);
     fn raw_write_buf(&mut self, buf: &[u8]);
+    
     fn raw_read_write_buf(&mut self, buffer: &mut [u8], bytes: &[u8]);
+    fn raw_read_write_in_place(&mut self, buf: &mut [u8]);
 }
 
 /// Wrapper for the CS pin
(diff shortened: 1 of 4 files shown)


avr-hal-generic/src/spi.rs
```diff
@@ -266,6 +306,59 @@ where
...
+    fn transfer_in_place(&mut self, buffer: &mut [u8]) -> Result<(), Self::Error> {
+        self.p.raw_read_write_in_place(buffer);
+        Ok(())
+    }
+    fn flush(&mut self) -> Result<(), Self::Error> {
+        if self.write_in_progress {
+            while !self.p.raw_check_iflag() {}
+            self.write_in_progress = false;
+        }
+        Ok(())
```
Rahix, line 358: This is the only thing I'm still not entirely sure about:  As outlined in comment https://github.com/Rahix/avr-hal/pull/503#discussion_r1485565042, I see potential issues in the interaction of the new methods with the old ones.  While the new methods always leave the peripheral in an idle state, the old ones do not.  So we need to prepare for situations where both kinds are called from a single program.

I think the most simple solution for now is to just call `self.flush()` at the beginning of all the new methods (`SpiBus::{read,write,transfer,transfer_in_place}`).  With a comment to explain why, for example:

```rust
    fn write(&mut self, write: &[u8]) -> Result<(), Self::Error> {
        // Must flush() first, if asynchronous operations happened before this
        self.flush().unwrap_infallible();
        Ok(self.p.raw_write_buf(write))
    }
```

Rahix (commented):
Thanks a lot for picking up the changeset from the other PR.  How much did you test this on real hardware so far?

Regarding the code, I only have one topic left to discuss, see below.

rubend056:
Now that I re-read the code, these function names are confusing `raw_read` and `raw_read_buf`, their name implies the only difference is one is a single byte read, and the other is multiple bytes read. But functionally one **just** reads `spdr` and the other makes a transaction on the bus by exchanging 0x00 with slave, reading incoming, repeating for every byte.

I understand this doesn't concern the end user of the library but it makes it confusing to understand by the maintainers.

rubend056:
Ok, I've reworked the code, it should be functionally the same just slightly more clear.

rubend056:
The code was tested/re-tested with an Uno sending/receiving packets using the nrf24l01 module, register readings and transmission were successful. The functions directly used were `transfer` and `transfer_in_place`.

Rahix (approved):
Looking very good, thanks a lot!

Issue #521, opened 2024-03-16
Add configurable board options to ravedude


Creative0708:
I'm using Rust to program an ATtiny85 through an Arduino Nano, and I'd like to use ravedude as a convenient runner for this. This is easily doable with avrdude:
```
avrdude -p t85 -P /dev/<serial> -c avrisp -b 19200 -U flash:w:<file>
```
However, currently the board options are hardcoded in ravedude. I've resorted to modifying ravedude and adding a custom board option:
```rs
struct AvrIspAttiny85;

impl Board for AvrIspAttiny85 {
    // [...]

    fn avrdude_options(&self) -> avrdude::AvrdudeOptions {
        avrdude::AvrdudeOptions {
            programmer: "avrisp",
            partno: "t85",
            baudrate: Some(19200),
            do_chip_erase: true,
        }
    }

    // [...]
}
```

This isn't ideal. The proposal in #278 which uses TOML files to configure ravedude seems like the best option, and I'd be happy to work on implementing this into ravedude.
There are some considerations (at least, some that I've thought about) before work can be started/continued on this:
- Whether we want this at all. This seems true, as making ravedude a generic tool for all AVR boards has been mentioned before (in the issue), but I wanted to confirm. After all, I don't want to write a PR that may not actually be merged
- Whether the TOML format in the issue is finalized
- Something about progress already made on this? Rahix mentioned starting work on something like this but not being able to finish it.

All things considered, this seems like a fun feature to implement and I'd like to work on this. If there are any issues that someone has with this plan, please leave feedback!

I could have made this a GitHub discussion, but since this is "technically" a feature request I figured it'd be better as an issue. If this would have been better as something else, let me know.

Thanks for making such an amazing tool!

stappersg:
On Fri, Mar 15, 2024 at 08:35:40PM -0700, Colin Cai wrote:
> All things considered, this seems like a fun feature to implement
> and I'd like to work on this.

Go for it!

Your proposal seems reasonable.
Make the next step.



> If there are any issues that someone has with this plan,
> please leave feedback!

Don't let that block you.


Regards
Geert Stappers
Fanboy of this project, surely not project lead
-- 
Silence is hard to parse

Rahix:
> This isn't ideal.

100% agree.  It was a quick and dirty "get something shipped" approach but it is horrible in every regard.

> The proposal in #278 which uses TOML files to configure ravedude seems like the best option, and I'd be happy to work on implementing this into ravedude.

[...]

PR #522, opened 2024-03-18
Add file config support for ravedude

Creative0708:
Followup PR to #521 

What the title says. This adds a board named `custom` to ravedude that reads from Ravedude.toml from the project directory.

Works with this config (in Ravedude.toml):

```toml
name = "ATtiny85 through AVRISP"

[reset]
automatic = true

[avrdude]
programmer = "avrisp"
partno = "t85"
baudrate = 19200
do_chip_erase = true
```

And this cargo runner:
```toml
runner = "/path/to/ravedude custom -P /dev/ttyUSB0"
```

Minor detail but this PR also changes `get_board` to return `anyhow::Result<Box<dyn Board>>`, if that matters.

Anyways, format and implementation of custom board support in ravedude is up for discussion. If we reach a consensus about what would be best, this PR will be marked as ready for review.


Rahix:
> Leveraging the power of serde and the toml crate, we get nice error messages whenever a field is missing:

nice! :+1:

> This is fully backwards-compatible with previous ravedude configurations, as it just adds a board named "custom". Now that I've got some chance to think, this feels a bit... clunky.

Feel free to drop the entire `Board` trait abstraction. I was assuming that you were going to do that anyway :D  I see no reason to stick to the existing architecture here - it was mostly written this way to quickly get something running.

Whatever we end up doing with the "builtin boards", they should also simply get encoded in the new TOML format somewhere, as you suggested.

> I'd rather have a mandatory Ravedude.toml in the project root for all boards.

Interesting idea.  My original design approach was to model `ravedude` after [`probe-run`](https://github.com/knurling-rs/probe-run).  A simple tool to be dropped into the cargo config as a project runner.

Your suggestion reminds me of the approach of [`cargo-embed`](https://probe.rs/docs/tools/cargo-embed/) which also has a more or less mandatory config file for all its parameters.

I don't see a big reason why not to go down this route, although I think we should consider the specifics carefully:

- First of all, I would then suggest to drop a lot of the commandline arguments and push them into `Ravedude.toml` instead.  Something like `-P` should stay, but also get a corresponding setting in the file.

- We suggest building new projects using the [`avr-hal-template`](https://github.com/Rahix/avr-hal-template) which could also generate an initial `Ravedude.toml` without problems.

- I'm not a fan of the suggestion that people will always encode all board settings in their own `Ravedude.toml` rather than referencing a "well known board".  The reason is that this prevents us from rolling out fixes/changes for the well known boards.  Unless customization is necessary, I think users should still just reference a board by its name.

- A possible approach would maybe be one where a board may be referenced by name, but whether it is or isn't, custom settings can be used to override or define parameters.

To put it into concrete examples:

```toml
# Ravedude.toml for using well known board
[general]
console = true
baudrate = 57600

board = "uno"
```

```toml
# Ravedude.toml for using a completely custom board
[general]
console = true
baudrate = 57600

[board]
name = "ATtiny85 through AVRISP"

[board.reset]
automatic = true

[board.avrdude]
programmer = "avrisp"
partno = "t85"
baudrate = 19200
do_chip_erase = true
```

```toml
# Ravedude.toml for customizing well known board
[general]
console = true
baudrate = 57600

[board]
name = "Uno without automatic reset"
inherit = "uno"

[board.reset]
automatic = false
```

What do you think?

Creative0708:
> Feel free to drop the entire `Board` trait abstraction. I was assuming that you were going to do that anyway :D

> Whatever we end up doing with the "builtin boards", they should also simply get encoded in the new TOML format somewhere, as you suggested.

Sounds good!

> * First of all, I would then suggest to drop a lot of the commandline arguments and push them into `Ravedude.toml` instead.  Something like `-P` should stay, but also get a corresponding setting in the file.

I assume in this case command-line options would override the TOML file. Would we want a warning to be printed in that case?

> I don't see a big reason why not to go down this route,

Does this signify approval? Because I'd rather not change things just for the sake of changing things. If mandatory TOML configs don't benefit the project in any way, they shouldn't be implemented.

> * I'm not a fan of the suggestion that people will always encode all board settings in their own `Ravedude.toml` rather than referencing a "well known board".  The reason is that this prevents us from rolling out fixes/changes for the well known boards.  Unless customization is necessary, I think users should still just reference a board by its name.

This is a very good point, and I hadn't considered that. Yeah, I agree that this would be a better solution.

The examples look great! However, I would like a consistent case across keys. (Either snake_case or kebab-case is fine, although I'm leaning towards kebab case to mirror Cargo.toml.)

Also, annoyingly, serde doesn't have "inheritance" behavior built-in, and from some very basic searching around the web I've gathered that the best way to do this is just to make everything an `Option<whatever>` and merge the configs after deserialization. Not such a big deal though.


A final note/question: Would this be considered a breaking change...? If not, we'd need to find a way to keep this backwards-compatible. Maybe a "default" config that's used if the file doesn't exist? It depends on how much of a priority backwards compatibility is.

Rahix:
Hi, sorry for only getting back to you now...

> A final note/question: Would this be considered a breaking change...? If not, we'd need to find a way to keep this backwards-compatible. Maybe a "default" config that's used if the file doesn't exist? It depends on how much of a priority backwards compatibility is.

To me, it's really important not to completely break existing users.  At least, there should be a trivial migration path that doesn't require deeper understanding of the inner workings of `ravedude` and `avrdude`...

For simplicity, my suggestion would be to keep the existing invocations working as is, as long as no `Ravedude.toml` is present.  In such a case, I'd like a warning to inform users that they should migrate.

If all else fails, a hacky solution to do this might be to throw the current "legacy" argument parser into a separate module and invoke it when no `Ravedude.toml` is found.  Then the given options are used to construct an in-memory "compatibility" `Ravedude.toml` and use that with the new code.

> I assume in this case command-line options would override the TOML file. Would we want a warning to be printed in that case?

Hm, so let's look at all the CLI args we have right now and I'll share my thoughts for each one:

- `-c, --open-console`: Field in `Ravedude.toml`, we might want `true`/`false` overwrite from CLI to be possible.
- `--debug-avrdude`: Keep as is, no field in `Ravedude.toml`
- `-b, --baudrate <baudrate>`: Field in `Ravedude.toml`, can be overwritten from CLI
- `-P, --port <port>`: Optional field in `Ravedude.toml`, can be overwritten from CLI or env (`$RAVEDUDE_PORT`)
- `-d, --reset-delay <reset-delay>`: Field in `Ravedude.toml`, can be overwritten from CLI
- `<BOARD>`: The chosen board should only be passed on CLI in legacy mode.  When `Ravedude.toml` is present, passing a board should result in an error.

Generally, I wouldn't display warnings for the overrides.  I'd only show warnings for deprecated things or errors when someone combines legacy invocations with a new `Ravedude.toml`.

> Does this signify approval? Because I'd rather not change things just for the sake of changing things. If mandatory TOML configs don't benefit the project in any way, they shouldn't be implemented.

Yeah, let's do it.  For the purpose that `ravedude` wants to fulfil, this seems the most appropriate path.

> although I'm leaning towards kebab case to mirror Cargo.toml

Same here :+1:

---

Thanks again for working on this!  I'm excited to see the result :)  This will make `ravedude` so much more usable and finally elevate it beyond the ugly hack that it currently is...

Creative0708:
Whew! Yeah, no worries, I've been busy as well...

Anyways, a ton of stuff is complete! Additionally, it's now in a state where this could be merged, although a few (minor) things differ from your suggestions. Listing them off:

* Board inheritance now works! Something like

  ```toml
  [general]
  open-console = true
  serial-baudrate = -1
  port = "/dev/ttyUSB0"
  
  [board]
  name = "Arduino uno without automatic reset, for some reason"
  inherit = "uno"
  
  [board.reset]
  automatic = false
  message = "Take a minute to admire the coolness of the RAVEDUDE!"
  ```
  is valid. Everything (well, everything that can be overriden) in the `RavedudeConfig` struct is now an `Option<T>`, and ravedude "merges" the user-provided configuration to one of the hardcoded boards in this case.

* As in the TOML above, entries for the command-line options are now present.
* Just using a named board works too:
  ```toml
  [general]
  open-console = true
  port = "/dev/ttyUSB0"
  
  board = "uno"
  ```

* All the hardcoded boards have been translated it to `src/boards.toml`. Currently the entire file is just `include_str!`'d into the binary because I can't seem to find a way to embed the actual configs into the program instead of parsing the toml at runtime (`static-toml` looks promising, but it only returns a variable TOML mapping, not a specific, validated type). There's a test added to make sure the boards are valid, though. Maybe add that to CI if we can't find a better solution?
* A warning is displayed for people still using the old format:
  ```sh
  $ # cargo runner is `ravedude uno -P /dev/ttyUSB0`
  $ cargo run
  Warning: Passing the board as command-line argument is deprecated, use Ravedude.toml instead:
  
  # Ravedude.toml
  [general]
  port = "/dev/ttyUSB0"

  [board]
  inherit = "uno"

  # ravedude continues like normal...
  ```
  Warning text/format is up for discussion, obviously.
* Various other errors are displayed if the user doesn't include a port, avrdude programmer, avrdude part number, passes a board name as a command-line argument while using Ravedude.toml, doesn't provide a board config, etc.

Additionally, using Ravedude.toml I have successfully uploaded a rust program to the ATtiny85 through the Arduino Nano!

Anyways, yeah, unless we want to make changes to any of this (or if there are bugs), this PR is ready to merge! 🎉 (Not saying that it should be merged right now, but that it _can_ be... there are also merge conflicts that need clearing up)

> Thanks again for working on this! I'm excited to see the result :)

No problem! ravedude (and the entirety of avr-hal, really) makes my life a lot easier when using Rust for AVR-related projects! Happy to help :D

Creative0708:
Oh yeah ­-- We should also probably modify stuff like [the avr-hal template](https://github.com/Rahix/avr-hal-template/tree/main) to use Ravedude.toml, along with adding options for completely custom setups.

ravedude/src/main.rs
```diff
@@ -91,12 +96,53 @@ fn main() {
...
+                args.bin = Some(std::path::PathBuf::from(board));
+            } else {
+                anyhow::bail!("can't pass board as command-line argument when Ravedude.toml is present; set `board = \"{}\"` under [general] in Ravedude.toml", board)
+            }
+        }
+    } else {
+        warning!(
+            "Passing the board as command-line argument is deprecated, use Ravedude.toml instead:\n\n# Ravedude.toml\n{}",
+            toml::to_string(&config::RavedudeConfig::from_args(&args)?)?
+        );
```
Rahix, line 123: This warning is also shown when no commandline arguments are passed at all (and no `Ravedude.toml` is present.  This is a bit confusing, instead we should show an error about `Ravedude.toml` being missing.

ravedude/src/main.rs
```diff
@@ -91,12 +96,53 @@ fn main() {
...
+    let mut args: Args = structopt::StructOpt::from_args();
+    if args.dump_config {
+        println!(
+            "{}",
+            toml::to_string(&board::get_board(args.board.as_deref())?)?
+        );
+        return Ok(());
+    }
+
+    let ravedude_config_exists = std::path::Path::new("Ravedude.toml").try_exists()?;
```
Rahix, line 108: Right now you are searching for `Ravedude.toml` in the current working directory.  This breaks when you e.g. `cargo run` inside the `src/` directory.  Maybe you can check how other tools like `cargo-embed` find their config file and implement something similar?

As a hint, it seems cargo sets some environment variables when running a runner process.  One of them is `CARGO_MANIFEST_DIR` which seems to indicate the directory where cargo found its own `Cargo.toml`...

ravedude/src/config.rs
```diff
@@ -0,0 +1,181 @@
...
+}
+
+#[derive(serde::Serialize, serde::Deserialize, Debug, Default)]
+#[serde(rename_all = "kebab-case")]
+pub struct RavedudeGeneralConfig {
+    pub open_console: Option<bool>,
+    pub serial_baudrate: Option<NonZeroU32>,
+    pub port: Option<std::path::PathBuf>,
+    pub reset_delay: Option<u64>,
+    pub board: Option<String>,
```
Rahix, line 64: Instead of making all fields `Option<>`s, you can also leverage [serde's default attribute](https://serde.rs/field-attrs.html#default).  Where it makes sense, of course.

Here, I think `open_console` would make much more sense that way:

```rust
#[serde(default)]
pub open_console: bool,
```

ravedude/src/config.rs
```diff
@@ -0,0 +1,181 @@
...
+        }
+    }
+}
+
+#[derive(serde::Serialize, serde::Deserialize, Debug)]
+#[serde(rename_all = "kebab-case")]
+pub enum BoardUSBInfo {
+    PortIds(Vec<BoardPortID>),
+    #[serde(rename = "error")]
+    Error(String),
```
Rahix, line 149: Should we really let the config specify the error message here?  I know the old ravedude had each board config manually error out, but we don't need to keep that without good reason.

I think it'd be better for the config to simply state that port guessing is not supported and ravedude then provides a standardized error message.

ravedude/src/main.rs
```diff
@@ -69,9 +74,9 @@ struct Args {
     /// * trinket-pro
     /// * trinket
     /// * nano168
-    /// * duemilanove
+    /// * duemilanovepriori
```
Rahix, line 77: What's this change about?

ravedude/src/main.rs
```diff
@@ -91,12 +96,53 @@ fn main() {
...
+        return Ok(());
+    }
+
+    let ravedude_config_exists = std::path::Path::new("Ravedude.toml").try_exists()?;
+
+    if ravedude_config_exists {
+        if let Some(board) = args.board.take() {
+            if args.bin.is_none() {
+                // The board arg is taken before the binary, so rearrange the args when Ravedude.toml exists
+                args.bin = Some(std::path::PathBuf::from(board));
```
Rahix, line 114: Parsing the binary path into a `String` and then turning it back into a `PathBuf` is not a good idea:  `PathBuf` is based on `OsString` which can potentially represent things a `String` can't.  So doing this may error on certain paths because we try parsing them as utf-8.  We should really handle the path opaquely as a `PathBuf` all the way.

I think the easiest solution for the time being is to make `board` a `PathBuf`, too, and only attempt converting it to a `String` in the legacy usecase. Or making `board` an `OsString` and then converting either direction.

ravedude/src/config.rs
```diff
@@ -0,0 +1,181 @@
...
+#[derive(serde::Serialize, serde::Deserialize, Debug)]
+#[serde(rename_all = "kebab-case")]
+pub struct BoardAvrdudeOptions {
+    pub programmer: Option<String>,
+    pub partno: Option<String>,
+    #[serde(
+        serialize_with = "serialize_baudrate",
+        deserialize_with = "deserialize_baudrate"
+    )]
+    pub baudrate: Option<Option<NonZeroU32>>,
```
Rahix, line 130: Can you explain the double `Option` here?

ravedude/src/board.rs
```diff
@@ -1,439 +1,89 @@
...
-            programmer: "arduino",
-            partno: "atmega328p",
-            baudrate: Some(57600),
-            do_chip_erase: true,
+use std::collections::HashMap;
+
+use crate::config;
+
+fn get_all_boards() -> anyhow::Result<HashMap<String, config::BoardConfig>> {
+    toml::from_str(include_str!("boards.toml")).map_err(|err| {
```
Rahix, line 6: To me this `include_str!()` is fine, I wouldn't waste time trying to parse the file at compile-time.

We should just make sure your testcase actually gets run in CI to catch errors in contributed pull-requests early.  Can you adjust `.github/workflows/ci.yml` to also run `cargo test` for ravedude, in addition to `cargo check` please?

ravedude/src/board.rs
```diff
@@ -1,439 +1,89 @@
...
-    fn guess_port(&self) -> Option<anyhow::Result<std::path::PathBuf>> {
-        Some(Err(anyhow::anyhow!("Not able to guess port")))
-    }
+    })
 }
 
-struct ArduinoNanoNew;
+pub fn get_board(board_name: Option<&str>) -> anyhow::Result<config::RavedudeConfig> {
+    Ok(match board_name {
+        Some(board_name) => {
```
Rahix, line 23: This match arm is just for the legacy usecase, right?

Wouldn't it make more sense to split this into separate functions?  The newer code from `None =>` also feels to me like it should be a part of `config.rs` instead.

ravedude/src/board.rs
```diff
@@ -1,439 +1,89 @@
...
+        for board in all_boards.values() {
+            assert!(board.name.is_some());
+            assert!(board.inherit.is_none());
+            assert!(board.reset.is_some());
+            assert!(board.avrdude.is_some());
+            let avrdude = board.avrdude.as_ref().unwrap();
+            assert!(avrdude.programmer.is_some());
+            assert!(avrdude.partno.is_some());
+            assert!(avrdude.baudrate.is_some());
+            assert!(avrdude.do_chip_erase.is_some());
```
Rahix, line 84: Bonus points for making these asserts show what board failed the test :)  You can do it like this:

```rust
for (name, board) in all_boards.items() {
    assert!(board.name.is_some(), "Board {name:?} is missing a `name`");
    assert!(board.inherit.is_none(), "Board {name:?} has illegal `inherit` key");
```

Rahix (commented):
Finally found time to take a closer look.  Thanks a lot for working on this, it's coming together quite nicely :)

Please check my comments below!

ravedude/src/main.rs
Creative0708: Fixed :+1:

ravedude/src/main.rs
Creative0708: `cargo-embed` uses [figment](https://docs.rs/figment/latest/figment/), which looks interesting. I'm most likely gonna switch ravedude over to it to mirror its behavior (and cargo's behavior) with environment variables, although I've held back on it for now due to it possibly being overkill (?) although ravedude itself isn't going to run on a microcontroller so that probably wasn't a good reason.

`CARGO_MANIFEST_DIR` seems to be only available when using ravedude as a runner. ravedude should be able to access Ravedude.toml no matter how it's being run, in my opinion.

Currently, I just have ravedude scan all parent directories until it finds a `Ravedude.toml` somewhere, which works fine for now... this behavior will probably be replaced by figment, though.

ravedude/src/config.rs
Creative0708: Done! The only other place where I think this would be sensible is `do_chip_erase` in `BoardAvrdudeOptions`, but I don't want to use `#[serde(default)]` on it in case it's missing from the included `boards.toml` and goes unnoticed.

ravedude/src/config.rs
Creative0708: > Should we really let the config specify the error message here?
Honestly, probably not. I just kept it in because that's how it was before and I didn't want to make modifications to the preexisting behavior.

I've removed it. Now the message is just
```
Warning: this board cannot reset itself.

Once reset, press ENTER here:
```
for all the boards without automatic reset. If the behavior should be changed, let me know!

ravedude/src/main.rs
Creative0708: Oops, that's embarrassing. I must've accidentally pasted or typed something and somehow missed it. My bad, it's been removed.

ravedude/src/main.rs
Creative0708: Done :+1:. `board` is now an `OsString` which gets converted into a `String` when used without Ravedude.toml.

ravedude/src/config.rs
Creative0708: The "inner" `Option<NonZeroU32>` signifies the avrdude baudrate, or `None` if it shouldn't be passed (i.e. use the default baudrate). The outer `Option` is there to allow overriding the inherited board's baudrate:
```toml
[board]
inherit = "nano"

[board.avrdude]
# The nano has `baudrate = 57600` but it can be overriden.

# Some(None)
baudrate = -1

# Some(Some(NonZeroU32(8000))
baudrate = 8000

# None
# < no baudrate field set >
```
You can think of it as an `Option<i32>`, but it's an `Option<NonZeroU32>` instead of an `i32` where `-1` is mapped to `None`. So, `Option<Option<NonZeroU32>>`. If this is too convoluted we can change it back into an `i32`.

ravedude/src/board.rs
Creative0708: Done :+1:

ravedude/src/board.rs
Creative0708: Done :+1:. Yes, the match arm was for passing a named board. Now there's two functions, `get_board_from_name` and `get_board_from_manifest`. They're still both in `board.rs` because I feel like `config.rs` should be relatively separate from how it's actually read from the filesystem (`get_all_boards` which parses the included `boards.toml` is also in `board.rs`).

ravedude/src/board.rs
Creative0708: Done :+1:

Creative0708:
Hi, again, sorry for the delay. I've made modifications to the code and left comments based on your feedback. Thanks!

ravedude/src/main.rs
```diff
@@ -91,44 +100,105 @@ fn main() {
...
+        );
+    }
+
+    let mut ravedude_config = match manifest_path.as_deref() {
+        Some(path) => board::get_board_from_manifest(path)?,
+        None => board::get_board_from_name(
+            args
+            .board
+            .as_deref()
+            .ok_or_else(|| anyhow::anyhow!("no board given and couldn't find Ravedude.toml in project, either pass a board as an argument or make a Ravedude.toml."))?
```
Rahix, line 146: Actually, don't reference the deprecated board name method in this error message.  We don't want to advertise the legacy mechanism in any way.  Simply state that we couln't find a Ravedude.toml and give a pointer on how to create one.

ravedude/src/main.rs
```diff
@@ -92,44 +101,105 @@ fn main() {
...
+
+        for dir_to_test in current_dir.ancestors() {
+            let path_to_test = dir_to_test.join(Path::new("Ravedude.toml"));
+            if path_to_test.exists() {
+                break 'manifest_path Some(path_to_test);
+            }
+        }
+
+        None
+    };
```
Rahix, line 119: I wonder if we should still honor `CARGO_MANIFEST_DIR` when it is present in the environment.  The reason I have in mind is usecases when calling `cargo run` with `--manifest-path` while the working directory is completely different.

ravedude/src/main.rs
```diff
@@ -72,7 +79,9 @@ struct Args {
     /// * nano168
     /// * duemilanove
     #[structopt(name = "BOARD", verbatim_doc_comment)]
-    board: String,
+    // When Ravedude.toml exists, the binary is placed where the board should be. This is an OsString to not lose
+    // informaton when we have to take the board as the binary.
+    board: Option<OsString>,
```
Rahix, line 84: Got a bit of a tricky problem here right now:  The `ravedude --help` output is still completely wrong.  Ideally, we can get it to just show the new way of doing things and secretly handle the legacy method in the background.

To do this, I think we need to take the following steps:

1. Rename the current `bin` to something like `oldbin` with the help message stating that this is an argument used for backwards compatibility.
2. Rename the current `board` to `bin`.  Adjust the docstring to the current one from `bin` and drop the `verbatim_doc_comment` setting.

Rahix (commented):
Heads up: I rebased your changes ontop of the latest main to resolve some conflicts.  Please make sure to continue working ontop of the rebased version.

I have a few more comments, check below.  I'll hopefully get to testing the new ravedude in more depth soon, then I'll report back if I find anything else.  It's looking very very promising already, though! :)

ravedude/src/main.rs
Creative0708: Done :+1:. Currently there isn't a descriptive error message, unfortunately. I'm thinking of something like
```
couldn't find Ravedude.toml in project; see `ravedude --help-toml` for more details
```
or even
```
couldn't find Ravedude.toml in project; see `ravedude doc` for more details
```
The message should lead the user to the documentation as just "create a `Ravedude.toml`" doesn't show the correct format for the manifest. Whatever the documentation happens to be, I'll change the error message to link to it afterwards.

ravedude/src/main.rs
Creative0708: Good point, done :+1:

ravedude/src/main.rs
Creative0708: ravedude's manifest format is already a bit complex. I'd say a full description of the manifest keys and what they do would be too long to be directly included in the `ravedude --help` output, so I have a few ideas:

1. Add another command (E.g. `ravedude --help-toml` or the like) which prints out documentation for the manifest format. This would very simply and effectively get the job done but might be hard to scroll/search/etc.
2. Use mdBook or the like to generate online documentation for ravedude. Apart from being harder to setup (especially for offline viewing), it's probably out of scope for this PR. I think this would be the best solution long-term, though, especially if we can write and maintain docs for the rest of avr-hal as well. Or we could just link to GitHub's built-in Markdown viewer/a GitHub wiki.

Regardless, I'm going to work on some basic documentation for the manifest format. I think Option 1 is the best for now.

tippfehlr: another option would be writing/generating a `ravedude.5` man page (5 is for file formats).

The advantage is that this is available online (either the github markdown renderer or manpage sites) and offline and is searchable.

A manpage probably can’t be shipped with cargo install, but it can be shipped by packaging ravedude. (why not?)

Rahix: Sorry for the long wait...  I agree that documentation is out of scope for this PR, let's take care of that in the next step.  My original comment was mostly about fixing the help generated from structopt which you have taken care of, thanks a lot.

For documentation I created issue #579 to track it.

Rahix (approved):
Very, very nice work, thanks so much again for all your efforts here.  I am going to merge it as it stand now, my tests were all successful.  Apart from one minor problem with backwards compatibility, but I'm going to fix that myself :)

We will still need to add documentation as discussed (issue #579), then I'll cut a release and this feature is finally ready for general use.  Such a big step forward for ravedude!

Issue #442, opened 2023-11-04
Pin D10 / PB2 unavailable for PWM Output while SPI enabled


Fundevoge:
I am trying to rewrite my Arduino C project using this crate and found out that it is impossible to use D10/PB2 as a PWM output while SPI is in use:
```rs
fn main(){
    let dp = arduino_hal::Peripherals::take().unwrap();
    let pins = arduino_hal::pins!(dp);

    let timer_1_pwm = Timer1Pwm::new(dp.TC1, Prescaler::Direct);
    let d10 = pins.d10.into_output().into_pwm(&timer_1_pwm);

    let (spi, d10_pwm) = Spi::new(
        dp.SPI,
        pins.d13.into_output(),
        pins.d11.into_output(),
        pins.d12.into_pull_up_input(),
        d10,
        Settings::default(),
    );
}
```

According to The Documentation, PB2  should be in Output Mode while the SPI is in use, otherwise the chip could be pushed into SPI Slave Mode (I only found out about this because of `Spi::new` requires PB2 as an output, which is very good to know, imagine trying to find that bug).
The Documentation also says:
`If SS is configured as an output, the pin is a general output pin which does not affect the SPI system. Typically,
the pin will be driving the SS pin of the SPI Slave.`
In my case, I need PB2 as a PwmOutput and use a different Pin as ChipSelect, but the current code requires PB2 to be configured as a regular Output Pin.

Maybe a fix could be to make a Spi::new constructor where cs is a PwmOutput, something like this:
```rs
pub fn new<TC>(
        p: SPI,
        sclk: port::Pin<port::mode::Output, SCLKPIN>,
        mosi: port::Pin<port::mode::Output, MOSIPIN>,
        miso: port::Pin<port::mode::Input<port::mode::PullUp>, MISOPIN>,
        cs: port::Pin<port::mode::PwmOutput<TC>, CSPIN>,
        settings: Settings,
    ) -> (Self, ChipSelectPinPwm<CSPIN, TC>) {
        let mut spi = Self {
            p,
            sclk,
            mosi,
            miso: miso.forget_imode(),
            write_in_progress: false,
            _cs: PhantomData,
            _h: PhantomData,
        };
        spi.p.raw_setup(&settings);
        (spi, ChipSelectPinPwm(cs))
    }
```
And then add the matching traits to a new `ChipSelectPinPwm` struct, similar to `ChipSelectPin`

Rahix:
First of all, sorry for the long silence...

[...]

PR #529, opened 2024-04-01
generic: spi: Add escape hatch for the ChipSelectPin

Rahix:
There are situations where the `ChipSelectPin` abstraction is not enough - for example when a user wants to use the `ChipSelectPin` as a PWM pin.

To enable such usecases anyway, add an escape hatch for getting back the original `Pin<>` type that's hidden in the `ChipSelectPin`.  This is unsafe, for obvious reasons.

Related to #442.


Rahix:
Merging this anyway even though #442 should get some more attention.

PR #539, opened 2024-04-25
Enabled and fixed SPI for attiny mcu

tbienko:
SPI was not exposed for attiny mcu by `lib.rs` so I've added it. Also I've corrected the `spi.rs` to use `Attiny` instead `Atmega`. This was tested using board with atttiny88 and neopixels connected using SPI driver and seems to work fine


Rahix (approved):
Crazy to think this wasn't tested when it was originally added :D

In any case, thanks a lot for fixing it and also running actual tests to confirm the fix! :)

Issue #537, opened 2024-04-18
cargo build - error[E0658] - proc_macro::Literal::byte_character(byte)


Rahix:
_Originally reported in #535._

Due to a change in `proc-macro2` version `v1.0.80`, the toolchain version we use is no longer supported (see https://github.com/dtolnay/proc-macro2/issues/451).  To reproduce:

```terminal
cargo generate --git https://github.com/rahix/avr-hal-template -r bb141bd00f6b9687e986db8aeb66a3f3f50cca12 -n proc-macro-regression
cd proc-macro-regression/
cargo build
```

which end with the following build error:

```text
   Compiling proc-macro2 v1.0.81
error[E0658]: use of unstable library feature 'proc_macro_byte_character'
   --> /home/rahix/.cargo/registry/src/index.crates.io-6f17d22bba15001f/proc-macro2-1.0.81/src/wrapper.rs:871:21
    |
871 |                     proc_macro::Literal::byte_character(byte)
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #115268 <https://github.com/rust-lang/rust/issues/115268> for more information
    = help: add `#![feature(proc_macro_byte_character)]` to the crate attributes to enable
    = note: this compiler was built on 2024-03-21; consider upgrading it if it is out of date

error[E0658]: use of unstable library feature 'proc_macro_c_str_literals'
   --> /home/rahix/.cargo/registry/src/index.crates.io-6f17d22bba15001f/proc-macro2-1.0.81/src/wrapper.rs:898:21
    |
898 |                     proc_macro::Literal::c_string(string)
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #119750 <https://github.com/rust-lang/rust/issues/119750> for more information
    = help: add `#![feature(proc_macro_c_str_literals)]` to the crate attributes to enable
    = note: this compiler was built on 2024-03-21; consider upgrading it if it is out of date

For more information about this error, try `rustc --explain E0658`.
error: could not compile `proc-macro2` (lib) due to 2 previous errors
```

---

The solution to this issue comes in two parts:

- [ ] We need to upgrade to a newer rust toolchain version again.
- [ ] We need to pin `proc-macro2` to `v1.0.79` somewhere to ensure this won't hit us again.  I am unsure where the right place for this is:
[...]

PR #543, opened 2024-05-04
mega2560 example: rgb led

Rhovian:
Added an example for cycling through colors using an RGB LED.

When attempting to build the executable i ran into #537 , so I pinned the version in the mega2560 crate. Let me know if there is a better way of handling that.

### DIFF
--- a/examples/arduino-mega2560/src/bin/mega2560-rgb-led.rs
+++ b/examples/arduino-mega2560/src/bin/mega2560-rgb-led.rs
@@ -1,0 +1,61 @@
+//! This example demonstrates how to fade an RGB LED connected to an Arduino board.
+//!
+//! Wiring:
+//! - Connect the common cathode of the RGB LED to GND.
+//! - Connect the red LED anode to pin D6 through a current-limiting resistor.
+//! - Connect the green LED anode to pin D5 through a current-limiting resistor.
+//! - Connect the blue LED anode to pin D3 through a current-limiting resistor.
+//!
+//! Note: The current-limiting resistor values depend on the specific RGB LED and the desired brightness.
+//! Typically, a resistor value between 220Ω and 1kΩ is suitable.
+
+#![no_std]
+#![no_main]
+
+use panic_halt as _;
+use arduino_hal::simple_pwm::IntoPwmPin;
+use arduino_hal::simple_pwm::Prescaler;
+use arduino_hal::simple_pwm::{Timer3Pwm, Timer4Pwm};
+
+#[arduino_hal::entry]
+fn main() -> ! {
+    let dp = arduino_hal::Peripherals::take().unwrap();
+    let pins = arduino_hal::pins!(dp);
+
+    let timer0 = Timer4Pwm::new(dp.TC4, Prescaler::Prescale64);
+    let timer1 = Timer3Pwm::new(dp.TC3, Prescaler::Prescale64);
+
+    let mut d6 = pins.d6.into_output().into_pwm(&timer0);
+    let mut d5 = pins.d5.into_output().into_pwm(&timer1);
+    let mut d3 = pins.d3.into_output().into_pwm(&timer1);
+
+    let max_duty_d6 = d6.get_max_duty();
+    let max_duty_d5 = d5.get_max_duty();
+    let max_duty_d3 = d3.get_max_duty();
+
+    let delay_time = 10;
+
+    d6.enable();
+    d5.enable();
+    d3.enable();
+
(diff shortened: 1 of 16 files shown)


examples/arduino-mega2560/src/bin/mega2560-rgb-led.rs
```diff
@@ -0,0 +1,67 @@
+#![no_std]
```
Rahix, line 1: Please add a doc-comment explaining how the RGD LED should be wired to the board.

examples/arduino-mega2560/src/bin/mega2560-rgb-led.rs
```diff
@@ -0,0 +1,67 @@
...
+        for i in 0..max_duty_d3 {
+            d3.set_duty(i);
+            arduino_hal::delay_ms(delay_time)
+        }
+
+        // Fade out blue
+        for i in (0..max_duty_d3).rev() {
+            d3.set_duty(i);
+            arduino_hal::delay_ms(delay_time)
+        }
```
Rahix, line 65: Why are you not using `..=` ranges for these loops as well?

examples/arduino-mega2560/Cargo.toml
```diff
@@ -5,6 +5,9 @@ authors = ["Rahix <rahix@rahix.de>"]
 edition = "2021"
 publish = false
 
+[build-dependencies.proc-macro2]
+version = "=1.0.79"
+
```
Rahix, line 10: Hm, I guess this is unavoidable at this time...  I think the best solution is for me to just add the pinning to all example crates for now.  I'll do this and then you can rebase your branch on top.

examples/arduino-mega2560/src/bin/mega2560-rgb-led.rs
Rahix, line 1: If possible, would you mind moving this example over to Arduino Uno?  That board is sort of our reference platforms and I'd prefer to have most of the examples there.  This will help users to find them all in one place.

Rahix (commented):
Thanks a lot for the contribution!

Please check my comments below :)

examples/arduino-mega2560/Cargo.toml
Rahix: Added the pinning in #544.  Please rebase on latest `main`.

examples/arduino-mega2560/src/bin/mega2560-rgb-led.rs
Rhovian: I don't have an arduino uno to test it on :(

examples/arduino-mega2560/src/bin/mega2560-rgb-led.rs
Rhovian: I've refactored this to better match the existing simple_pwm example for the uno

examples/arduino-mega2560/src/bin/mega2560-rgb-led.rs
Rhovian: done, please let me know if the doc comment is what you're looking for.

examples/arduino-mega2560/Cargo.toml
Rhovian: removed my duplicate dep after rebase

examples/arduino-mega2560/src/bin/mega2560-rgb-led.rs
Rahix: Yeah, that's exactly what I wanted :)

I have one minor nit:  Please place the comment at the top of the file, using `//!` instead of `///`.  That's the way to make a doc-comment for the entire file.

Rahix: To be honest, then I'd suggest to keep the examples in the Arduino Mega 2560 folder.  I'd rather have them where they were tested and confirmed to work then to move them over.  In the future, someone else can always port them to Uno if the need arises.

Rhovian: Okay sounds good, in the future I may get the uno board and do it myself

Rhovian: done, let me know if there's anything else. Looks like i rolled in some linting changes here, if thats problematic I can close this one in favor of a PR with just the new example.

Rahix (approved):

PR #545, opened 2024-05-04
treewide: Apply changes for new svd2rust API

Rahix:
As avr-device is upgrading to use svd2rust version 0.33.1 (see https://github.com/Rahix/avr-device/pull/155), there are some significant changes in the generated API.  We have to adapt the HAL code to use the new API whereever relevant.  This PR is a preview of what will change.

The following major changes are necessary:

- Registers are no longer accessed as struct-fields of the peripheral, but rather as methods of the peripheral.
- Safe writing to a field or register is now done using a new `.set()` method instead of `.bits()`.

Fortunately, these changes can be trivially automated from rustc error messages using the following script:

```bash
cargo build --message-format json 2>/dev/null \
  | jq '.message.children[].spans[] | {file: .file_name, line: .line_start, col: (.text[0].highlight_start - 1), insert: .suggested_replacement}' 2>/dev/null \
  | jq -r '"sed -ri '"'"'" + (.line | tostring) + "s/^(.{" + (.col | tostring) + "})/\\1" + .insert + "/'"'"' $(cd ../..; realpath " + .file + ")"' \
  | sort | uniq | bash
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,3 +42,6 @@ exclude = [
     "ravedude",
 ]
 resolver = "2"
+
+[patch.crates-io]
+avr-device = { path = "../avr-device" }
--- a/avr-hal-generic/src/adc.rs
+++ b/avr-hal-generic/src/adc.rs
@@ -250,17 +250,17 @@ macro_rules! impl_adc {
 
             #[inline]
             fn raw_read_adc(&self) -> u16 {
-                self.adc.read().bits()
+                self.adc().read().bits()
             }
 
             #[inline]
             fn raw_is_converting(&self) -> bool {
-                self.adcsra.read().adsc().bit_is_set()
+                self.adcsra().read().adsc().bit_is_set()
             }
 
             #[inline]
             fn raw_start_conversion(&mut self) {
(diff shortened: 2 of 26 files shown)


Rahix:
Via rahix/avr-device#157, we have now upgraded to an even newer version of svd2rust, making more changes necessary.  So far, I have discovered:

- Peripheral and field names have been adjusted to proper Rust naming styles.  We need to adjust all peripheral and field names now.
- `.modify()` and `.write()` now have a return value which tripped up some of the code we had.
- `avr-device` now enables `critical-section` by default to accommodate for a change in `svd2rust` [^1] so we need to change our `critical-section` related crate features.

[^1]: https://github.com/Rahix/avr-device/pull/195

Rahix:
Not at all a friend of the code churn produced by the identifier changes.  See commit https://github.com/Rahix/avr-hal/pull/545/commits/ef69e1b1ddc6aafcbfda61372d9946a019d4cedc and https://github.com/Rahix/avr-device/issues/197.

Rahix:
https://github.com/Rahix/avr-device/issues/197 was agreed upon, so I reverted the identifier changes here.

PR #549, opened 2024-05-19
[ravedude] add output modes, newline after n bytes or after char

tippfehlr:
From `--help`:
```
    -o <output-mode>
            Output mode. Can be ascii, hex, dec or bin

        --newline-on <newline-on>
            Print a newline after this byte
            not used with output_mode ascii
            hex (0x) and bin (0b) notations are supported.
            matching chars/bytes are NOT removed
            to add newlines after \n (in non-ascii mode), use \n, 0x0a or 0b00001010
        --newline-after <newline-after>
            Print a newline after n bytes
            not used with output_mode ascii
            defaults to 16 for hex and dec and 8 for bin
            if dividable by 4, bytes will be grouped to 4
```

This helps when working with non-ascii output.

Other solutions:

- use another serial console: too much context switching for me, having the serial console in watch mode is way better.
- use ravedude and pipe into hexdump or xxd: they only handle streamed data line by line, meaning the data is not printed immediately. They also can’t add newlines.

### DIFF
--- a/ravedude/src/config.rs
+++ b/ravedude/src/config.rs
@@ -48,6 +48,9 @@ impl RavedudeConfig {
                 port: args.port.clone(),
                 reset_delay: args.reset_delay,
                 board: args.legacy_board_name().clone(),
+                output_mode: args.output_mode,
+                newline_after: None,
+                newline_on: None,
             },
             board_config: Default::default(),
         })
@@ -63,6 +66,9 @@ pub struct RavedudeGeneralConfig {
     pub port: Option<std::path::PathBuf>,
     pub reset_delay: Option<u64>,
     pub board: Option<String>,
+    pub output_mode: Option<OutputMode>,
+    pub newline_on: Option<char>,
+    pub newline_after: Option<u8>,
 }
 
 impl RavedudeGeneralConfig {
@@ -82,6 +88,9 @@ impl RavedudeGeneralConfig {
         }
         if let Some(reset_delay) = args.reset_delay {
             self.reset_delay = Some(reset_delay);
+        }
+        if let Some(output_mode) = args.output_mode {
+            self.output_mode = Some(output_mode);
         }
         Ok(())
     }
@@ -180,4 +189,12 @@ impl BoardConfig {
             None => None,
         }
     }
+}
+
+#[derive(Debug, Clone, Copy, Deserialize, Serialize)]
+pub enum OutputMode {
+    Ascii,
+    Hex,
+    Dec,
+    Bin,
 }
--- a/ravedude/src/config.rs
+++ b/ravedude/src/config.rs
@@ -67,7 +67,7 @@ pub struct RavedudeGeneralConfig {
     pub reset_delay: Option<u64>,
     pub board: Option<String>,
     pub output_mode: Option<OutputMode>,
-    pub newline_on: Option<char>,
+    pub newline_on: Option<String>,
     pub newline_after: Option<u8>,
 }
 
--- a/ravedude/src/config.rs
+++ b/ravedude/src/config.rs
@@ -1,5 +1,6 @@
+use anyhow::Context as _;
 use serde::{Deserialize, Serialize};
-use std::num::NonZeroU32;
+use std::{num::NonZeroU32, str::FromStr};
 
 #[derive(serde::Serialize, serde::Deserialize, Debug, Default)]
 #[serde(rename_all = "kebab-case")]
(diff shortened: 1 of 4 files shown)


Rahix:
Thanks a lot for your contribution!  I like this change a lot.

Right now, we are doing a rather large rework of ravedude in #522.  I'll get back to your PR once that work is finished, otherwise things will get a bit too messy.  I don't expect a lot of conflicts with your changes, though.

Rahix:
> I keep output-mode as a cli override?

Yeah, I think adding a CLI override is a good idea.

ravedude/src/console.rs
```diff
@@ -41,7 +85,38 @@ pub fn open(port: &std::path::Path, baudrate: u32) -> anyhow::Result<()> {
...
+                                    write!(stdout, "{:02x} ", byte).unwrap();
+                                }
+                                Some(Dec) => {
+                                    write!(stdout, "{:03} ", byte).unwrap();
+                                }
+                                Some(Bin) => {
+                                    write!(stdout, "{:08b} ", byte).unwrap();
+                                }
+                            }
+                            // don’t execute in ascii mode, ascii is unreachable here
```
Rahix, line 107: this comment should be dropped?  the unreachable is clear from the match above.

ravedude/src/config.rs
```diff
@@ -63,6 +66,9 @@ pub struct RavedudeGeneralConfig {
     pub port: Option<std::path::PathBuf>,
     pub reset_delay: Option<u64>,
     pub board: Option<String>,
+    pub output_mode: Option<OutputMode>,
+    pub newline_on: Option<String>,
+    pub newline_after: Option<u8>,
```
Rahix, line 71: I think I would handle these via a method on `RavedudeGeneralConfig` to solve the same handling issues as with `OutputMode`.  Right now the parsing, validation, and defaulting of these settings is littered over many places.

```rust
pub enum NewlineMode {
    /// Break lines when encountering this byte
    On(u8),
    /// Break lines after this many bytes
    After(usize),
}

impl RavedudeGeneralConfig {
    pub fn newline_mode(&self) -> anyhow::Result<NewlineMode> {
        if self.newline_on.is_some() && self.newline_after.is_some() {
            anyhow::bail!("newline_on and newline_after cannot be used at the same time");
        }

        if let Some(on_str) = self.newline_on.as_ref() {
            Ok(NewlineMode::On(parse_newline_on(on_str)?))
        } else if let Some(after) = self.newline_after {
            Ok(NewlineMode::After(usize::from(after)))
        } else {
            // default newline mode
            Ok(NewlineMode::After(16))
        }
    }
}
```

You then call this method once in `main` and pass the `NewlineMode` value to `console::open()`.

ravedude/src/console.rs
```diff
@@ -14,12 +41,29 @@ pub fn open(port: &std::path::Path, baudrate: u32) -> anyhow::Result<()> {
...
+            Some(Bin) => 8,
+            _ => 0,
+        },
+    };
+
+    let (spaces, space_after) = if newline_on.is_none() && newline_after % 4 == 0 {
+        (true, 4)
+    } else {
+        (false, 0)
+    };
```
Rahix, line 63: I assume this is a fancyness feature to print output in groupings of 4 when the `newline-after` value allows for this? :)

Such a piece of code screams for the use of an `Option` instead of a `bool` value.  This way you eliminate possible bugs by forgetting to check the `bool`.

```rust
// assuming the `NewlineMode` discussed above
let space_after = match newline_mode {
    NewlineMode::After(after) if after % 4 == 0 => Some(4),
    _ => None,
}
```

...and then further down, the use of this value becomes:

```rust
if let Some(after) = space_after {
    if byte_count % after == 0 {
        write!(stdout, " ").unwrap();
    }
}
```

or alternatively:

```rust
if matches!(space_after, Some(after) if byte_count % after == 0) {
    write!(stdout, " ").unwrap();
}
```

ravedude/src/main.rs
```diff
@@ -107,6 +109,64 @@ mod ui;
...
+            "hex" => Ok(Self::Hex),
+            "dec" => Ok(Self::Dec),
+            "bin" => Ok(Self::Bin),
+            _ => Err(anyhow::anyhow!("unknown output mode")),
+        }
+    }
+}
+
+// this could have fewer nested if’s, but this produces better error messages
+fn parse_newline_on(s: &str) -> Result<char, anyhow::Error> {
```
Rahix, line 126: The `newline_on` value shouldn't be a `char`, but rather a `u8`.  `char` can represent arbitrary unicode codepoints, but really we want to compare bytes here.

Rahix (changes requested):
Thanks a lot for picking this up again!

I have a few comments, see below :)  Please also check the CI after updating, it currently seems to throw some error.

ravedude/src/main.rs
```diff
@@ -107,6 +109,64 @@ mod ui;
...
+            } else {
+                bail!("invalid binary byte")
+            }
+        } else {
+            bail!("binary byte must have 8 characters")
+        }
+    } else {
+        bail!("must be a single character or a byte in hex or binary notation")
+    }
+}
```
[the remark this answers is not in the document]
Rahix: On that note, I have one more change to improve this:

```rust
if let Ok(n) = u8::from_str_radix(&s[2..4], 16) {
    return Ok(n);
} else {
    bail!("invalid hex byte");
}
```

can be shortened to

```rust
return u8::from_str_radix(&s[2..4], 16).context("invalid hex byte in `newline-on`");
```

tippfehlr:
Thanks for the very thorough review, I’ll do that tomorrow. I was happy that my git rebase was successful and also still want to test this with an Arduino.

ravedude/src/console.rs
```diff
@@ -1,8 +1,35 @@
...
+    task_message!(
+        "Output",
+        "{}",
+        match output_mode {
+            Some(Ascii) | None => "ascii",
+            Some(Hex) => "hexadecimal",
+            Some(Dec) => "decimal",
+            Some(Bin) => "binary",
+        }
+    );
```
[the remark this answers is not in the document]
tippfehlr: I removed this completely since it’s obvious anyways.

tippfehlr:
That should address all your comments, I didn’t test it yet though.

Rahix:
Please run `cargo fmt` on the code, there seem to be a few formatting issues.  Let me know once you have tested it :)

tippfehlr:
That should do it, sorry for not formatting it right away. I also tested it now, no problems (only that the console info was printed twice, fixed in the last commit. forgot to mention that in the commit message though)

tippfehlr:
Is something blocking this?

Rahix:
> I didn't get what your push was for, that's why I asked.

That was just a rebase to ensure CI is still passing with things that had happened on `main`. :)

ravedude/src/main.rs
```diff
@@ -257,6 +264,12 @@ fn ravedude() -> anyhow::Result<()> {
         anyhow::bail!("no named board given and no board options provided");
     };
 
+    if ravedude_config.general_options.newline_on.is_some()
+        && ravedude_config.general_options.newline_after.is_some()
+    {
+        anyhow::bail!("newline_on and newline_after cannot be used at the same time");
+    }
+
```
Rahix, line 272: This check can now be dropped that we ensure the same thing in `RavedudeGeneralConfig::newline_mode()`, right?

tippfehlr: true, good catch!

tippfehlr:
Thanks for merging this! I just update the AUR package to 0.2.1 too.

PR #556, opened 2024-06-03
impl wdt modes

### LINKED ISSUE #552: Watchdog intterupt mode
Hello,

i have a intentional deadlock, that is resolved by the watchdog of the arduino. The deadlock occurs, because i want my arduino to restart, if my host computer crashes  (happens quite frequently...). The watchtog restarts the board, but i also need to keep some variables accross this restart. I could resolve my issue, if i could set the watchdog to interrupt-mode, as that would tell it to not restart the board and give me time to do something about it, but i have not found a way to do that.  
Is this possible with this crate?  
If yes, it would be very cool, if you could hint me in the right direction!
If not, see this as a friendly feature request.

Regards, Jonas

JonasFocke01:
This adds the ability to set the mode for the watchdog.  
Previously the wdt mode was enforced to softreset the board.  
Now we can also set it to trigger the `WDT` interrupt by calling `watchdog.set_mode()`.  

This PR closes #552 

Please let me know if there is anything i need to change!

### DIFF
--- a/avr-hal-generic/src/wdt.rs
+++ b/avr-hal-generic/src/wdt.rs
@@ -37,6 +37,12 @@ pub trait WdtOps<H> {
     /// **Warning**: This is a low-level method and should not be called directly from user code.
     fn raw_init(&mut self, m: &Self::MCUSR);
 
+    /// Sets the mode for the watchdog timer.
+    /// Refer to the `WdtMode` enum for more infos.
+    ///
+    /// **Warning**: This is a low-level method and should not be called directly from user code.
+    fn raw_set_mode(&mut self, mode: WdtMode);
+
     /// Start the watchdog timer with the specified timeout.
     ///
     /// If the timeout value is not supported, `Err(())` should be returned.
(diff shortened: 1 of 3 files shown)


avr-hal-generic/src/wdt.rs
```diff
@@ -66,6 +87,14 @@ impl<H, WDT: WdtOps<H>> Wdt<H, WDT> {
         Self { p, _h: PhantomData }
     }
 
+    pub fn rearm(&mut self, m: &WDT::MCUSR) {
```
Rahix, line 90: Please add documentation: What is this function used for? What effects does it have?

avr-hal-generic/src/wdt.rs
```diff
@@ -66,6 +87,14 @@ impl<H, WDT: WdtOps<H>> Wdt<H, WDT> {
...
+        self.p.raw_init(m);
+    }
+
+    pub fn set_mode(&mut self, mode: WdtMode) {
```
Rahix, line 94: Please add documentation.

avr-hal-generic/src/wdt.rs
```diff
@@ -55,6 +61,21 @@ pub trait WdtOps<H> {
     fn raw_stop(&mut self);
 }
 
+#[derive(Copy, Clone)]
+/// The mode that dictates how the watchdog should behave.
+/// This needs to be called **after** the `start` function.
```
Rahix, line 66: You cannot "call" an enum.  This statement probably belongs into the documentation of one of the methods?

avr-hal-generic/src/wdt.rs
```diff
@@ -55,6 +61,21 @@ pub trait WdtOps<H> {
...
+pub enum WdtMode {
+    /// Trigger the interrupt `WDT`, but do not perform a softreset.
+    /// **Warning**: On many `atmega` devices, a watchdog timeout resets the mode automatically, to
+    /// prevent the microcontroller from getting stuck in a infinite loop. So subsequent timeouts
```
Rahix, line 70: Please be specific here: Which MCUs actually behave like this and which don't?  Is it even a good idea to distingush about this or should we rather require the call of `rearm()` on _every_ AVR microcontroller?

Rahix (commented):
Hi, thanks a lot for the contribution!

Please have a look at the CI failures. Is this something you need to fix on your side or is it about bad register definitions that need to be fixed in `avr-device`?

Additionally, please add an example to `examples/arduino-uno` which demonstrates the use of the WDT interrupt.

PR #559, opened 2024-06-09
arduino-hal: provide renamed types matching fields

### LINKED ISSUE #555: peripherals type alias
i think peripherals type names are hard to write and remember currently,

it makes some problem because rust requires you to put type on generic parameters, function arguments or static/const values

for example how can tell what's type of pin d9 on UNO is? `atmega_hal::port::PB1`

however names should be different depending on the board

first, maybe we can introduce some macro that parses given value and return the type

iirc this is used in syn with `token!` macro

or we can just create type alias

### LINKED ISSUE #558: Enhance `avr_hal_generic::renamed_pins! {}` to also generate type aliases for each pin
The [`renamed_pins! {}`](https://github.com/Rahix/avr-hal/blob/38549cc4142881d302374535b3621578ffccaff2/avr-hal-generic/src/port.rs#L772) macro is used to generate a `Pins` struct where the pins can be identified by their names on a specific board.  This is nice because users don't have to look up the pin mapping every time.  The macro is used extensively in `arduino-hal` to name pins by their "board name" on each Arduino board.

However, when pins need to be referred to by their type, this abstraction breaks as there is no way to name a pin type by the board name right now.  It would be great if the `renamed_pins! {}` macro could also create type aliases for each pin.  Something like this:

```rust
// From this line:
//         pub a0: atmega_hal::port::PC0 = pc0,
pub type A0<MODE> = Pin<MODE, atmega_hal::port::PC0>;

//         pub d9: atmega_hal::port::PB1 = pb1,
pub type D9<MODE> = Pin<MODE, atmega_hal::port::PB1>;
```

CC: @LuigiPiucco, @djdisodo

LuigiPiucco:
A very simple improvement to fix #558 and fix #555.

The approach is a basic application of renaming with `paste!`, but I choose to generate slightly different code than suggested:

```rust
pub struct Pins {
    // Notice the lack of documentation here, I thought it became redundant...
    pub a0: PinA0<
        mode::Input<mode::Floating>,
    >,
    // ...
}

// ...since I moved it to the type aliases themselves.

/// `A0`
///
/// * ADC0 (ADC input channel 0)
/// * PCINT8 (pin change interrupt 8)
pub type A0 = atmega_hal::port::PC0;
// ...

// It seemed more consistent to use the pin name in uppercase for the
// raw type and "Pin<pin name in uppercase>" for the wrapped version.

/// Safely wrapped version of:
/// `A0`
///
/// * ADC0 (ADC input channel 0)
/// * PCINT8 (pin change interrupt 8)
pub type PinA0<MODE> = Pin<MODE, atmega_hal::port::PC0>
```

In rust-analyzer the type aliases are resolved to the underlying type, so in this implementation the documentation popups still show the MCU names. We could fix this by using a newtype with `Deref`, but that seems like overkill.

### DIFF
--- a/avr-hal-generic/src/port.rs
+++ b/avr-hal-generic/src/port.rs
@@ -781,13 +781,18 @@ macro_rules! renamed_pins {
             type McuPins = $mcu_pins:ty;
         }
     ) => {
-        $(#[$pins_attr])*
-        pub struct Pins {
-            $($(#[$pin_attr])*
-            pub $pin_name: $pin_wrapper<
-                $crate::port::mode::Input<$crate::port::mode::Floating>,
-                $pin_type,
-            >,)+
+        $crate::paste::paste! {
+            $(#[$pins_attr])*
+            pub struct Pins {
+                    $(pub $pin_name: $pin_wrapper<
+                        $crate::port::mode::Input<$crate::port::mode::Floating>,
+                        [<$pin_name:upper>],
+                    >,)+
+            }
+        }
+
+        $crate::paste::paste! {
+            $($(#[$pin_attr])* pub type [<$pin_name:upper>] = $pin_type;)+
         }
 
         impl Pins {


avr-hal-generic/src/port.rs
```diff
@@ -781,13 +781,18 @@ macro_rules! renamed_pins {
...
+            pub struct Pins {
+                    $(pub $pin_name: [<Pin $pin_name:upper>]<
+                        $crate::port::mode::Input<$crate::port::mode::Floating>,
+                    >,)+
+            }
+        }
+
+        $crate::paste::paste! {
+            $($(#[$pin_attr])* pub type [<$pin_name:upper>] = $pin_type;)+
+            $(#[doc = " Safely wrapped version of:"] $(#[$pin_attr])* pub type [<Pin $pin_name:upper>]<MODE> = $pin_wrapper<MODE, $pin_type>;)+
```
Rahix, line 795: Now that you mentioned it, I think this second type should be dropped.  Only keeping what you called the "raw type" seems more consistent to me.

Rahix (commented):
Hey, thanks a lot for taking care of this.

> In rust-analyzer the type aliases are resolved to the underlying type, so in this implementation the documentation popups still show the MCU names. We could fix this by using a newtype with Deref, but that seems like overkill.

Same issue with compiler error messages.  Such is the life of using type aliases in Rust at the moment...  The only thing we can do is making sure people find the aliased type names in the docs easily.

Newtype won't work when HAL functions expect a specific pin.  Yes, you can fix that using even more layers of traits and generics, but that's way worse still.

Rahix (approved):
Thanks a lot!!

Issue #561, opened 2024-06-22
Utilising the `atmega2560`'s "USART in SPI Mode"


CoolSlimbo:
According the the [`atmega2560` datasheet](https://ww1.microchip.com/downloads/en/devicedoc/atmel-2549-8-bit-avr-microcontroller-atmega640-1280-1281-2560-2561_datasheet.pdf), its `USART` modules can be used in "SPI Mode," which essentially gives an extra bus (or four) to add SPI devices to it.
However, within the current implementation of `avr_hal`/`atmega_hal`, it's not plausible to use the USART modules in SPI mode.

Given this is hardware supported on the `atmega`'s (versions of it), it would make sense to have a hardware implementation of it it in the HAL's.

Free to give this an attempt if given a suggestion in the right way for how to get started.

___
### Related Issue:
- #450

Rahix:
Yes, this isn't supported yet.

If anyone is interested in working on a "USART as SPI" driver and needs guidance on how to do this, feel free to ping me!

CoolSlimbo:
I'll gladly give this a shot in implementation, if you could aid a hand in the direction it needs to go in!

Currently, my confusion is just on how one would implement it, whilst also preventing usage of the USART for USART purposes.

Rahix:
> whilst also preventing usage of the USART for USART purposes.

This part is actually very simple.  The `UsartSpi` driver constructor consumes the `USART` peripheral, just like the current UART driver consumes it.  Due to Rust's ownership rules, this means only one of the two can ever exist:

```rust
    let dp = atmega_hal::Peripherals::take().unwrap();
    let pins = atmega_hal::pins!(dp);
    let mut serial = Usart::new(
        dp.USART0, // <- USART0 is moved into `serial` here
        pins.pe0,
        pins.pe1.into_output(),
        Baudrate::<crate::CoreClock>::new(57600),
    );

    let mut spi = UsartSpi::new(dp.USART0, ...);  // <- compiler error because USART0 was moved previously and is no longer available
```

Rahix:
> if you could aid a hand in the direction it needs to go in!

Generally, you'll need to add a `usart_spi` module in `avr-hal-generic`, similar in structure to the existing `spi` or `usart` modules.  This defines the generic `UsartSpi` driver and a macro for the MCU-specific HALs.  This macro is then instanciated in `atmega-hal` to implement USART as SPI for the available USART peripherals.

CoolSlimbo:
[...]

PR #562, opened 2024-06-23
`UsartSpi` Support

CoolSlimbo:
PR regarding issue #561.

This is just so I have the PR made for now, work still needs to be made on it.

So far, `USART0` is implemented for most of the `atmega` mc's. However, they are not tested, as I am lacking something to test it with.

### DIFF
--- a/avr-hal-generic/src/usart_spi.rs
+++ b/avr-hal-generic/src/usart_spi.rs
@@ -36,11 +36,11 @@ macro_rules! impl_usart_spi {
                         w.[<umsel $n>]().spi_master();
 
                         match settings.data_order {
-                            DataOrder::MostSignificantFirst => match settings.mode.phase {
+                            crate::spi::DataOrder::MostSignificantFirst => match settings.mode.phase {
                                 spi::Phase::CaptureOnFirstTransition => w.[<ucsz $n>]().chr5(),
                                 spi::Phase::CaptureOnSecondTransition => w.[<ucsz $n>]().chr6(),
                             },
-                            DataOrder::LeastSignificantFirst => match settings.mode.phase {
+                            crate::spi::DataOrder::LeastSignificantFirst => match settings.mode.phase {
                                 spi::Phase::CaptureOnFirstTransition => w.[<ucsz $n>]().chr7(),
                                 spi::Phase::CaptureOnSecondTransition => w.[<ucsz $n>]().chr8(),
                             },
--- a/avr-hal-generic/src/usart_spi.rs
+++ b/avr-hal-generic/src/usart_spi.rs
@@ -1,7 +1,36 @@
 //! MSPIM Implimentation
-use crate::spi;
+use crate::{port::PinOps, spi};
 
 // This module just impliments a macro for SpiOps, since underlyingly, the Spi type can still be used since it just needs SpiOps
+
+/// Dummy Pin for MPSPIM
+pub struct UsartSPIDummyPin;
+
+impl PinOps for UsartSPIDummyPin {
+    type Dynamic = Dynamic;
+
+    fn into_dynamic(self) -> Self::Dynamic {
+        todo!()
+    }
+
+    unsafe fn out_set(&mut self) {}
+
+    unsafe fn out_clear(&mut self) {}
+
+    unsafe fn out_toggle(&mut self) {}
+
+    unsafe fn out_get(&self) -> bool {
+        false
+    }
+
+    unsafe fn in_get(&self) -> bool {
+        true
+    }
+
+    unsafe fn make_output(&mut self) {}
+
+    unsafe fn make_input(&mut self, pull_up: bool) {}
+}
 
 pub type UsartSpi<H, USART, SCLKPIN, MOSIPIN, MISOPIN, CSPIN> =
     spi::Spi<H, USART, SCLKPIN, MOSIPIN, MISOPIN, CSPIN>;
@@ -60,7 +89,6 @@ macro_rules! add_usart_spi {
                     self.[<ucsr $n b>].write(|w| w
                         .[<txen $n>]().set_bit()
                         .[<rxen $n>]().set_bit()
-                        .[<rxcie $n>]().set_bit()
                     );
 
                     // Set the baudrate of the UBRRn, idk what it should be set to, so for now, it'll be set to 0
@@ -68,8 +96,8 @@ macro_rules! add_usart_spi {
                 }
 
                 fn raw_release(&mut self) {
-                    // Probably a better way to "release" the SPI interface, but from the datasheet, this is what they suggest, so ig it works
                     self.[<ucsr $n c>].write(|w| w.[<umsel $n>]().usart_async());
+                    self.[<ucsr $n b>].reset();
                 }
 
                 fn raw_check_iflag(&self) -> bool {
--- a/avr-hal-generic/src/usart_spi.rs
+++ b/avr-hal-generic/src/usart_spi.rs
@@ -19,6 +19,8 @@ macro_rules! add_usart_spi {
         cs: $cspin:ty,
     ) => {
         $crate::paste::paste! {
+            // This is quite a messy way to get the doc string working properly... but it works!
+            #[doc = concat!("**Clock:** `", stringify!($sclkpin), "`<br>**MOSI:** `", stringify!($mosipin), "`<br> **MISO:** `", stringify!($misopin), "`<br> **CS:** `", stringify!($cspin), "`")]
             pub type [<Usart $n Spi>] = avr_hal_generic::usart_spi::UsartSpi<$HAL, $USART_SPI, $sclkpin, $mosipin, $misopin, $cspin>;
 
             impl $crate::spi::SpiOps<$HAL, $sclkpin, $mosipin, $misopin, $cspin> for $USART_SPI {
--- a/avr-hal-generic/src/usart_spi.rs
+++ b/avr-hal-generic/src/usart_spi.rs
@@ -25,7 +25,7 @@ macro_rules! impl_usart_spi {
 
                     // Setup control registers
                     // We start by setting the UBBRn to 0
-                    self.[<ubrr $n>].write(|w| w.bits(0));
+                    self.[<ubrr $n>].write(|w| unsafe {w.bits(0)});
 
                     // We have to translate the character size register into the 2 bits which are the MSB/LSB and the phase
                     // 5 Bit Char = MSB and 1st
@@ -60,7 +60,7 @@ macro_rules! impl_usart_spi {
                     );
 
                     // Set the baudrate of the UBRRn, idk what it should be set to, so for now, it'll be set to 0
-                    self.[<ubrr $n>].write(|w| w.bits(0));
+                    self.[<ubrr $n>].write(|w| unsafe{w.bits(0)});
                 }
 
                 fn raw_release(&mut self) {
--- a/avr-hal-generic/src/usart_spi.rs
+++ b/avr-hal-generic/src/usart_spi.rs
@@ -7,7 +7,7 @@ use crate::{port::PinOps, spi};
 pub struct UsartSPIDummyPin;
 
 impl PinOps for UsartSPIDummyPin {
-    type Dynamic = Dynamic;
+    type Dynamic = todo!();
 
     fn into_dynamic(self) -> Self::Dynamic {
         todo!()
@@ -32,8 +32,8 @@ impl PinOps for UsartSPIDummyPin {
     unsafe fn make_input(&mut self, pull_up: bool) {}
 }
 
-pub type UsartSpi<H, USART, SCLKPIN, MOSIPIN, MISOPIN, CSPIN> =
-    spi::Spi<H, USART, SCLKPIN, MOSIPIN, MISOPIN, CSPIN>;
+pub type UsartSpi<H, USART, SCLKPIN, MOSIPIN, MISOPIN> =
+    spi::Spi<H, USART, SCLKPIN, MOSIPIN, MISOPIN, UsartSPIDummyPin>;
 
 // Impliment SpiOps trait for USART
 #[macro_export]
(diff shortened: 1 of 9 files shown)


mcu/atmega-hal/src/spi.rs
```diff
@@ -31,6 +31,66 @@ avr_hal_generic::impl_spi! {
     cs: port::PB0,
 }
 
+#[cfg(any(
```
stappersg, line 34:
```
// Devices where UART port can be a SPI port
```

stappersg:
Idea, maybe a wild idea: The new code in **one**  `#[cfg(any(`  `))]` block?

Anyway: Success with testing.

CoolSlimbo:
Wait. It actually worked on hardware? Part of me is surprised lol

I'll probably move it into its own module, but can't use one config, because they align to different ports between devices.

CoolSlimbo:
Currently added in (I think) at minimum all the USART0 implimentations. But there's still a large portion to go, and finding the XCK pin requires getting the datasheet for each.

avr-hal-generic/src/usart_spi.rs
```diff
@@ -0,0 +1,91 @@
...
+        hal: $HAL:ty,
+        peripheral: $USART_SPI:ty,
+        register_suffix: $n:expr,
+        sclk: $sclkpin:ty,
+        mosi: $mosipin:ty,
+        miso: $misopin:ty,
+        cs: $cspin:ty,
+    ) => {
+        $crate::paste::paste! {
+            impl $crate::spi::SpiOps<$HAL, $sclkpin, $mosipin, $misopin, $cspin> for $USART_SPI {
```
Rahix, line 22: Wow, didn't think you'd be able to get by with just implementing the existing `SpiOps`.  That's pretty neat :)

CoolSlimbo: I realised, underlyingly, `Spi` just utilises `SpiOps`, and could cheap out.

Either sometime in the next 3-4 hours, or sometimes in 14 hours, I will have the implementations for the rest of the boards.

Rahix: Haha, don't stress over it!

BTW, please check the failing CI jobs, seems there are still some kinks to iron out...

CoolSlimbo:
Implements USART, correctly, on all devices. Should past CI successfully. 

An example could be in order, however, I am lacking any SPI slaves, so I cannot test this myself.

Rahix:
> An example could be in order, however, I am lacking any SPI slaves, so I cannot test this myself.

SPI can be tested very easily by looping back MOSI to MISO (so TX to RX in this case).  We also have this in our real SPI example, maybe it makes sense to add an example that does the same for `UsartSpi`?

Rahix:
Hey, there seem to be some compiler errors, as seen in CI.  Can you take a look?

armandas:
I went on and tested the `atmega2560-usart_spi-feedback` example. I tested USART1, USART2 and USART3 and they all work. USART0 us connected to USB, so I did not test that.

My diff to get it to compile:

```diff
diff --git a/examples/atmega2560/src/bin/atmega2560-usart_spi-feedback.rs b/examples/atmega2560/src/bin/atmega2560-usart_spi-feedback.rs
index 3a2d145..bce2ad4 100644
--- a/examples/atmega2560/src/bin/atmega2560-usart_spi-feedback.rs
+++ b/examples/atmega2560/src/bin/atmega2560-usart_spi-feedback.rs
@@ -36,12 +36,12 @@ fn main() -> ! {
 
     // Create SPI interface.
     let (mut spi, _) = usart_spi::Usart1Spi::new(
-        dp.SPI,
+        dp.USART1,
         pins.pd5.into_output(),
         pins.pd3.into_output(),
         pins.pd2.into_pull_up_input(),
         pins.pd4.into_output().downgrade(),
-        spi::Settings::default(),
+        atmega_hal::spi::Settings::default(),
     );
 
     loop {
```

Other configs, for reference:

USART2:

```rust
    let (mut spi, _) = usart_spi::Usart2Spi::new(
        dp.USART2,
        pins.ph2.into_output(),
        pins.ph1.into_output(),
        pins.ph0.into_pull_up_input(),
        pins.pd4.into_output().downgrade(),
        atmega_hal::spi::Settings::default(),
    );
```

USART3:

```rust
    let (mut spi, _) = usart_spi::Usart3Spi::new(
        dp.USART3,
        pins.pj2.into_output(),
        pins.pj1.into_output(),
        pins.pj0.into_pull_up_input(),
        pins.pd4.into_output().downgrade(),
        atmega_hal::spi::Settings::default(),
    );
```

avr-hal-generic/src/usart_spi.rs
```diff
@@ -0,0 +1,95 @@
...
+
+                    // Enable receiver and transmitter, and also the rec interrupt.
+                    self.[<ucsr $n b>].write(|w| w
+                        .[<txen $n>]().set_bit()
+                        .[<rxen $n>]().set_bit()
+                        .[<rxcie $n>]().set_bit()
+                    );
+
+                    // Set the baudrate of the UBRRn, idk what it should be set to, so for now, it'll be set to 0
+                    self.[<ubrr $n>].write(|w| unsafe{w.bits(0)});
```
armandas, line 67: UBRR sets the SPI clock frequency, so it should look something like this? I tested and it works, but I'm not sure if this is the best way to write the register in Rust.

```suggestion
                    // Set the clock divider for SPI clock.
                    self.[<ubrr $n>].write(|w| {
                        match settings.clock {
                            crate::spi::SerialClockRate::OscfOver2 => w.bits(0),
                            crate::spi::SerialClockRate::OscfOver4 => w.bits(1),
                            crate::spi::SerialClockRate::OscfOver8 => w.bits(3),
                            crate::spi::SerialClockRate::OscfOver16 => w.bits(7),
                            crate::spi::SerialClockRate::OscfOver32 => w.bits(15),
                            crate::spi::SerialClockRate::OscfOver64 => w.bits(31),
                            crate::spi::SerialClockRate::OscfOver128 => w.bits(63),
                        }
                    });
```

armandas:
@Rahix I was wondering if it would make sense to re-export configuration structs and enums defined in `spi` from `usart_spi`?

avr-hal-generic/src/usart_spi.rs
Rahix: If you are certain on the numbers, you can make use of the commit 65b304ec974a ("generic: spi: Assign discriminant values to clock rate enum") that I just pushed:

```rust
self.[<ubrr $n].write(|w| {
    2u8.pow(settings.clock as u32) - 1
});
```

Rahix:
> I was wondering if it would make sense to re-export configuration structs and enums defined in spi from usart_spi?

To be honest, I would prefer not to add such re-exports.  The more paths there are to reach an item the more convoluted user code will get.

avr-hal-generic/src/usart_spi.rs
```diff
@@ -0,0 +1,95 @@
...
+            // This is quite a messy way to get the doc string working properly... but it works!
+            #[doc = concat!("**Clock:** `", stringify!($sclkpin), "`<br>**MOSI:** `", stringify!($mosipin), "`<br> **MISO:** `", stringify!($misopin), "`<br> **CS:** `", stringify!($cspin), "`")]
+            pub type [<Usart $n Spi>] = avr_hal_generic::usart_spi::UsartSpi<$HAL, $USART_SPI, $sclkpin, $mosipin, $misopin, $cspin>;
+
+            impl $crate::spi::SpiOps<$HAL, $sclkpin, $mosipin, $misopin, $cspin> for $USART_SPI {
+                fn raw_setup(&mut self, settings: &crate::spi::Settings) {
+                    use $crate::hal::spi;
+
+                    // Setup control registers
+                    // We start by setting the UBBRn to 0
```
Rahix, line 31: Don't comment _what_ code is doing, that should be obvious from reading the line below.  Please comment _why_ it is doing this &mdash; what's the purpose of setting UBBRn to 0 here?

avr-hal-generic/src/usart_spi.rs
```diff
@@ -0,0 +1,95 @@
...
+                            spi::Polarity::IdleHigh => w.[<ucpol $n>]().set_bit(),
+                        }
+                    });
+
+                    // Enable receiver and transmitter, and also the rec interrupt.
+                    self.[<ucsr $n b>].write(|w| w
+                        .[<txen $n>]().set_bit()
+                        .[<rxen $n>]().set_bit()
+                        .[<rxcie $n>]().set_bit()
+                    );
```
Rahix, line 64: Don't enable the interrupt by default - it should always be explicitly enabled by the user through a method call, see e.g.

https://github.com/Rahix/avr-hal/blob/65b304ec974a13f27ebc2ced21aa462c4b81ce82/avr-hal-generic/src/usart.rs#L300-L308

You don't need to implement this here, though &mdash; it should be part of the `spi` module.

avr-hal-generic/src/usart_spi.rs
```diff
@@ -0,0 +1,95 @@
...
+
+                    // Set the baudrate of the UBRRn, idk what it should be set to, so for now, it'll be set to 0
+                    self.[<ubrr $n>].write(|w| unsafe{w.bits(0)});
+                }
+
+                fn raw_release(&mut self) {
+                    // Probably a better way to "release" the SPI interface, but from the datasheet, this is what they suggest, so ig it works
+                    self.[<ucsr $n c>].write(|w| w.[<umsel $n>]().usart_async());
+                }
```
Rahix, line 73: Similar to the USART code, you also need to disable the USART peripheral here:

https://github.com/Rahix/avr-hal/blob/65b304ec974a13f27ebc2ced21aa462c4b81ce82/avr-hal-generic/src/usart.rs#L506-L510

(The UCSRnB reset is the important part).

With that included, you can drop your comment.

mcu/atmega-hal/src/usart_spi.rs
```diff
@@ -0,0 +1,104 @@
...
+
+#[cfg(any(feature = "atmega1280", feature = "atmega2560"))]
+avr_hal_generic::add_usart_spi! {
+    hal: crate::Atmega,
+    peripheral: crate::pac::USART0,
+    register_suffix: 0,
+    sclk: port::PE2,
+    mosi: port::PE1,
+    miso: port::PE0,
+    cs: port::Dynamic,
```
Rahix, line 32: Hm, this is quite a hack.  We really don't need the `cs` dance for MSPIM so it would be nice to somehow wrangle the SPI module to not require a CS pin for MSPIM...

avr-hal-generic/src/usart_spi.rs
CoolSlimbo: Unsure of the purpose.

But code examples in the data sheet all do so, explicitly setting it to 0, before setting the baudrate (which I will update the calculation for)

mcu/atmega-hal/src/usart_spi.rs
CoolSlimbo: This was done because I impliment SpiOps to cheap out on work, and not rewrite everything.

I can either reimplement a new spiops, or, force the use of a DummyPin internally (so users don’t have to worry about it) via the DummyPin crate (I believe that’s its name)

Rahix: I would be fine with defining a `UsartSpiDummyCs` pin type that needs to be passed to `Spi::new()` to sidestep this.  Pulling in an external crate for this is not a good idea, I think.

CoolSlimbo: An alternative to the Dummy CS Pin (and what I believe my original thought process was), is how it is, where it's dynamic, meaning the CS can be handled by the driver in some which way, by manual toggle instead of SPI handling it.

avr-hal-generic/src/usart_spi.rs
armandas: Yes, the numbers are correct for UBRR. The numbers do not apply to SPI peripheral, but there it is `match`ed, so no issue.

avr-hal-generic/src/usart_spi.rs
CoolSlimbo: Could you elaborate on what you mean by this?

I understand why to remove it (idk why it was there to begin with), but do you want me to add methods to change the interrupts? Or something other, because I don't understand "it should be part of the spi module"

avr-hal-generic/src/usart_spi.rs
```diff
@@ -0,0 +1,132 @@
...
+
+/// Dummy Pin for MPSPIM
+pub struct UsartSPIDummyPin;
+
+impl PinOps for UsartSPIDummyPin {
+    type Dynamic = todo!();
+
+    fn into_dynamic(self) -> Self::Dynamic {
+        todo!()
+    }
```
armandas, line 14: I found this in port.rs, gave it a try and at least it built. What do you think?

```suggestion
    type Dynamic = Self;

    fn into_dynamic(self) -> Self::Dynamic {
        self
    }
```

avr-hal-generic/src/usart_spi.rs
```diff
@@ -0,0 +1,132 @@
...
+        false
+    }
+
+    unsafe fn in_get(&self) -> bool {
+        true
+    }
+
+    unsafe fn make_output(&mut self) {}
+
+    unsafe fn make_input(&mut self, pull_up: bool) {}
```
armandas, line 32:
```suggestion
    unsafe fn make_input(&mut self, _pull_up: bool) {}
```

avr-hal-generic/src/usart_spi.rs
```diff
@@ -0,0 +1,132 @@
...
+// Impliment SpiOps trait for USART
+#[macro_export]
+macro_rules! add_usart_spi {
+    (
+        hal: $HAL:ty,
+        peripheral: $USART_SPI:ty,
+        register_suffix: $n:expr,
+        sclk: $sclkpin:ty,
+        mosi: $mosipin:ty,
+        miso: $misopin:ty,
```
armandas, line 47:
```suggestion
        miso: $misopin:ty,
        cs: $cspin:ty,
```

armandas (commented):
I messed with the code for a bit and got it to compile. Don't have a board on hand at the moment, but will try to run the code this week.

avr-hal-generic/src/usart_spi.rs
Rahix: From my side this is fine.  This pin-type is non-functional anyway - it just exists to bypass API problems.

Rahix: `Spi` only takes the CS pin to enforce a hardware restriction, it actually doesn't do anything with the CS pin related to "CS" operation at all.  This is an unfortunate side-effect of the AVR SPI peripheral design.  You can read up more on this in #27.

Actually, I had the same thought as you here:  We need the Dummy Pin to make the type-system happy, but this should be hidden from the user if at all possible.  The easiest way to hide it is by adding a new constructor, like the one you showcased here.  I think that's the best option.

armandas: Thanks for the reference, I did not know about this behaviour 🤦‍♂️ 

My remaining issue with the separate constructor is how to enforce that the user passes actual USART peripheral? Because, as it stands, the following compiles...

```Rust
let mut spi = spi::Spi::new_from_usart(
    dp.SPI,
    pins.pb1.into_output(),
    pins.pb2.into_output(),
    pins.pb3.into_pull_up_input(),
    spi::Settings::default(),
);
```

Rahix: Hm, can we add a trait bound on `UsartOps` maybe?  If not, I'd introduce a new marker trait `UsartSpiDevice` that is implemented alongside `SpiOps` for USART peripherals.  We can then add a trait bound on this marker trait to the new constructor.

Rahix:
BTW, @armandas, if you want to pick up this work while @CoolSlimbo isn't around, feel free to open a new PR based on top of this one.

avr-hal-generic/src/usart_spi.rs
armandas: Adding a trait bound on `UsartOps` seems to do it. I'll test on hardware and, if everything goes well, make a PR as you advised.

Rahix:
Closing this in favor of the follow-up #588.  Thanks a lot @CoolSlimbo for your work here!

PR #581, opened 2024-09-15
specs: Set max-atomic-width to 16

taiki-e:
Reflect rust-lang/rust change by https://github.com/rust-lang/rust/pull/114495.


Rahix (approved):
Thanks!

PR #588, opened 2024-10-05
UsartSpi Support

### LINKED ISSUE #561: Utilising the `atmega2560`'s "USART in SPI Mode"

armandas:
This is the continuation of the @CoolSlimbo's good work in #562. Closes #561.

I've made the following changes:

- Resolved the CS issue by adding a new constructor to Spi: `new_from_usart`
- Updated `usart_spi-feedback` example
- Fixed clippy warnings in the macro as well as some typos in the comments
- Address some of the remaining comments

I have run the `usart_spi-feedback` example on an Arduino Mega 2560 and confirmed that USART 1, 2 and 3 work.

### DIFF
--- a/avr-hal-generic/src/lib.rs
+++ b/avr-hal-generic/src/lib.rs
@@ -21,6 +21,7 @@ pub mod port;
 pub mod simple_pwm;
 pub mod spi;
 pub mod usart;
+pub mod usart_spi;
 pub mod wdt;
 
 /// Prelude containing all HAL traits
--- a/avr-hal-generic/src/port.rs
+++ b/avr-hal-generic/src/port.rs
@@ -781,13 +781,18 @@ macro_rules! renamed_pins {
             type McuPins = $mcu_pins:ty;
         }
     ) => {
-        $(#[$pins_attr])*
-        pub struct Pins {
-            $($(#[$pin_attr])*
-            pub $pin_name: $pin_wrapper<
-                $crate::port::mode::Input<$crate::port::mode::Floating>,
-                $pin_type,
-            >,)+
+        $crate::paste::paste! {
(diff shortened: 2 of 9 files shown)


armandas:
Hmm, `atmega1284p` build fails. Seems like `avr-device` issue?

`avr-device-0.5.4/src/devices/atmega2560/usart0/ubrr0.rs`

```Rust
    #[doc = "Writes raw bits to the register."]
    #[inline(always)]
    pub fn bits(&mut self, bits: u16) -> &mut Self {
        unsafe { self.0.bits(bits) };
        self
    }
```

`avr-device-0.5.4/src/devices/atmega1284p/usart0/ubrr0.rs`:

```Rust
    #[doc = "Writes raw bits to the register."]
    #[inline(always)]
    pub unsafe fn bits(&mut self, bits: u16) -> &mut Self {
        self.0.bits(bits);
        self
    }
```

armandas:
@Rahix I think this PR should be good to review. I tried to address the comments in #562, but let me know if there are any changes needed.

Issue #590, opened 2024-10-12
Adding code examples to the documentation


armandas:
I'm thinking to make documentation more useful and accessible by adding code examples. Would you be happy to accept this?
To start with, I would add just a basic example to the documentation of each module, like below:

![image](https://github.com/user-attachments/assets/f936bf28-bddb-493b-b73c-4deca2fc987e)

I would also like to make one PR per crate or possibly per module. What to you think?

Rahix:
Yes please, the documentation is still quite lacking at the moment.  I think it may also be a good idea to link to relevant example programs from the docs.  I think working through this per-crate is fine, it that works for you.

PR #591, opened 2024-10-12
Add examples to atmega-hal documentation

armandas:
Part of #590

### DIFF
--- a/examples/atmega2560/src/bin/atmega2560-eeprom.rs
+++ b/examples/atmega2560/src/bin/atmega2560-eeprom.rs
@@ -1,0 +1,39 @@
+#![no_std]
+#![no_main]
+
+use atmega_hal::delay::Delay;
+use atmega_hal::usart::{Baudrate, Usart};
+use atmega_hal::Eeprom;
+use embedded_hal::delay::DelayNs;
+use panic_halt as _;
+
+// Define core clock in the root crate
+type CoreClock = atmega_hal::clock::MHz16;
+
+const BOOT_COUNT_OFFSET: u16 = 0;
+
+#[avr_device::entry]
+fn main() -> ! {
+    let dp = atmega_hal::Peripherals::take().unwrap();
+    let pins = atmega_hal::pins!(dp);
+
+    let mut delay = Delay::<crate::CoreClock>::new();
+
+    // set up serial interface for text output
+    let mut serial = Usart::new(
+        dp.USART0,
+        pins.pe0,
+        pins.pe1.into_output(),
+        Baudrate::<crate::CoreClock>::new(57600),
+    );
+
+    let mut eeprom = Eeprom::new(dp.EEPROM);
+
+    let mut boot_count = eeprom.read_byte(BOOT_COUNT_OFFSET);
+    boot_count = boot_count.wrapping_add(1);
+    eeprom.write_byte(BOOT_COUNT_OFFSET, boot_count);
+
+    ufmt::uwriteln!(&mut serial, "Boot count: {}", boot_count).unwrap();
+
+    loop {}
+}
--- a/mcu/atmega-hal/src/adc.rs
+++ b/mcu/atmega-hal/src/adc.rs
@@ -2,8 +2,8 @@
 //!
 //! # Example
 //!
-//! Complete example source code can be found in the
-//! [repository](https://github.com/Rahix/avr-hal/blob/main/examples/atmega2560/src/bin/atmega2560-adc.rs).
+//! Complete example source code can be found in the repository:
+//! [`atmega2560-adc.rs`](https://github.com/Rahix/avr-hal/blob/main/examples/atmega2560/src/bin/atmega2560-adc.rs)
 //!
 //! ```
 //! let dp = atmega_hal::Peripherals::take().unwrap();
--- a/mcu/atmega-hal/src/adc.rs
+++ b/mcu/atmega-hal/src/adc.rs
@@ -1,4 +1,28 @@
 //! Analog-to-Digital Converter
+//!
+//! # Example
+//!
+//! Complete example source code can be found in the
+//! [repository](https://github.com/Rahix/avr-hal/blob/main/examples/atmega2560/src/bin/atmega2560-adc.rs).
+//!
+//! ```
+//! let dp = atmega_hal::Peripherals::take().unwrap();
+//! let pins = atmega_hal::pins!(dp);
+//!
+//! let mut adc = Adc::new(dp.ADC, Default::default());
+//!
+//! let channels: [atmega_hal::adc::Channel; 4] = [
+//!     pins.pf0.into_analog_input(&mut adc).into_channel(),
+//!     pins.pf1.into_analog_input(&mut adc).into_channel(),
+//!     pins.pf2.into_analog_input(&mut adc).into_channel(),
+//!     pins.pf3.into_analog_input(&mut adc).into_channel(),
+//! ];
+//!
+//! for (index, channel) in channels.iter().enumerate() {
+//!     let value = adc.read_blocking(channel);
+//!     ufmt::uwrite!(&mut serial, "CH{}: {} ", index, value).unwrap();
+//! }
+//! ```
 
 use crate::port;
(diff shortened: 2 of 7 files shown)


examples/atmega2560/src/bin/atmega2560-eeprom.rs
```diff
@@ -0,0 +1,39 @@
...
+        Baudrate::<crate::CoreClock>::new(57600),
+    );
+
+    let mut eeprom = Eeprom::new(dp.EEPROM);
+
+    let mut boot_count = eeprom.read_byte(BOOT_COUNT_OFFSET);
+    boot_count = boot_count.wrapping_add(1);
+    eeprom.write_byte(BOOT_COUNT_OFFSET, boot_count);
+
+    ufmt::uwriteln!(&mut serial, "Boot count: {}", boot_count).unwrap();
```
Rahix, line 36: A minor thing, but because people like copying example code for production use, let's talk about it:

I think this should be using a `saturating_add()` instead of `wrapping_add()` so the value doesn't cycle back to zero after 256 boots.  The print statement should then display a different message on overflow.  E.g.

```suggestion
    let mut boot_count = eeprom.read_byte(BOOT_COUNT_OFFSET);
    boot_count = boot_count.saturating_add(1);
    eeprom.write_byte(BOOT_COUNT_OFFSET, boot_count);

    if boot_count == u8::MAX {
        ufmt::uwriteln!(&mut serial, "Boot count: >={} (overflow)", boot_count).unwrap();
    } else {
        ufmt::uwriteln!(&mut serial, "Boot count: {}", boot_count).unwrap();
    }
```

mcu/atmega-hal/src/adc.rs
```diff
@@ -1,4 +1,28 @@
 //! Analog-to-Digital Converter
+//!
+//! # Example
+//!
+//! Complete example source code can be found in the
+//! [repository](https://github.com/Rahix/avr-hal/blob/main/examples/atmega2560/src/bin/atmega2560-adc.rs).
```
Rahix, line 6: Also minor, but i'd prefer to have the link be the filename of the example so it's easier to reference offline:

```suggestion
//! Complete example source code can be found in the repository:
//! [`atmega2560-adc.rs`](https://github.com/Rahix/avr-hal/blob/main/examples/atmega2560/src/bin/atmega2560-adc.rs)
```

Same for all the other examples, of course.

mcu/atmega-hal/src/usart.rs
```diff
@@ -1,3 +1,33 @@
+//! USART
+//!
+//! # Examples
+//!
+//! Complete example source code can be found in the
+//! [repository](https://github.com/Rahix/avr-hal/blob/main/examples/atmega2560/src/bin/atmega2560-usart.rs).
+//!
+//! Terminal input/output using [ufmt](https://crates.io/crates/ufmt/):
```
Rahix, line 8: `ufmt` is just used for formatting output.  Maybe leave a note here that `ufmt` is used instead of `core::fmt` because `core::fmt` code quickly grows too large for AVR platforms.

Rahix (commented):
Very nice, thanks a lot!

examples/atmega2560/src/bin/atmega2560-eeprom.rs
armandas: I agree, examples are how people learn, so we should try to show the best code.

The issue with EEPROM is that the data is `0xFF` in the default/erased state, so we need to "initialize" it on first boot. A couple of possible solutions would be:

- Check and stop counting at `u8::MAX - 1`
- Use a larger integer (`u32` should be enough for everyone™)

mcu/atmega-hal/src/usart.rs
armandas: Thanks, let me just copy-paste that.

mcu/atmega-hal/src/usart.rs
```diff
@@ -1,3 +1,33 @@
+//! USART
+//!
+//! # Examples
+//!
+//! Complete example source code can be found in the
+//! [repository](https://github.com/Rahix/avr-hal/blob/main/examples/atmega2560/src/bin/atmega2560-usart.rs).
```
armandas, line 6:
```suggestion
//! Complete example source code can be found in the repository:
//! [`atmega2560-usart.rs`](https://github.com/Rahix/avr-hal/blob/main/examples/atmega2560/src/bin/atmega2560-usart.rs)
```

mcu/atmega-hal/src/eeprom.rs
```diff
@@ -1,3 +1,23 @@
+//! EEPROM
+//!
+//! # Example
+//!
+//! Complete example source code can be found in the
+//! [repository](https://github.com/Rahix/avr-hal/blob/main/examples/atmega2560/src/bin/atmega2560-eeprom.rs).
```
armandas, line 6:
```suggestion
//! Complete example source code can be found in the repository:
//! [`atmega2560-eeprom.rs`](https://github.com/Rahix/avr-hal/blob/main/examples/atmega2560/src/bin/atmega2560-eeprom.rs)
```

mcu/atmega-hal/src/i2c.rs
```diff
@@ -1,3 +1,24 @@
+//! I2C
+//!
+//! # Example
+//!
+//! Complete example source code can be found in the
+//! [repository](https://github.com/Rahix/avr-hal/blob/main/examples/atmega2560/src/bin/atmega2560-i2cdetect.rs).
```
armandas, line 6:
```suggestion
//! Complete example source code can be found in the repository:
//! [`atmega2560-i2cdetect.rs`](https://github.com/Rahix/avr-hal/blob/main/examples/atmega2560/src/bin/atmega2560-i2cdetect.rs)
```

mcu/atmega-hal/src/port.rs
```diff
@@ -1,3 +1,22 @@
+//! Port
+//!
+//! # Example
+//!
+//! Complete example source code can be found in the
+//! [repository](https://github.com/Rahix/avr-hal/blob/main/examples/atmega2560/src/bin/atmega2560-blink.rs).
```
armandas, line 6:
```suggestion
//! Complete example source code can be found in the repository:
//! [`atmega2560-blink.rs`](https://github.com/Rahix/avr-hal/blob/main/examples/atmega2560/src/bin/atmega2560-blink.rs)
```

mcu/atmega-hal/src/spi.rs
```diff
@@ -1,3 +1,33 @@
+//! SPI
+//!
+//! # Example
+//!
+//! Complete example source code can be found in the
+//! [repository](https://github.com/Rahix/avr-hal/blob/main/examples/atmega2560/src/bin/atmega2560-spi-feedback.rs).
```
armandas, line 6:
```suggestion
//! Complete example source code can be found in the repository
//! [`atmega2560-spi-feedback.rs`](https://github.com/Rahix/avr-hal/blob/main/examples/atmega2560/src/bin/atmega2560-spi-feedback.rs)
```

mcu/atmega-hal/src/usart.rs
```diff
@@ -1,3 +1,33 @@
+//! USART
+//!
+//! # Examples
```
armandas, line 3:
```suggestion
//! # Example
```

mcu/atmega-hal/src/usart.rs
armandas, line 8:
```suggestion
//! *Note: [ufmt](https://crates.io/crates/ufmt/) is used instead of `core::fmt` because
//! `core::fmt` code quickly grows too large for AVR platforms.*
```

examples/atmega2560/src/bin/atmega2560-eeprom.rs
Rahix: Hmm, I see it is getting even more difficult...  Tell you what, let's merge it as is and then maybe think about improvements in the future.  I smell the solution being a custom "driver" to manage a bootcount, to contain all this complexity...

PR #592, opened 2024-10-20
Add examples to attiny-hal documentation

armandas:
Part of https://github.com/Rahix/avr-hal/issues/590

This one is a bit more tricky, as there are no plain `attiny-hal` examples and I don't have access to an ATtiny board.

For now, I  copied the examples from `atmega-hal` and modified them for `attiny-hal`, referencing ATtiny84 datasheet for pinouts. As a compromise, I link to ATmega examples.

Do you think this is acceptable?


Rahix (approved):
I'm totally fine with doing it this way, it's much better than not having examples at all.

Looks great, thanks a lot :)

Issue #602, opened 2024-11-17
Hot take: the way avr-hal uses features is frustrating


innermatrix:
As I am sure you know, the recommended way to use features is to make them additive, and to avoid mutually exclusive features. However, `avr-hal` relies heavily on non-additive mutually exclusive features. I find this to be a pain. 

This setup works great as long as you are building a crate that generates one executable for one board. It doesn't work great if you are

* writing a library on top of `avr-hal`, because it forces the library into the same disrecommended mutually-exclusive feature behavior. 
* writing a crate that outputs multiple executables for different MCUs

The core issue — as I see it — is not that `avr-hal` is using features, it's the way it's using them. Specifically, the core issue is that `avr-hal` has some symbols whose implementation is selected mutually-exclusively using a feature — for example, `pins` in `atmega-hal`. The fact that `atmega-hal` supports multiple MCUs, but can only ever export one `pins` symbol, is the root of the conflict. 

IMO a better way to manage this would be to use features as they are intended: a feature flag selects whether a piece of functionality is included in the library, but allows multiple features to be selected. For example, I should be able to select both `atmega32u4` and `atmega328` features — and if I do, instead of them fighting over who gets to control the global `pins` symbol, I should get both `atmega32u4::pins` and `atmega328::pins` symbols. 

Taking this all the way up to the board level, what this would mean is that  I would be able to use `arduino-hal.features = ["atmega32u4", "atmega328"]` when importing `arduino-hal`— which would indicate that I want `arduino-hal` to include support for both of those MCUs (using features additively, as they were intended) — and then in my `main.rs` I would write

```
#[cfg(any(feature = "atmega328"))]
use arduino_hal::atmega328 as board;

#[cfg(any(feature = "atmega32u4"))]
use arduino_hal::atmega32u4 as board;
```

I think this would completely eliminate all the shenanigans with mutually-exclusive features, and simplify the experience for anyone trying to use `avr-hal` in a non-trivial way. 

Thoughts?

stappersg:
On Sun, Nov 17, 2024 at 02:02:09PM -0800, Iris Artin wrote:
> 
> Thoughts?
> 

Patches welcome

stappersg:
On Wed, Nov 20, 2024 at 05:53:21PM -0800, Iris Artin wrote:
> @stappersg I have started this work in https://github.com/Rahix/avr-hal/pull/603

Yep.  Due "Notify all" I saw it already.


Please don't wait for me
[...]

PR #603, opened 2024-11-21
Make attiny_hal::pins! and atmega_hal::pins! feature-agnostic

innermatrix:
This removes the feature-specific (and therefore MCU-specific) `pins!` macro and replaces it with a feature-agnostic `pins!` macro + an MCU-specific wrapper function. 

This is initial work to enable a refractor that will switch all crates in `arduino-hal` to use additive features, as is the recommended best practice. See https://github.com/Rahix/avr-hal/issues/602 for more context. 

Testing: 

* Verified that `cargo build -p arduino-hal --features <board>` succeeds for all boards
* Verified that `cargo build --bins` succeeds for every `example/*`

Using stacked PRs — see https://github.com/innermatrix/avr-hal/pull/1#issuecomment-2489875246 for a preview of the follow-on work in progress.

### DIFF
--- a/avr-hal-generic/src/port.rs
+++ b/avr-hal-generic/src/port.rs
@@ -586,7 +586,7 @@ macro_rules! impl_port_traditional {
 
             impl Pins {
                 pub fn new(
-                    $(_: $port,)+
+                    $(_: &$port,)+
                 ) -> Self {
                     Self {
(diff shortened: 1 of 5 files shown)


innermatrix:
Canceling this PR because I came up with a way of doing this that doesn't require code changes in clients of `arduino-hal`

PR #605, opened 2024-11-24
Refactored `attiny_hal` for additive features

innermatrix:
See https://github.com/Rahix/avr-hal/issues/602 for context and overall plan. This PR refactors `attiny_hal` to make conflicting globals optional (but available by default). 

Changes:

* Added `deprecated-globals` feature (on by default)
* Added `<mcu>-no-deprecated-globals` features for all MCU, which can be used to build the crate for multiple MCUs
* Added a module for each supported MCU


tones111:
The issue is a bit old at this point, but perhaps this restructure could be an item in #130 under "Redesign HAL APIs/internals for more flexibility towards supported devices".

innermatrix:
@tones111 Thanks for pointing that out! Yes, this work is very much aligned with that refactor. (There are many other desired / planned improvements that are being tracked by #130, and I am not trying to address those here; specifically, I am not introducing any API changes here, I am just restructuring modules and macros, which I think has organizational value in its own right.)

stappersg:
On Wed, Jan 01, 2025 at 04:28:01AM -0800, Iris Artin wrote:
> Thanks for pointing that out!
> Yes, this work is very much aligned with that refactor.

This issue now has a reference to #623.

innermatrix:

PR #606, opened 2024-11-24
Refactored `atmega_hal` for additive features

innermatrix:
See https://github.com/Rahix/avr-hal/issues/602 for context and overall plan. This PR refactors `atmega_hal` to make conflicting globals optional (but available by default). This PR is stacked on top of #605; only the last commit is new in this PR. 

Changes:

* Added `deprecated-globals` feature (on by default)
* Added `<mcu>-no-deprecated-globals` features for all MCU, which can be used to build the crate for multiple MCUs
* Added a module for each supported MCU


innermatrix:

PR #612, opened 2024-12-09
Turn off deprecated globals by default in `attiny_hal`, `atmega_hal`, and `arduino_hal`

innermatrix:
See https://github.com/Rahix/avr-hal/issues/602 for context and overall plan. This PR changes the `deprecated-globals` feature in `attiny_hal`, `atmega_hal`, and `arduino_hal` to off by default (but available as opt-in). This PR is stacked on top of #611; only the last commit is new in this PR. 

Changes:

* Changed `deprecated-globals` to off by default
* Bumped versions to 0.3


innermatrix:

PR #613, opened 2024-12-09
Remove deprecated globals

innermatrix:
See https://github.com/Rahix/avr-hal/issues/602 for context and overall plan. This PR removes the `deprecated-globals` feature from all crates, and updates example code to match. This PR is stacked on top of #612; only the last commit is new in this PR. 

Changes:

* Remove the `deprecated-globals` feature from `attiny_hal`, `atmega_hal`, and `arduino_hal`
* Updated all code examples
* Bumped versions to 0.4


innermatrix:

PR #619, opened 2024-12-31
[avr-hal-generic] remove rustc version check

tones111:
AVR compilation requires a nightly compiler so it's unlikely users are still on a pre-1.59 version (Feb 2022).


stappersg:
image

Make it two separate merge requests and give both commits good commit messages.

tones111:
As requested, I've updated the commit message to more clearly describe the motivation and split the dependent change into #624 (but I think that makes the review more complicated due to GitHub's lack of support for dependent reviews).

Rahix (approved):
Fair enough, I think we no longer need to keep this back-compat code here.  Users of current versions of `avr-hal` will be on more recent compilers I would assume.  If not, we can add it back in when someone complains.

PR #620, opened 2024-12-31
add support for SparkFun Pro Micro 3.3V (8MHz)

### LINKED ISSUE #614: SparkFun Pro Micro assumes 16 MHz clock for both 3.3V and 5V variants
(I am pretty new to the whole Rust ecosystem so please bear with me if I am overlooking something obvious)

The SparkFun Pro Micro, just like the SparkFun Pro Mini has a 3.3V/8MHz and a 5V/16MHz variant. For the Pro Mini, there's two different feature flags (`sparkfun-promini-5v` and `sparkfun-promini-3v3`) which result in different clock speeds in [clock.rs](https://github.com/Rahix/avr-hal/blob/949130a5377b9ec5426b80699e56f3e876b6be41/arduino-hal/src/clock.rs#L25-L31):
```rust
        feature = "sparkfun-promini-5v",
        feature = "trinket-pro",
        feature = "nano168",
    ))]
    pub type DefaultClock = avr_hal_generic::clock::MHz16;
    #[cfg(any(feature = "trinket", feature = "sparkfun-promini-3v3"))]
    pub type DefaultClock = avr_hal_generic::clock::MHz8;
```

In contrast, the `sparkfun-promicro` flag does not differentiate between the two variants and always assumes a 16 MHz clock.
It seems as if both variants are meant to be supported, since [boards.toml](https://github.com/Rahix/avr-hal/blob/949130a5377b9ec5426b80699e56f3e876b6be41/ravedude/src/boards.toml#L159-L165) explicitly mentions both versions:
```toml
[promicro.usb-info]
    port-ids = [
        { vid = 0x1B4F, pid = 0x9205 }, # 5V
        { vid = 0x1B4F, pid = 0x9206 }, # 5V
        { vid = 0x1B4F, pid = 0x9203 }, # 3.3V
        { vid = 0x1B4F, pid = 0x9204 }, # 3.3V
    ]
```
Treating both variants the same seems to be incorrect, the 3.3 V variant has a 8 MHz clock and should be treated differently than the 5V variant (SparkFun boards.txt which mentions this difference [here](https://github.com/sparkfun/Arduino_Boards/blob/96a65e9e0cd4600792a1f5f4cff196e655de5ae8/sparkfun/avr/boards.txt#L107)).

Should both variants be behind different feature flags so they are treated differently?

tones111:
This is an attempt to resolve #614, but I don't have any hardware to run against.  Lets see if the CI tests pass.


stappersg (changes requested):
Short:  reject

Longer:
Make the commit matching `add support for SparkFun Pro Micro 3.3V (8MHz)`. Current is the commit `add support for SparkFun Pro Micro 3.3V (8MHz) and a rename of the existing SparkFun Pro Micro`.

tones111:
I was a bit lazy and didn't rename the examples.  Now that the CI is running I see the failures and will get them fixed.

PR #621, opened 2025-01-01
[arduino-hal] remove embedded-hal-v0 dependency

tones111:


arduino-hal/src/delay.rs
```diff
@@ -11,7 +11,8 @@ pub type Delay = avr_hal_generic::delay::Delay<crate::DefaultClock>;
 /// Busy-loop for the given time.  This function assumes the default clock speed defined by
 /// [`arduino_hal::DefaultClock`][crate::DefaultClock].
 pub fn delay_ms(ms: u16) {
-    Delay::new().delay_ms(ms)
+    // TODO: update signature to match embedded-hal
```
tones111, line 14: I would really like to fix this here, but then this becomes a breaking change and I'm not aware of a good mechanism to identify what code within `avr-hal` will fail to compile as a result.  Manually compiling every example is an inefficient strategy.

Rahix: Just update the PR and see what CI makes of it.  I think that's the most conclusive test for "does this change break anything?".

tones111: I must have forgotten to hide the comment.  I made the change in a second commit and used @stappersg loop method to verify the examples compile.  This PR should be complete.

Rahix (approved):
Looks great, thanks a lot!

PR #616, opened 2024-12-29
Fix doctests in `attiny-hal`

innermatrix:
This PR fixes all existing doctests in `attiny-hal` and adds `cargo test --doc` to CI for `attiny-hal`. Stacked on https://github.com/Rahix/avr-hal/pull/605.

PR #624, opened 2025-01-02
avr-hal-generic: prefer compilation error for unsupported targets

tones111:
The build.rs logic is only used to detect a relatively old compiler version (Feb 2022). Since a nightly build is required to target AVR it's unlikely anyone is benefiting from this check.  By removing it we reduce dependencies and implementation complexity.

### DIFF
--- a/avr-hal-generic/src/delay.rs
+++ b/avr-hal-generic/src/delay.rs
@@ -57,11 +57,6 @@ fn busy_loop(mut c: u16) {
     }
 }
 
-#[cfg(not(target_arch = "avr"))]
-fn busy_loop(_c: u16) {
-    unimplemented!("Implementation is only available for avr targets!")
-}
-
 // Clock-Specific Delay Implementations ----------------------------------- {{{
 impl delay_v0::DelayUs<u16> for Delay<crate::clock::MHz24> {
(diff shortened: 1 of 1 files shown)


tones111:
As requested this commit has been split from #619, however, since it depends on the previous change I've stacked it.  As discussed [here](https://github.com/Rahix/avr-hal/pull/616#issuecomment-2565796887) GitHub doesn't really support splitting dependent changes across multiple reviews.

Rahix:
So far we've tried to keep `avr-hal` building on non-avr targets.  The motivation was mainly to allow running tests on other platforms without having to remove all `avr-hal` dependencies from an application.  Adding a compile-time error would of course break this.

---

On a different note, even if we were to introduce a compile-time error, I do not think that this is the proper way to do it:  The resulting error message is extremely confusing and doesn't hint at the underlying problem at all.  The proper way would be a crate-global cfg'ed `compile_error!()`, similar to this one:

https://github.com/Rahix/avr-hal/blob/3c089795cadbbc7fa83f45958128689fee7ba1e4/mcu/atmega-hal/src/lib.rs#L27-L48

The tricky part is ensuring that such an error also short-circuits all other compiler errors that are caused by the same reason, to ensure clean and easy-to-navigate diagnostics.

PR #628, opened 2025-01-13
arduino-uno: simple-servo example

ddaniel27:
Hey, first, thanks for this repo! I'm a rust beginner and to be honest I want to use it for some embedded programming hobby, no professional use right now. So I will be really happy to get some honest feedback about this code.

I know there is a lot of unsafe blocks, so it will be great if you can point some ways to deal with that, improve the code or anything honestly. Basically I just read the official arduino servo library where uses interruptions to update the servos when needed. I translate most of the code and after a lot of frustration, I ended with this code that allows the user to create a servo, attach it and set some angles in a very simple way. 

Let me know if more info or changes are needed.


kbleeke:
I am in a similar situation that I want to control some servos for a hobby project. I have Rust experience, although not much on embedded. 

Some initial API feedback: I think any servo controller has to somehow take ownership of TC1. This seems to be the idea behind most things in embedded Rust that to use a resource, you have to pass the peripheral handle. As the timers on the atmega328p are all share the timer and pwm modes, we have to exclusively reserve this timer for the serve control and disallow the creation of e.g. `Timer1Pwm` at the same time.

PR #631, opened 2025-01-20
Update ravedude to clap 3.0

innermatrix:
Because structopt is in maintenance mode, and clap 3.0 is the replacement. Future work: update clap to 4.0.

Clap 3 eliminated the flags vs options distinction and if there are no strong feelings, I don't see a reason to reinstate it ourselves. 

`--help` output before

```
ravedude 949130a5377b
ravedude is a rust wrapper around avrdude for providing the smoothest possible development experience with rust on AVR
microcontrollers.

ravedude is primarily intended to be used as a "runner" in the cargo configuration.

USAGE:
    ravedude [FLAGS] [OPTIONS] [ARGS]

FLAGS:
        --dump-config      
            Utility flag for dumping a config of a named board to TOML

    -c, --open-console     
            After successfully flashing the program, open a serial console to see output sent by the board and possibly
            interact with it
        --debug-avrdude    
            Print the avrdude command that is executed for flashing the binary

    -h, --help             
            Prints help information

    -V, --version          
            Prints version information


OPTIONS:
    -b, --baudrate <baudrate>          
            Baudrate which should be used for the serial console

    -P, --port <port>                  
            Overwrite which port to use. By default ravedude will try to find a connected board by itself [env:
            RAVEDUDE_PORT=]
    -d, --reset-delay <reset-delay>    
            This assumes the board is already resetting. Instead of giving the reset instructions and waiting for user
            confirmation, we wait the amount in milliseconds before proceeding. Set this value to 0 to skip the board
            reset question instantly

ARGS:
    <BINARY>           
            The binary to be flashed.
            
            If no binary is given, flashing will be skipped.
    <LEGACY BINARY>    
            Deprecated binary for old configurations of ravedude without `Ravedude.toml`. Should not be used in newer
            configurations
```

and after

```
ravedude 259f246f88f2
ravedude is a rust wrapper around avrdude for providing the smoothest possible development
experience with rust on AVR microcontrollers.

ravedude is primarily intended to be used as a "runner" in the cargo configuration.

USAGE:
    ravedude [OPTIONS] [ARGS]

ARGS:
    <BINARY>
            The binary to be flashed.
            
            If no binary is given, flashing will be skipped.

    <LEGACY BINARY>
            Deprecated binary for old configurations of ravedude without `Ravedude.toml`. Should not
            be used in newer configurations

OPTIONS:
        --dump-config
            Utility flag for dumping a config of a named board to TOML

    -c, --open-console
            After successfully flashing the program, open a serial console to see output sent by the
            board and possibly interact with it

    -b, --baudrate <BAUDRATE>
            Baudrate which should be used for the serial console

    -P, --port <PORT>
            Overwrite which port to use. By default ravedude will try to find a connected board by
            itself
            
            [env: RAVEDUDE_PORT=]

    -d, --reset-delay <RESET_DELAY>
            This assumes the board is already resetting. Instead of giving the reset instructions
            and waiting for user confirmation, we wait the amount in milliseconds before proceeding.
            Set this value to 0 to skip the board reset question instantly

        --debug-avrdude
            Print the avrdude command that is executed for flashing the binary

    -h, --help
            Print help information

    -V, --version
            Print version information
```


Rahix (approved):
Thanks a lot!

In the long run, I'd want to move away from clap as `ravedude` isn't really primarily used via CLI arguments and clap just adds so much dependency bloat...  With other tools, I've been using the `gumdrop` CLI argument parser quite successfully, maybe that's a good choice for ravedude as well.

But for the time being, staying up to date sure sounds better than legacy dependencies.

PR #634, opened 2025-02-22
Update Rust toolchain to nightly-2025-03-03

### LINKED ISSUE #537: cargo build - error[E0658] - proc_macro::Literal::byte_character(byte)

tones111:
This change bumps the toolchain to a recent nightly that incorporates LLVM codegen improvements and avr-none target.

Much thanks to @Patryk27 for all your hard work

closes #537


tones111:
I'd like to phone a friend (@Patryk27)

When I build locally I get an error like
```/usr/bin/avr-ld: avr architecture of input file `/tmp/rustcyPlIoS/symbols.o' is incompatible with avr:51 output```
which led me [here](https://stackoverflow.com/questions/51505057/avr-ld-avr51-architecture-of-input-file-main-o-is-incompatible-with-avr-outp) and suggests a `-mavr51` flag needs to be passed to the linker.

I'm not sure where symbols.o comes from.  Maybe we need to (somehow) pass extra flags when building the `core` library?

It appears the `avr-gcc` on the CI system [doesn't understand that flag](https://github.com/Rahix/avr-hal/actions/runs/13477769806/job/37659040553?pr=634#step:5:101)

my versions
```
$ avr-gcc --version
avr-gcc (GCC) 14.2.0
$ avr-ld --version
GNU ld (GNU Binutils) 2.43
```

**Comment by Patryk27** [comment]
> [...] suggests a -mavr51 flag needs to be passed to the linker.

Yes, this is supposed to be done automatically by rustc:

https://github.com/rust-lang/rust/blob/bb2cc59a2172d6e35c89b409a4e6b5058d9039d7/compiler/rustc_codegen_ssa/src/back/linker.rs#L580

I'll take a look later at what's going on here!

File: .github/workflows/ci.yml [hunk 1]
```diff
@@ -110,19 +110,19 @@ jobs:
...
-        run: cd "examples/${{ matrix.m.name }}" && cargo build --bins
+        run: cd "examples/${{ matrix.m.name }}" && cargo build --release --bins
       - name: Compile board crate (without examples)
         if: "${{ matrix.m.type == 'board' && !matrix.m.examples }}"
-        run: cd "arduino-hal/" && cargo build --features "${{ matrix.m.name }}"
+        run: cd "arduino-hal/" && cargo build --release --features "${{ matrix.m.name }}"
       - name: Test-compile HAL crate for an MCU
         if: "${{ matrix.m.type == 'mcu' }}"
-        run: cd "mcu/${{ matrix.m.crate }}" && cargo build --features "${{ matrix.m.name }}" -Z build-std=core --target "../../avr-specs/avr-${{ matrix.m.spec }}.json"
+        run: cd "mcu/${{ matrix.m.crate }}" && cargo build --release --features "${{ matrix.m.name }}" -Z build-std=core
```
Patryk27, line 125: shouldn't we have something like `--target=avr-none -C target-cpu={{matrix.something}}` in here?

Rahix: Yeah, with `avr-none` being upstream now, we should start making use of it everywhere.  I do wonder, however, what to do with the other target spec changes we currently do:

https://github.com/Rahix/avr-hal/blob/d1782c69236dcd4cb73dd1d7c414136cced86091/avr-specs/sync-from-upstream.py#L57-L64

https://github.com/Rahix/avr-hal/blob/d1782c69236dcd4cb73dd1d7c414136cced86091/avr-specs/sync-from-upstream.py#L103

Patryk27: `max-atomic-width` is already set to 16:

https://github.com/rust-lang/rust/blob/ad27045c31a9f37ad7d44ca2a403de52d1a896d3/compiler/rustc_target/src/spec/targets/avr_none.rs#L25

I'm not sure on `no-default-libraries`, though - gotta look at it.

As for the extra link arguments, they can be always passed via `RUSTFLAGS` (e.g. through `.cargo/config`) or even through a `build.rs` script.

tones111: I tried adding `-C default-linker-libraries` to Cargo.toml for one of the problematic cpus and while it removed the `-nodefaultlibs` argument from the `avr-gcc` invocation the linker error remains.

tones111:
Pinging @Rahix for another look now that the CI is passing.

Rahix:
At last!  Thanks everyone, especially @Patryk27, for the effort!  I just smoke tested the uno examples to make sure we're not masking any obvious regressions - see below for documentation.

- [ ] `uno-16chan-servo-driver`
- [ ] `uno-74hc595`
- [x] `uno-adc`
- [x] `uno-blink`
- [x] `uno-blink-embedded-hal`
- [x] `uno-eeprom`
- [x] `uno-ext-interrupt`
- [ ] `uno-hc-sr04`
- [x] `uno-i2cdetect` ([^1])
- [ ] `uno-infrared`
- [ ] `uno-manual-servo`
- [x] `uno-millis`
- [x] `uno-panic` (only builds in release now!)
- [x] `uno-pin-change-interrupt`
- [x] `uno-println`
- [x] `uno-simple-pwm`
- [x] `uno-simple-pwm-embedded-hal`
- [x] `uno-spi-feedback`
- [x] `uno-timer`
- [x] `uno-usart`
- [x] `uno-watchdog`

[^1]: was not verified with actual devices on the bus

If someone is willing and able to run some of the ATtiny examples to gather a few more data points, that would not hurt.

.github/workflows/ci.yml
```diff
@@ -110,19 +101,19 @@ jobs:
...
       - name: Install avr-gcc, binutils, and libc
         run: sudo apt-get install -y avr-libc binutils-avr gcc-avr
       - name: Compile board crate and examples
         if: "${{ matrix.m.type == 'board' && matrix.m.examples }}"
-        run: cd "examples/${{ matrix.m.name }}" && cargo build --bins
+        run: cd "examples/${{ matrix.m.name }}" && cargo build --release --bins
       - name: Compile board crate (without examples)
         if: "${{ matrix.m.type == 'board' && !matrix.m.examples }}"
-        run: cd "arduino-hal/" && cargo build --features "${{ matrix.m.name }}"
+        run: cd "arduino-hal/" && cargo build --release --features "${{ matrix.m.name }}"
```
Rahix, line 113: I assume you added `--release` here because some of the examples no longer build without release-mode, right?

From what I noticed, it is mainly the usual problem of panic bloat that gets cut down a bit more in release builds where less panicking code-paths remain.

README.md
```diff
@@ -47,7 +47,7 @@ Go into `./examples/arduino-uno` (or the directory for whatever board you want),
 cd examples/arduino-uno
 
 # Build and run it on a connected board
-cargo run --bin uno-blink
+cargo run --release --bin uno-blink
```
Rahix, line 50: We really shouldn't require release builds.  If panic bloat is more of a problem now, we'll have to be more explicit about this topic or, worst-case, adjust the dev-profile even further.

.github/workflows/ci.yml
tones111: Correct, some of the example builds were running out of space so building for release feels like a better default. This is consistent with the AVR platform notes.
> Note that since AVRs have rather small amounts of registers, ROM and RAM, it's 
 recommended to always use `--release` to avoid running out of space.

I'm not familiar with any workflows that would make use of debug info so documenting `--release` as the default way to build seems ok. If there's a way to remotely gdb that would be great to document.

tones111: `uno-panic` is one of the problematic binaries (see #639).  I agree changing those dev profile flags would be surprising to users as it essentially turns dev into a release build.  At that point I think telling users to explicitly build with `--release` is the better option.

Baseline sizes for a couple binaries are
```
   text	   data	    bss	    dec	    hex	filename
    892	      0	      1	    893	    37d	../../target/avr-none/debug/uno-timer.elf
   1016	      0	      1	   1017	    3f9	../../target/avr-none/release/uno-timer.elf
   1504	      0	      1	   1505	    5e1	../../target/avr-none/release/uno-panic.elf
```

Turning off debug-assertions
```
   text	   data	    bss	    dec	    hex	filename
    908	      0	      1	    909	    38d	../../target/avr-none/debug/uno-timer.elf
   4336	      0	      1	   4337	   10f1	../../target/avr-none/debug/uno-panic.elf
```

Turning off debug-assertions and overflow-checks
```
   text	   data	    bss	    dec	    hex	filename
   1016	      0	      1	   1017	    3f9	../../target/avr-none/debug/uno-timer.elf
   3062	      0	      1	   3063	    bf7	../../target/avr-none/debug/uno-panic.elf
```

Only if we also set `codegen-units=1` does the size of `uno-panic` come down.

For these devices which have a reputation for being extremely space limited I fail to understand the value in trying to support the debug profile.  However, tracking code size regressions for release builds seems like a worthwhile endeavor.  Panic bloat will become increasingly problematic as application complexity scales beyond the small examples here.

tones111: After more investigation I've narrowed the debug bloat down to [printing the panic information](https://github.com/Rahix/avr-hal/blob/main/examples/arduino-uno/src/bin/uno-panic.rs#L37-L43).  Debug binary size is small when commented out, but as soon as we try to format anything from the `PanicInfo` (even just the line number) we get the linker error.

Rahix: > After more investigation I've narrowed the debug bloat down to printing the panic information.

Yeah, this is what I suspected :( The problem is called panic bloat - the optimizer isn't smart enough to see what information is actually used by the panic handler and what isn't.  So it all stays in the binary (and even worse, gets places in SRAM).

> Turning off debug-assertions and overflow-checks
> Only if we also set codegen-units=1 does the size of uno-panic come down.

The optimizer can't work across codegen-units so this is also not surprising.  I wouldn't care about setting `codegen-units=1`, but overflow-checks and debug-assertions really need too stay...

---

One more thought: Maybe we just shouldn't advertise the use of panic info at all.  Yes, this sucks, but until upstream rust-lang comes up with a solution, we cannot do anything about this.  Best we could do is adopt an approach similar to defmt and sidestep the standard panicking machinery where possible.

Rahix: >  At that point I think telling users to explicitly build with --release is the better option.

Not allowing dev builds sucks for correctness - suddenly you won't ever notice accidental integer overflows, etc.

I personally would rather accept not knowing the origin of a panic than silent data corruption.

tones111: > Maybe we just shouldn't advertise the use of panic info at all

HAL crates have the luxury of providing small targeted examples that may be able to link with the limited panic bloat.  I don't think this property will hold for the majority of applications that will depend on this crate. Any non-trivial real world logic is going to have panicking paths that the user won't be able to ignore, so, I think it would be misleading to promote debug builds here.

I agree disabling overflow panics is unfortunate but holding back this PR prevents some targets from being able to create binaries at all. Even if the embedded Rust community banded together and raised this as an issue I'm not sure there's much the compiler team can do to improve the situation.  I certainly don't see any progress happening in the short term.

tones111:
This review has stagnated.  @Rahix What's the path forward for getting the toolchain updated?

Rahix (approved):
Hi, sorry for the silence :(  I've unfortunately been kept from continuing here for quite a while.

To get things moving, I've decided the best course of action is to merge your PR as it stands and then I will apply my changes on top.  This should get us forward as fast as possible.

See my (coming) responses to the other review comments for the rationale of my solution to the discussed problems.

Thanks so much again for everyones work here and also for your patience...

.github/workflows/ci.yml
Rahix: I think I wasn't describing the problem in adequate depth earlier and this led to a bit of a misunderstanding.

The problem isn't the existence or amount of panicking code-paths - the problem is instead the access to panic info in the panic handler.  As soon as we touch the `PanicInfo` argument, the optimizer can no longer optimize out the panic messages and surrounding formatting code.  This is what causes the big bloat.

If we do not access the `PanicInfo`, everything is "fine" and this is what I have now done in commit a060a430d22b6e68b63bc628a5ff751ec2b63499.  The downside to this is that the panic handler can no longer give a hint on the precise location of the panic.  But that is unfortunately what we have to live with for now...

A medium-term solution is to incorporate the tricks done by [defmt](https://defmt.ferrous-systems.com/) to sidestep panic bloat.  Essentially it boils down to controlling panicking code-paths with custom code.

PR #642, opened 2025-03-10
Fix toggle() for some old microcontrollers

andber1:
Some microcontrollers do not support toggling the output via the PIN register. On these microcontrollers the PIN register is readonly and toggling is done by reading and modifying the PORT register.

| Microcontroller | atomic toggle supported? |
| --- | --- |
| atmega48p |        ✅ |
| atmega16 |         ❌ |
| atmega164pa |      ✅ |
| atmega168 |        ✅ |
| atmega328p |       ✅ |
| atmega328pb |      ✅ |
| atmega32a |        ❌ |
| atmega32u4 |       ✅ |
| atmega2560 |       ✅ |
| atmega128a |       ❌ |
| atmega1280 |       ✅ |
| atmega1284p |      ✅ |
| atmega8 |          ❌ |
| atmega88p |        ✅ |
| attiny2313 |       ✅ |
| attiny167 |        ✅ |
| attiny84 |         ✅ |
| attiny85 |         ✅ |
| attiny88 |         ✅ |

I tested this on an ATMega16.

Related to #501.


andber1:
Thanks very much for your detailed explanations. That is very helpful. I will try to implement your suggestions and update this pull request soon.

andber1:
> but you need to verify that this actually lowers to atomic register accesses and it should be evaluated whether this leads to less instructions than just wrapping your modify() in a critical section.

The if else solution needs one more instruction than the critical section solution. So I used the critical section here.

Rahix:
> On these microcontrollers the PIN register is readonly and toggling is done by reading and modifying the PORT register.

I think this is unfortunately not a reliable indicator.  To give two examples, here is an excerpt from the ATmega328P datasheet, stating that it does support the feature:

image

And this is from the ATmega168 datasheet:

image

I think the only way to figure this out is to check each datasheet for a similar passage :( I would assume that most chips do support this though, only that one very old family seems not to.

andber1:
You are right. The following microcontrollers do indeed support this feature: atmega48p, atmega88p, atmega168, atmega328p, attiny88
I have updated the code and the table above.

Rahix (approved):
Very nice, thanks a lot for fixing this!

PR #645, opened 2025-03-11
Add ATmega16 support

andber1:
It is pretty much the same as ATmega164pa with some exceptions. I checked it against the datasheet and it should work. However, I have not tested all features on my board.


Rahix (approved):
Looks great, thanks a lot!

Issue #601, opened 2024-11-11
ravedude: Add a check if avrdude is even installed


Rahix:
Add a check to ravedude on whether avrdude is even available.  If not, the error message should give pointers on where to get the correct avrdude version.

(From #600)

AlternativleyCoolSlimbo:
I'll claim this.
And this time, I won't vanish off the face of the earth, leaving someone else to pick up the slack (sorry Rahix, and thank you armandas)!

To add to your initial part of the request, are you suggesting different error messages, dependant on platform, similar to what's already present in the readme?

Rahix:
> To add to your initial part of the request, are you suggesting different error messages, dependant on platform, similar to what's already present in the readme?

Yeah, something that helps the particular user on getting their setup working.  So platform-specific for sure and ideally all the necessary steps to get it running.  For Linux, I'd just cover the distros we mention already. Any special-distro users should be advanced enough to figure it out themselves.

AlternativleyCoolSlimbo:
Roger roger. I'll also include making sure `avr-gcc` is mentioned for the error message.

PR #649, opened 2025-03-26
Feature Addition: ravedude: Add a check if avrdude is even installed 

AlternativleyCoolSlimbo:
See #601 for details.

These are more or less copy-pasted from the main readme.

### DIFF
--- a/ravedude/Cargo.toml
+++ b/ravedude/Cargo.toml
@@ -21,4 +21,6 @@ serde = { version = "1.0.197", features = ["serde_derive"] }
 toml = "0.8.11"
 either = "1.10.0"
 clap = { version = "4.0.0", features = ["derive", "env"] }
+
+[target.'cfg(windows)'.dependencies]
 terminal-link = "0.1.0"
--- a/ravedude/Cargo.toml
+++ b/ravedude/Cargo.toml
@@ -21,3 +21,4 @@ serde = { version = "1.0.197", features = ["serde_derive"] }
 toml = "0.8.11"
 either = "1.10.0"
 clap = { version = "4.0.0", features = ["derive", "env"] }
+terminal-link = "0.1.0"
--- a/ravedude/src/avrdude/mod.rs
+++ b/ravedude/src/avrdude/mod.rs
@@ -14,6 +14,28 @@ pub struct Avrdude {
 }
 
 impl Avrdude {
+    fn avrdude_installed((req_major, req_minor): (u8, u8)) -> anyhow::Result<()> {
+        let output = process::Command::new("avrdude").arg("-?").output()?;
+        let status = output.status;
+        if !status.success() {
+            // Response should specify wrong version,
+            // but then platform specific advice on
+            // what to use instead.
+            let default_error = format!(
+                "Avrdude is not installed. Please install v{}.{}, in accordance with the below instructions.",
+                req_major, req_minor
+            );
+            #[cfg(target_os = "windows")]
+            {
+                anyhow::bail!(
+                    "{} \n\
+                    Use winget(https://learn.microsoft.com/en-us/windows/package-manager/winget/) on Windows 10 and 11.\n \
+                    \"winget install AVRDudes.AVRDUDE\"\n\
+                    On older systems, use scoop.", );
+            }
+        }
+    }
+    
     fn get_avrdude_version() -> anyhow::Result<(u8, u8)> {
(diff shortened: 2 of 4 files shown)


ravedude/Cargo.toml
```diff
@@ -21,3 +21,4 @@ serde = { version = "1.0.197", features = ["serde_derive"] }
 toml = "0.8.11"
 either = "1.10.0"
 clap = { version = "4.0.0", features = ["derive", "env"] }
+terminal-link = "0.1.0"
```
Rahix, line 24: Personally, I am not a fan of doing such fancy things here, especially as it is only for a very niche case (when `avrdude` is not installed).  Worst case, someone has a terminal that doesn't support such links and they get broken output...

But if you insist, I am fine with keeping this in for Windows.  In that case, please make the dependency windows-only, though:

```suggestion

[target.'cfg(target_os = "windows")'.dependencies]
terminal-link = "0.1.0"
```

ravedude/src/avrdude/mod.rs
```diff
@@ -14,6 +13,61 @@ pub struct Avrdude {
...
+            .status();
+        let gcc_output = process::Command::new("avr-gcc")
+            .arg("-v")
+            .stderr(std::process::Stdio::null())
+            .status();
+        if dude_output.is_err() || gcc_output.is_err() {
+            // Response should specify wrong version,
+            // but then platform specific advice on
+            // what to use instead.
+            let default_error = format!("{} and/or {} is not installed.\nPlease install the latest version, in accordance with the below instructions.\nFor more help, visit {}. \n", "Avrdude".bright_black(), "avr-gcc".bright_black(), terminal_link::Link::new("here", "https://github.com/Rahix/avr-hal/wiki/Setting-up-environment"));
```
Rahix, line 29: Please reformat with multi-line string to keep line character limit.

Rahix (commented):
Hi, thanks a lot!  Two things in addition to my comments below:

- What platforms did you test this on?
- Can you please capture the output of the error on each platform (by just hacking the `#[cfg()]`s) and paste it here, so it's easier to understand what it will look like to users?

Rahix:
(Once this PR is merged, I will finally cut a new release of `ravedude`, with all the long-overdue changes)

ravedude/Cargo.toml
AlternativleyCoolSlimbo: From what I've found, most modern (5 or so years) terminal will work.

And I tested on original windows command prompt (and not windows terminal), it seems to just hide the link part and the text is shown as normal without any hyperlink reference.

AlternativleyCoolSlimbo:
Linux Command Preview:
linux
Mac Command Preview:
mac
Windows Command Preview:
windows

ravedude/src/avrdude/mod.rs
```diff
@@ -14,6 +13,67 @@ pub struct Avrdude {
...
+                    brew tap osx-cross/avr\n \
+                    brew install avr-gcc avrdude"
+                        .bright_black()
+                );
+            }
+            #[cfg(target_os = "linux")]
+            {
+                anyhow::bail!("{} \n\
+                Use the following command: {}", default_error, "sudo apt install avr-libc gcc-avr pkg-config avrdude libudev-dev build-essential".bright_black());
+            }
```
Rahix, line 71: Sorry, one more thing that I noticed.  There should be a fall-back when none of the platforms match.  Right now, it would just fall through the `if` which is quite unexpected.

The cleanest solution would be to use `cfg_if::cfg_if!{}` here, I think.  Then you can define an else where you just display your `default_error` message.

Alternatively, you could also just add another `anyhow::bail!` after all the platform specific variants which will of course only be reachable when none of the earlier bails hit.  But I think that will generate a `dead_code` warning so it isn't quite as neat.

AlternativleyCoolSlimbo: Unsure of how you wanted the error message.

Given it would be a rare occurrence (on top of an already rare occurrence), I think just pointing them to the wiki is ideal.

Rahix: That will do, thanks a lot!

Rahix (approved):
Thanks again for working on this!

Rahix:
Alright, new version of ravedude is out: https://github.com/Rahix/avr-hal/blob/main/ravedude/CHANGELOG.md#020---2025-04-22

PR #651, opened 2025-04-18
Update toolchain to 2025-04-27

Rahix:
Some floating point intrinsics were added in this newer version which helps on systems where these intrinsics are not provided by avr-gcc.

Cc: @Patryk27

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,7 +101,7 @@ jobs:
       - name: Install Rust
         uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: nightly-2025-03-03
+          toolchain: nightly-2025-04-27
           components: rust-src
       - name: Install avr-gcc, binutils, and libc
         run: sudo apt-get install -y avr-libc binutils-avr gcc-avr
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Install Rust
         uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: nightly-2025-03-03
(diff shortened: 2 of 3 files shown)


Rahix:
@Patryk27, can you check the CI failures please?  Looks like there was a panicking code-path left in one of the intrinsics if I interpret this correctly?

Patryk27:
Status: not really a bug, it's the same case as https://github.com/rust-lang/rust/issues/139737#issuecomment-2816772743.

tl;dr sometimes for unreachable code paths LLVM inserts a special kind of panic-instruction - e.g. on x86 this:
```rust
#[inline(never)]
pub fn foo() {
    unsafe {
        core::hint::unreachable_unchecked();
    }
}
```
... gets compiled into:
```asm
playground::foo:
	ud2
```
... with ud2 being this sort of hardware-level panic / abort instruction.

There's no ud2 equivalent for AVR, so LLVM just inserts a call to a function called `abort`, expecting for user to define what should happen then (we might `loop {}`, write something to uart, blink leds etc.). Notably, this is different from `#[panic_handler]`, because `panic!()` is a Rust-specific construct, while trapping (aborting) is much more low-level (there's no `&PanicInfo` available etc.).

Overall:

> So tl;dr not a bug - at the same time, I understand that 99% of people don't want to be bothered with this, "do whatever" will work just fine for them; I think we might be able to get the best of both worlds (provide a default function, but allow to overwrite it) by utilizing weak symbols (we'd create a weak fn abort() within compiler-builtins).
>
> I'll experiment and report back.

Patryk27:
Ok, seems we're ready!

I've checked `nightly-2025-04-27` and the examples build on my machine now 🙂

Rahix:
Very nice, thanks you!!

PR #655, opened 2025-05-02
Nix update and revamp

### LINKED ISSUE #632: ravedude nix flake is out of date.
When trying to build `ravedude` with nix I get the following error.
```
error: package `toml v0.8.11` cannot be built because it requires rustc 1.70 or newer, while the currently active rustc version is 1.67.1
Either upgrade to rustc 1.70 or newer, or use
cargo update -p toml@0.8.11 --precise ver
where `ver` is the latest version of `toml` supporting rustc 1.67.1
[naersk] cargo returned with exit code 101, exiting
```

This is fixable by updating the `flake.lock` file but it might be better to point naersk to the `rust-toolchain.toml` file.

LuigiPiucco:
Fix #632. I asked there if the second commit of this PR would be accepted, but let's move the discussion here.

This PR updates the ravedude flake and revises it to follow `ravedude/rust-toolchain.toml`. It also adds the latest stable date to the toolchain, to avoid silently using the older version if it's not in the locked `fenix` input (also because it is generally good to specify the date).

Beyond that, it adds a Nix development shell to the root, which similarly follows `rust-toolchain.toml` and provides runtime programs like avr-gcc too. It does not build the project, just helps testing it. For convenience, minimal descriptions of each relevant command are printed when the user enters the shell, since it's built-in functionality of `devshell`.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,2 @@
+source_up_if_exists
+use flake
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,145 @@
+{
+  "nodes": {
+    "devshell": {
+      "inputs": {
+        "nixpkgs": "nixpkgs"
+      },
+      "locked": {
+        "lastModified": 1741473158,
+        "narHash": "sha256-kWNaq6wQUbUMlPgw8Y+9/9wP0F8SHkjy24/mN3UAppg=",
+        "owner": "numtide",
+        "repo": "devshell",
+        "rev": "7c9e793ebe66bcba8292989a68c0419b737a22a0",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "devshell",
+        "type": "github"
+      }
+    },
+    "fenix": {
+      "inputs": {
+        "nixpkgs": "nixpkgs_2",
+        "rust-analyzer-src": "rust-analyzer-src"
+      },
+      "locked": {
+        "lastModified": 1746167999,
+        "narHash": "sha256-18XGHsjk/5H8F0OGUCG56CeeW1u6qQ7tAfQK3azlwWg=",
+        "owner": "nix-community",
+        "repo": "fenix",
+        "rev": "bcbc23a4f3391c1c3657f1847cb693aaea3aed76",
+        "type": "github"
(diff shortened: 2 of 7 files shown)


LuigiPiucco:
At some point we should probably also condition the CI to ignore certain files, because it seems wasteful to run 27 tests on a PR that doesn't change any rust code... :laughing:

Rahix:
> At some point we should probably also condition the CI to ignore certain files, because it seems wasteful to run 27 tests on a PR that doesn't change any rust code... 😆

Heh, yeah.  I thought about this before, but it is really difficult to create sensible conditions.  Beyond the most trivial ones, you'd have to observe the `cfg()` statements in the code to figure out if a CI target is affected by a changeset and that's very difficult to implement.

Rahix (approved):
Looks good to me, as far as I can judge :D Thanks!

Rahix:
Just for my understanding, I will have to ping you/someone whenever I update the toolchain or crate versions so the `flake.lock` is adjusted to match it, right?

LuigiPiucco:
> Just for my understanding, I will have to ping you/someone whenever I update the toolchain or crate versions so the `flake.lock` is adjusted to match it, right?

Yes.

Technically, only if you update the compiler past the date in the `flake.lock`s. So for nightlies (in the root) this will almost always be the case, and for stable (in ravedude) it may take 2 or 3 bumps, if updated frequently. Updating Rust packages shouldn't be a problem unless the ravedude `Cargo.lock` is not up-to-date, and it only matters for ravedude, since the root is not built. In general the maintenance hurdle is summarized by "keep `Cargo.lock` up-to-date and ping a Nix user when changing the toolchain files".

Rahix:
Understood, will do :)

Rahix:
> At some point we should probably also condition the CI to ignore certain files, because it seems wasteful to run 27 tests on a PR that doesn't change any rust code... 😆

Pushed commit f146d0d3baf9 ("ci: Split avr-hal and ravedude CI to reduce superfluous jobs") to at least separate `avr-hal` and `ravedude` CI jobs...  If you want to add more excludes to the CI, feel free to submit a PR.

PR #658, opened 2025-05-07
avr-hal-generic: Fix spurious NACK in I2c::transaction

DmitriyNE:
When calling I2c::transaction with several subsequent reads, like [Operation::Read(..), Operation::Read(..)], all non-final reads should have all their bytes ACKed by the master and NACK should be sent only for the last byte in the chain. Previous implementation violated this contract, which has led to failures in downstream libs. This fix corrects this behaviour.


Rahix (approved):
Nice catch!  Implementation looks great, thanks a lot!

PR #665, opened 2025-05-26
fix ping_device hanging in read mode

### LINKED ISSUE #622: i2cdetect in the read direction hangs/crashes after finding a sensor
I'm using an Arduino Mega 2560 R3 with a MH-Z16 CO2 sensor in I2C mode

I basically just took the sample for I2Cdetect verbatim (just swapped the order of Read/Write)

```rust
#![no_std]
#![no_main]

use arduino_hal::{prelude::*, I2c};
use panic_halt as _;

const NDIR_ADDR: u8 = 0x4D;

#[arduino_hal::entry]
fn main() -> ! {

    let mut i2c = arduino_hal::I2c::new(
        dp.TWI,
        pins.d20.into_pull_up_input(),
        pins.d21.into_pull_up_input(),
        100000,
    );

    ufmt::uwriteln!(&mut serial, "Read direction test:\r").unwrap_infallible();
    i2c.i2cdetect(&mut serial, arduino_hal::i2c::Direction::Read)
        .unwrap_infallible();

    ufmt::uwriteln!(&mut serial, "Write direction test:\r").unwrap_infallible();
    i2c.i2cdetect(&mut serial, arduino_hal::i2c::Direction::Write)
        .unwrap_infallible();

    loop{};
}
```

```
/// serial output
Read direction test:
20: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- --
30: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- --
40: -- -- -- -- -- -- -- -- -- -- -- -- 4d
```

It seems to run until it finds the sensor, and then it freezes.
the Write direction seems to run fine, but the read direction hangs.
I've tried adding some code below the i2cdetect but it doesn't appear as if any of it actually runs.

I've tried using the normal Arduino/C++ code provided by the manufacturer and it worked (connected and sent measurements over serial) so I don't think I've setup the hardware incorrectly, but please let me know if I fumbled something obvious.

I also tried just calling `ping` and it seemed to behave the same way (unsurprisingly..) where the ping goes through, but all following calls to i2c hang/panic(?)

```rust
    let pinged = i2c
         .ping_device(NDIR_ADDR, arduino_hal::i2c::Direction::Read)
         .unwrap();
    ufmt::uwriteln!(&mut serial, "Pinged: {}\r", if pinged { 1 } else { 0 }).unwrap();

    ufmt::uwriteln!(&mut serial, "Read direction test:\r").unwrap_infallible();
// this code doesn't run (except the first row of labels)
    i2c.i2cdetect(&mut serial, arduino_hal::i2c::Direction::Read)
        .unwrap_infallible();
```

bdew:
This fixes a part of #622 (i2cdetect getting stuck)

It looks like you need to actually read after requesting data, otherwise something breaks in the i2c implementation and the next operation will just get stuck forever

### DIFF
--- a/avr-hal-generic/src/i2c.rs
+++ b/avr-hal-generic/src/i2c.rs
@@ -289,6 +289,9 @@ where
     pub fn ping_device(&mut self, address: u8, direction: Direction) -> Result<bool, Error> {
         match self.p.raw_start(address, direction) {
             Ok(_) => {
+                if direction == Direction::Read {
+                    self.p.raw_read(&mut [0], true)?
+                }
                 self.p.raw_stop()?;
                 Ok(true)
             }


Rahix (approved):
After more discussion [here](https://github.com/Rahix/avr-hal/issues/622#issuecomment-3330109360) it seems your fix is the only correct solution to this problem.  Thanks for the patch!

PR #667, opened 2025-06-09
Add some modern avr support

tomtor:
I added an mcu/avrmodern_hal with an implementation for usart and ports, for the attiny402 and attiny1614. Also some code for attiny3224 and avr128db28 (commented out because these are not in the current avr_device crate).

More can be added in the future.

Also added 2 examples and tested them on an attiny402.

It is a minimal implementation, but I think that it would be nice to have at least some code in this repository for the modern AVR range. The added examples/avrmodern/README.md has some generic hints on Rust for these devices (could also be added to the avr-device repository?)


nonik0:
@tomtor No worries. Yes, the avr_device_new will remove avr_spec files. Sorry, I forgot to link [my current branch](https://github.com/nonik0/avr-hal/tree/avr-device-new) that's based off #357. Here's a summary of what I have done in that branch (not too much):
- added support for attiny404/804/1604, mostly @agausmann work, I just added more boilerplate for the others
- rebased to avr_device_new, absorbing svd2rust changes... mostly (I was just looking and saw one macro still had a chunk of code commented out)
- added EEPROM support, using modified asm code from [MegaTinyCore](https://github.com/SpenceKonde/megaTinyCore)
- messy, unfinished ADC code - definitely can remove anything here

Also, a little bit more on my EEPROM impl--last time I had checked the asm in the impl was not compiling with the new build setup (with target=avr-none) in avr_device_new. So as a workaround I was still including the avr-spec file as a target in my project to get the asm to compile. I tried to do the EEPROM impl without asm and only using rust2svd register access but have been unsuccessful after several attempts (compiled code too slow I figure). However, after syncing my fork, it looks like I can now compile the asm without the target avr-spec file. But looks like there are optimizations really needed as just changing the target back also added over 1.1KB to my bin size, presumably due to less optimal compilation/instruction use, I see this showing up in cargo bloat:
```
File  .text    Size             Crate Name
0.0%   4.3%    614B compiler_builtins compiler_builtins::mem::memmove
0.0%   2.3%    334B compiler_builtins compiler_builtins::mem::memcpy
```

Anyway, since you're actively working on this, I am happy to proceed with whatever is easiest for you to do for merging the work because ideally I just to get some level of support in avr-hal first, and then supplement from there. My initial thoughts would be to merge with my existing branch since the design/naming/structure that @agausmann has already done has been reviewed by Rahix and is more up-to-date with the current state of the repo. I am also not sure how to proceed with the EEPROM impl given the above. Let me know if you have any thoughts, I believe the asm should also be the same for the XMEGA 0 and 1 series so should work for you with minimal adjustment.

PR #681, opened 2025-09-22
i2c: Prevent sending stop without data.

### LINKED ISSUE #622: i2cdetect in the read direction hangs/crashes after finding a sensor

booti386:
Sending a stop without sending at least one data packet is disallowed by spec:

> "ATmega48A/PA/88A/PA/168A/PA/328/P megaAVR® Data Sheet"
> Rev. B – 09/2020
> Paragraph 22.3.5 "Combining Address and Data Packets
> into a Transmission":
>
> A transmission basically consists of a START condition, a SLA+R/W,
> one or more data packets and a STOP condition. **An empty message,
> consisting of a START followed by a STOP condition, is illegal.**

What we can do instead is emulating a stop in this case by disabling the i2c unit (clear twcr::twen).
Fixes #622.

### DIFF
--- a/avr-hal-generic/src/i2c.rs
+++ b/avr-hal-generic/src/i2c.rs
@@ -189,11 +189,21 @@ pub trait I2cOps<H, SDA, SCL> {
 
     /// Send a stop-condition and release the bus.
     ///
-    /// This method must only be called after successfully starting a bus transaction.  This method
-    /// does not need to block until the stop condition has actually occured.
+    /// This method must only be called after successfully starting a bus transaction and
+    /// transmitting  at least one data packet. This method does not need to block until the stop
+    /// condition has actually occured.
     ///
     /// **Warning**: This is a low-level method and should not be called directly from user code.
     fn raw_stop(&mut self) -> Result<(), Error>;
+
+    /// Release the bus immediately.
+    ///
+    /// This method can be called anytime. The bus is switched off and all transmissions are
+    /// terminated, regardless of any ongoing operation. This method is useful mostly for probing
+    /// the bus without sending data.
+    ///
+    /// **Warning**: This is a low-level method and should not be called directly from user code.
+    fn raw_force_stop(&mut self) -> Result<(), Error>;
 }
 
 /// I2C driver
@@ -289,7 +299,7 @@ where
     pub fn ping_device(&mut self, address: u8, direction: Direction) -> Result<bool, Error> {
         match self.p.raw_start(address, direction) {
             Ok(_) => {
-                self.p.raw_stop()?;
+                self.p.raw_force_stop()?;
                 Ok(true)
             }
             Err(Error::AddressNack) => Ok(false),
@@ -376,7 +386,11 @@ impl<H, I2C: I2cOps<H, SDA, SCL>, SDA, SCL, CLOCK> embedded_hal_v0::blocking::i2
     fn write(&mut self, address: u8, bytes: &[u8]) -> Result<(), Self::Error> {
         self.p.raw_start(address, Direction::Write)?;
         self.p.raw_write(bytes)?;
-        self.p.raw_stop()?;
+        if bytes.len() > 0 {
+            self.p.raw_stop()?;
+        } else {
+            self.p.raw_force_stop()?;
+        }
         Ok(())
     }
 }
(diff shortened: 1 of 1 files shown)


avr-hal-generic/src/i2c.rs
```diff
@@ -609,6 +631,13 @@ macro_rules! impl_i2c_twi {
                     .write(|w| w.twen().set_bit().twint().set_bit().twsto().set_bit());
                 Ok(())
             }
+
+            #[inline]
+            fn raw_force_stop(&mut self) -> Result<(), Error> {
+                self.twcr
+                    .write(|w| w.twint().set_bit());
```
Rahix, line 638: I know it will be a no-op, but I think there is value in explicitly adding a 

```rust
.twen().clear_bit()
```

here to make it abundantly clear that `raw_force_stop()` works by completely disabling the peripheral.

Rahix (commented):
Hi,

thanks a lot for digging into this an providing a fix!  I think the approach sounds reasonable, however I would like to discuss possible side-effects of doing this before going ahead:

- First of all, I am concerned that this leaves the bus peripheral in a different state until the next operation starts (TWEN is off until the next START).  My understanding from the datasheets is that this will not have visible effects on the external signals.  Does this match your understanding?
- What happens when a new transaction is started immediately after `raw_force_stop()`?  I think we are missing a delay here to ensure the forced stop is actually witnessed by other bus participants correctly in all situations?
- Right now you only updated the old `embedded_hal_v0` impls to call `raw_force_stop()`.  Even though I feel like the real solution should be to forbid empty buffers, I think this kind of fallback is okay for a legacy solution so let's stick with it.
- But I wonder about the impl of the new [`I2c::transaction`](https://docs.rs/embedded-hal/latest/embedded_hal/i2c/trait.I2c.html#tymethod.transaction).  The contract defined upstream does not forbid empty data buffers.  So we could potentially hit the same illegal situation here ([L452](https://github.com/Rahix/avr-hal/pull/681/files#diff-ad5e049947ba3f5c9b2e51324221c391961c6db2df7809ea9140704895e4da4aL452)).  I do not think the forced stop is really appropriate, unless this is really the behavior `embedded-hal` wants here.  But silently not doing the whole transaction because it is empty does not look better either.  What are your thoughts?

Issue #577, opened 2024-09-05
Can't specify different upload and monitor serial ports.


paul-fornage:
The Sparkfun Pro Micro (and presumably the other 32u4 boards) use a different COM port depending on if it's in bootloader mode. 
I'm currently trying to just get a blink sketch working on a Pro Micro, but cant get it working. I am totally liable to be making a giant blunder, but I can only upload by specifying the COM port, and then the program immediately crashes after uploading with an unspecified error.

Tested on Arduino Uno [edit: with a new project] and without changing anything it works flawlessly.

stappersg:
On Wed, Sep 04, 2024 at 10:55:15PM -0700, Paul Fornage wrote:
> 
> I am totally liable to be making a giant blunder

Please continue, you will^Wmight learn a lot

 
> Tested on Arduino Uno and without changing anything it works flawlessly.

According https://store.arduino.cc/en-nl/products/arduino-uno-rev3
has an Arduino Uno an ATmega328p. And that is not a 32u4.


Groeten
Geert Stappers
-- 
Silence is hard to parse

paulwrath1223:
> According https://store.arduino.cc/en-nl/products/arduino-uno-rev3 has an Arduino Uno an ATmega328p. And that is not a 32u4. 

I should have been more specific, I have 2 boards.

When creating a project for the Uno, everything works fine (except the 'issue' described here: https://github.com/Rahix/avr-hal/issues/248)

However, for the Pro Micro, I need to specify a COM port for it to upload successfully, and the serial output port is different.

Rahix:
Yeah, I'm aware of this shortcoming in the current design.  I agree that it should be possible to specify an alternative port for the console and essentially disconnect the console logic from the flashing logic a lot more.

Let's take care of this after #522 is merged.

PR #683, opened 2025-09-25
Add option to specify serial port distinct from programming port

cmsd2:
possibly addresses https://github.com/Rahix/avr-hal/issues/577

usecase: i have an atmel-ice for programming using icsp, but i want to monitor the serial port using a silicon labs cp2102 usb->serial cable

this patch allows specifying the com port for the serial console separately from the programmer which in my case does not provide a serial port

### DIFF
--- a/ravedude/src/main.rs
+++ b/ravedude/src/main.rs
@@ -130,7 +130,12 @@ struct Args {
     #[clap(short = 'c', long = "open-console")]
     open_console: bool,
 
-    #[clap(short = 'C', long = "console-port", value_parser, env = "RAVEDUDE_CONSOLE_PORT")]
+    #[clap(
+        short = 'C',
+        long = "console-port",
+        value_parser,
+        env = "RAVEDUDE_CONSOLE_PORT"
+    )]
     console_port: Option<std::path::PathBuf>,
 
     /// Baudrate which should be used for the serial console.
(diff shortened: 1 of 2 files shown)


ravedude/src/main.rs
```diff
@@ -286,6 +294,8 @@ fn ravedude() -> anyhow::Result<()> {
         },
     }?;
 
+    let console_port = ravedude_config.general_options.console_port.clone();
+
```
Rahix, line 298: Why do this up here instead of futher down the function, right next to the

```rust
let port = console_port
```

part?

ravedude/src/main.rs
```diff
@@ -341,7 +351,9 @@ fn ravedude() -> anyhow::Result<()> {
                 "-b/--baudrate is needed for the serial console"
             })?;
 
-        let port = port.context("console can only be opened for devices with USB-to-Serial")?;
+        let port = console_port
+            .or_else(|| port)
+            .context("console can only be opened for devices with USB-to-Serial")?;
```
Rahix, line 356: This error message is no longer accurate.  I would suggest the following:

```suggestion
            .context("no programmer serial port or `console-port` was set")
            .context("cannot open console without a serial port")?;
```

Rahix (changes requested):
Very nice, thanks a lot for the contribution!  Please check my two comments below, then this PR is ready to be merged :)

ravedude/src/main.rs
cmsd2: no good reason. i'll move it.

PR #688, opened 2025-11-27
Fail loudly when I2C frequency too slow

meithecatte:
This makes sure that, if the I2C frequency chosen by the user is too slow to be handled without selecting a different prescaler, the issue doesn't get hidden by silent integer overflow.


Rahix:
Can you fix the formatting that CI complains about?

PR #689, opened 2025-11-28
atmega-hal: configure TC1 like all the other timers on atmega1280/2560

meithecatte:
Due to a missing write to the WGM bits in the TCCR1B register, PWM running off of TC1 was generating a frequency roughly two times slower. This commit fixes this issue.


Rahix (approved):
Thanks a lot!

PR #693, opened 2025-12-20
arduino-hal: Add support for Arduino Micro

Ivan-Johnson:
@Rahix, I've got a couple of questions:

1. Should we define `default_serial`?

   Given the fact that the ATmega32U4 doesn't have a hardware serial, my gut reaction is that we don't need it. However, both of the existing ATmega32U4 boards *do* define `default_serial`. Am I incorrect in assuming that `default_serial` is used for the default hardware serial bus?

1. Other than verifying that the new `arduino-micro` examples work (they do), are there any other tests that you'd like me to run?

    In particular: when I try running `cargo build` from the root directory, I get a *ton* of build errors. I get the same errors even without this PR applied ([build_logs.txt](https://github.com/user-attachments/files/24268353/build_logs.txt)). Is this an issue with my build environment, or is `cargo build` only expected to work when run from within an example package?


Rahix:
Thanks for picking this up!

> Should we define default_serial?

Yeah, I think we still should do it.  You can connect a separate USB-UART bridge to the RX/TX pins to talk to the board so it is nice to have the short-hand macro for setting up communication.

> Other than verifying that the new arduino-micro examples work (they do), are there any other tests that you'd like me to run?

No, this is entirely fine.  Running `cargo build` from the crate root is expected to fail, you must run it from inside each directory.  But in the end, just let the CI run those build tests for you ;)

Ivan-Johnson:
> You can connect a separate USB-UART bridge to the RX/TX pins to talk to the board

Ah, that makes sense; I didn't realize until now that UART only works on those two pins in particular.

I've pushed an update that enabled `default_serial`. I don't have easy access to a USB-UART bridge of my own, but I have created an `micro-usart.rs` so that we can at least verify that it compiles successfully. I assume that's good enough; as best I can tell the Arduino Micro and Arduino Leonardo hardware boards are essentially identical aside from their physical sizes

Rahix (approved):
Thanks a lot! :)

PR #694, opened 2025-12-29
Add USB support to arduino-micro

Ivan-Johnson:
I've started working on USB support for the Arduino Micro (#40). There's still a lot of work left to do, but I'd like to get your feedback on the big-picture design stuff before I get too far into the implementation @Rahix.

I've got a bunch of TODOs with design questions in the PR itself, but there are a couple of other things that I wanted to ask about as well:

* ~~I've created a new `usb-support` feature in both `arduino-hal` and `mcu/atmega-hal`; since I'm only testing `arduino-micro`, I'm thinking that it's better to have the new USB functionality disabled by default everywhere else. Is this the right call? Or is it better to remove the feature flag(s) and have the functionality always enabled in `atmega-hal`?~~

* When I make changes to this PR would you like me to squash the new changes with the previous version, or is it easier for you if I keep the commits separate until we're almost ready to merge?

### DIFF
--- a/arduino-hal/src/lib.rs
+++ b/arduino-hal/src/lib.rs
@@ -190,6 +190,26 @@ pub mod eeprom {
 #[cfg(feature = "board-selected")]
 pub use eeprom::Eeprom;
 
+#[doc(no_inline)]
+#[cfg(any(
+    feature = "arduino-micro",
+    feature = "arduino-leonardo",
+    feature = "sparkfun-promicro"
+))]
+pub use atmega_hal::default_usb_bus_with_pll;
+
+#[cfg(any(
+    feature = "arduino-micro",
+    feature = "arduino-leonardo",
+    feature = "sparkfun-promicro"
+))]
+#[macro_export]
+macro_rules! default_usb_bus_with_pll_macro {
+    ($p:expr) => {
+        $crate::default_usb_bus_with_pll::<$crate::DefaultClock>($p.USB_DEVICE, $p.PLL)
+    };
+}
+
 #[cfg(feature = "board-selected")]
 pub mod simple_pwm {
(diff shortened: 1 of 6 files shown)


Ivan-Johnson:
Over the last week I've spent quite a few hours reading through the atmega datasheet and working on implementing this PR. In the process, I've made a few design changes; in brief:

* Given all the time I'm putting into this I've decided that I might as well support all atmega32u4 platforms, rather than just the Arduino Micro. I haven't yet purchased an Arduino Leonardo or Sparkfun Promicro, but I'll do so eventually and verify that they work before we merge this PR
* I propose that for this PR, we disallow the hardware timer from using PLL while the USB bus is active. We can add support for that in a future PR. In order to make it easier to implement that feature without breaking backwards compatibility, I've made the `UsbdBus` struct private.

   There are docs with more details in the PR: [mcu/atmega-hal/src/lib.rs](https://github.com/Rahix/avr-hal/commit/1577c8380b950810bd890ee774ad64f33dcb2867#diff-37097d816b9f09507c70b98f268101706178980e23cc4566e8535d3a223b370e)

Ivan-Johnson:
Y'know what. I think I published this PR prematurely; I'll mark it as a draft.

At work I'm so used to the design being the hard part; when I work on something difficult, I'm now in the habit of asking for feedback early in order to avoid having to rewrite the whole thing from scratch multiple times. It turns out that doesn't really apply here because the design isn't the hard part; the vast majority of the difficulty is [grokking](https://en.wikipedia.org/wiki/Grok) how everything works 😅

Ivan-Johnson:
Status update: I got the happy path working last week, and I'm currently in the middle of cleaning up the code.

PR #700, opened 2026-01-31
added example: atmega2560/atmega2560-servo.rs

mocrsteel:
Based on some tinkering and the help, I would like to add this example  for the ATMega2560. I think it's a good one to show on how to interact with the MCU's registers too.

Code has been tested on my own Arduino Mega with an SG90 servo.


mocrsteel:
Refer to discussion #699

mocrsteel:
Updated the formatting by running `cargo fmt`. That should resolve the failed workflow. (I hope at least?)

PR #704, opened 2026-03-26
Atmega480x

peckpeck:
Add support to atmega4809 and 4808.

This is not finished and needs https://github.com/Rahix/avr-device/pull/250 to work.

I open the PR to start the discussion on this, because I think some items will necessitate a bit of rework of the hal API.


peckpeck:
For example the API for Usart is no adapted because it cannot handle alt mode. You would need to pass PORTMUX or  peripherals to the constructor to handle alt mode automatically.

peckpeck:
any thought on this before I go further ?

PR #707, opened 2026-08-26
Re-export avr_device from the HAL crates

### LINKED ISSUE #698: Consider re-exporting `avr_device` within `avr_hal`
Currently, to make use of `avr_device::interrupt::free`, I have to manually add a dependency on `avr_device` and make sure to keep it at the same version as the one being used by `avr_hal` (in order to make use of feature unification). I was a bit surprised to not find an appropriate reexport within `avr_hal`, especially considering that it already re-exports large parts of `avr_device` as `avr_hal::pac`.

ChrisJr404:
Fixes #698. The atmega-hal, attiny-hal and arduino-hal crates already re-export avr_device::entry and the selected chip module as pac, so this adds a plain `pub use avr_device;` to each. Downstream code can then reach items like avr_device::interrupt through the HAL without adding a separate avr-device dependency and keeping its version in sync manually.


Rahix (approved):
Thanks!

